PR #4, opened 2015-03-11
fix typo

Schaka:


AeroScripts:
Wow I'm surprised this wasn't noticed earlier...

PR #24, opened 2015-03-27
Fixed the map resetting when switching zones in cartograhper.

Logonz:
I use the old sane Player values if the player has the map opened on another zone, and will continue to update as soon as he switches back or closes the map.

The reset zone i switched to only force a reset if the mapID is -1 (Aka strange / null / don't exist)

Not 100% tested BUT should work, you are more familiar with the codebase hopefully you can look through it and know if this will fuck up anything else.


Schaka:
Thank you, gonna close the issue.

PR #182, opened 2016-03-09
Fix dragging frame off screen

wardz:
I think this should work better with SetClampedToScreen().
The old implementation doesn't seem to handle left/right/bottom corners properly in WoW window mode, and also if you barely move the questie frame near the top window corner, the whole frame will disappear untill a /reload.


Dyaxler:
Thank you! I've actually been meaning to add this. :) I'll merge this commit along with some other stuff I've been working on together into the next release.

Issue #246, opened 2016-05-21
Quest levels are incorrect


Ghamarro:
When I hover over an exclamation mark in the map, added there by Questie, I can see the quest level and the description just fine, but the quest level is sometimes incorrect. I checked some vanilla databases and verified that my realm, Kronos I, is actually blizzlike in terms of questing content, and they all gave me same result, so I'm guessing it's a bug or perhaps Questie's database is incorrect?

Dyaxler:
Quest levels are pulled from Questie's database which was built using: http://db.vanillagaming.org

Not all databases contain the same info since it's been passed down, around, interpreted from numerous other emu projects. Even WoWHeads data isn't the same anymore so it's not 100% accurate. You can use Questies quest level filters to force a quest to be hidden or shown depending on character level. You can read up on the command line options and FAQ here:

https://github.com/AeroScripts/QuestieDev/blob/master/Readme%20%26%20FAQ.txt

Ghamarro:
So is questie supposed to show the quest's required level or the actual level of the quest when hovering over the mark on the map? I remember this quest for instance [http://db.vanillagaming.org/?quest=6548](http://db.vanillagaming.org/?quest=6548) which says "level 12" when you hover over it in-game, but it is shown as a 18 quest in db.vanillagaming.com although with 12 as the required level. Is there a way to filter out map quest markers based on the actual quest level, not the required minum level?

Dyaxler:
>  [1917763378]={
  ['name']="Avenge My Village",
  ['startedType']="monster",
  ['finishedType']="monster",
  ['startedBy']="Makaba Flathoof",
  ['finishedBy']="Makaba Flathoof",
  ['level']=12,
  ['rr']=178
 },

Info in the database is correct. Level 12 is the "minimum" level you have to be in order to obtain the quest. db.vanillagaming.org says as well:

> 
    Level: 18
    Requires level: 12
    Side: Horde
    Start: Makaba Flathoof
    End: Makaba Flathoof
    Sharable

On db.vanillagaming.org - if you hover over Level: 18 under Quick Facts it'll show you the min - mid - max color values which is 15 - 19 - 22. In other words you can get the quest at level 12, the quest is RED from level 15-18, GREEN from level 19-21 and GREY from level 22 on up. Most sites say to do a quest before it turns green for maximum XP and reward. Not sure how db.valillagaming.org assigns these color values because I could swear that quests turn yellow then green. :)

[...]

PR #276, opened 2016-10-08
Added actual quest levels to use for tooltips.

ZoeyZolotova:
Updated the quest database with the addition of "questLevel" for each quest, which is the actual level of the quest, not the required level.
Removed/renamed/fixed some quests in the database.
Updated QuestieNotes to now have the available/complete quest tooltips show the questLevel instead of the minimum level.

### DIFF
--- a/!Questie/Database/addendum.lua
+++ b/!Questie/Database/addendum.lua
@@ -1423,10 +1423,6 @@ QuestieLevLookup = {
  ["Noboru the Cudgel"]={
   ["Bring Noboru's Cudgel to the leader of The Harborage in the Swamp of Sorrows."]={0,721449744},
  },
- ["Past Victories in Arathi"]={
-  [""]={0,1355872840},
-  [""]={0,1355873352},
- },
  ["Dreadnaught Waistguard"]={
   ["Korfax at Light's Hope Chapel in the Eastern Plaguelands will make a Dreadnaught Waistguard if you bring him the following items: 1 Desecrated Waistguard, 8 Wartorn Plate Scraps, 1 Arcanite Bar and 5 Cured Rugged Hides."]={0,3927722056},
  },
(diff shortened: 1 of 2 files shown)


Schaka:
What's the point of having the quest level? The level it was showing is for when you can pick up the quest. Questlevels can sometimes be 40, even if you can pick up the quest at level 18.

ZoeyZolotova:
Every other quest addon shows the quest level in square brackets, which makes it very confusing when Questie shows [18] but when you pick it up it says [40] and it seems like Questie is just wrong. It doesn't seem to me that the level you can pick up a quest is important information for a player. The only thing they really need to know is whether or not they can pick it up at all, and this change doesn't affect that since it still uses the min-level to determine when to show a quest on the map.

Schaka:
Good point, I'll merge it.

ZoeyZolotova:
Thanks! Mark #246 as fixed?

ZoeyZolotova:
For the record, I used https://vanilla-twinhead.twinstar.cz/ as the source of my data, with a few exceptions that I hard-coded (for example all the Call of Water quests are level 20 even though twinhead says 2 of them are 60). Group quests are marked with a + (e.g. [40+]), dungeon quests with a D (e.g. [25D]), raid quests with an R, and pvp quests with PVP.

Muehe:
Did you do anything about player level quests? No idea how twinhead lists them, but they would have level -1 in the DB.

ZoeyZolotova:
I was not aware of anything like that, so no.

Muehe:
Did a quick lookup, shouldn't be much of a problem:

```
mysql> SELECT entry FROM mangos.quest_template WHERE QuestLevel = -1;
+-------+
| entry |
+-------+
|  9029 |
|  9367 |
|  9368 |
+-------+

```

PR #279, opened 2016-10-24
Quality of Life change: Quests in the same location now combine their tooltips.

ZoeyZolotova:
Before, quests in the same location would overwrite each other on the map. Now you can see all the available quests.

example1
example2
example3
example4


Schaka:
This looks awesome. I'll merge it and actually put out a new release.
If you are really interested, check out the branch of "Questie2", it has tons of great stuff that ended up unfinished, including huge database cleanups and "areas" instead of single icons.

ZoeyZolotova:
Oh cool, I'll have a look. Maybe we can get available quests from objects and items working!

AeroScripts:
Maybe I'm mis remembering but I swear the available quests worked with objects when I first implemented it (with the old DB)

Issue #270, opened 2016-08-02
Very odd tooltip issue


ShaunDreclin:
So if I mouse over a monster, it shows me the quest info on the tooltip. But if I move my mouse away and then put it over that monster again, the quest info is gone and won't come back until I move my mouse away, wait for the tooltip window to completely fade away, then put my cursor over it again.

I'm guessing here, but it seems like adding quest info to the tooltip is only happening the /first/ time it pops up, and it's not re-adding the info when the tooltip is modified without being recreated.

Tested this with all other mods disabled and it's still happening, so it's not a conflict issue.

ShaunDreclin:
Talked to Myanrey about it and they found they can reproduce it by mousing over npcs or items that trigger quest info in the tooltip very quickly - but for me it's happening regardless of speed and which things I mouse over, any tooltip is doing it.

ShaunDreclin:
More info: I downloaded questie from http://addons.us.to/addon/questie and clicked the recommended link, https://github.com/AeroScripts/QuestieDev/archive/master.zip

That version of the addon is having the issues I posted about. When I went and downloaded it from this url instead, https://github.com/AeroScripts/QuestieDev/releases/download/v3.1/Questie-3.1.zip < that one worked much better and is behaving more similarly to how Myanrey was reporting it.

I'm actually having difficulty reproducing the bug now, but I noticed it was happening when I moused over a *non pvp flagged* player and then to a quest mob, it wasn't showing the quest info.

incore97:
I have the same problem. U found any solutions yet?

ShaunDreclin:
Quit vanilla for legion launch haha I may look into this again when/if I get bored of legion

AeroScripts:
Lots of people are reporting this issue. If nobody else discovers a solution before I start, this will be the first thing to get fixed.

AeroScripts:
Going to test a7437b400e8318a38b463426cb5b606e27182c93 because I believe this may be related to the tooltip updates @laytya did

laytya:
I know why its happend. [a7437b4](https://github.com/AeroScripts/QuestieDev/commit/a7437b400e8318a38b463426cb5b606e27182c93) not related to it. I ll check if i can fix it quick.

PR #283, opened 2016-11-11
Fix missing tooltip in #270

chrisjordan82:
Tooltip missing when triggering mouseover a second time before tooltip fades out. Removing QuestieDone check and forcing tooltip update resolves issue. Further refactoring required to Questie:Tooltip() in Modules/QuestieNotes.lua:218 in relation to this change?

### DIFF
--- a/!Questie/Questie.lua
+++ b/!Questie/Questie.lua
@@ -237,6 +237,7 @@ function Questie:OnLoad()
 	this:RegisterEvent("ADDON_LOADED");
 	this:RegisterEvent("VARIABLES_LOADED");
 	this:RegisterEvent("CHAT_MSG_LOOT");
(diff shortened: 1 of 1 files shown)


laytya:
ill provide fix for #270 very soon...

chrisjordan82:
Think I've got it. The frame tooltip is being cleared somewhere else in the default ui. Solution appears to be to null QuestieDone in the UPDATE_MOUSEOVER_UNIT event.

laytya:
i did fix already. Tested.

chrisjordan82:
me too

Schaka:
While this works, the fix laytya provided is more clean throughout the modules and I merged that one instead.

PR #332, opened 2017-01-08
fix typo in findFirst

Franzz:
not a bug but could lead to one


Dyaxler:
Nope. It's supposed to be like that. Function declarations can be used or not used....

Your proposed commit would break slash commands...

AeroScripts:
Something is funky with the code though. Maybe something like local i=string.find(haystack, needle or " "); would work

PR #386, opened 2017-01-28
Replaced primitive exact coordinate based grouping with distance based Hierarchical Clustering.

ZoeyZolotova:
Quests used to be grouped together if they have exactly the same coordinate. Now they are clustered by distance to each other using a Hierarchical Clustering algorithm. The clustering distance for the world map is set to 0.025 (or 2.5 in terms of coordinates that are shown to players), while the distance for the minimap is set to 0 so that the position of the quest is as accurate on the minimap is possible. The algorithm also limits the cluster size to 5 so that too many quests don't get grouped together and become overwhelming to the tooltip that displays their info, with the exception of quests that are at a distance of 0 from each other which always get clustered together. These values can easily be tweaked in the code, which should make it possible to easily make them user-configurable.

### DIFF
--- a/!Questie/Modules/QuestieNotes.lua
+++ b/!Questie/Modules/QuestieNotes.lua
@@ -938,26 +938,32 @@ function Cluster:CalculateClusters(clusters, distanceThreshold, maxClusterSize)
     end
 end
 
-function Questie:AddClusterFromNote(frame, v)
+function Questie:AddClusterFromNote(frame, identifier, v)
     if clustersByFrame == nil then
         clustersByFrame = {}
     end
     if clustersByFrame[frame] == nil then
         clustersByFrame[frame] = {}
+    end
+    if clustersByFrame[frame][identifier] == nil then
+        clustersByFrame[frame][identifier] = {}
     end
     local points = { v }
     local cluster = Cluster.new(points)
-    table.insert(clustersByFrame[frame], cluster);
+    table.insert(clustersByFrame[frame][identifier], cluster);
 end
 
-function Questie:GetClustersByFrame(frame)
+function Questie:GetClustersByFrame(frame, identifier)
     if clustersByFrame == nil then
         clustersByFrame = {}
     end
     if clustersByFrame[frame] == nil then
         clustersByFrame[frame] = {}
     end
-    return clustersByFrame[frame]
+    if clustersByFrame[frame][identifier] == nil then
+        clustersByFrame[frame][identifier] = {}
+    end
+    return clustersByFrame[frame][identifier]
 end
 ---------------------------------------------------------------------------------------------------
 -- Finds the index of an item in a table. Not sure if a function already exists somewhere.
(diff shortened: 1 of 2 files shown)


Muehe:
Ok, so I downloaded and tested your PR and it works fine. No overly long tooltips, finishing notes on top, minimap notes not merged and everything.

One little bug I found though: My test char was on the quests [Sarkoth](https://vanilla-twinhead.twinstar.cz/?quest=790) and [Sting of the Scorpid](https://vanilla-twinhead.twinstar.cz/?quest=789). Sarkoth drops items for both quests, and on earlier versions his note would correctly merge to show both drops, however with your current version it shows only one drop, depending on which quest has been accepted first (or is found first in the quest log if you are already on them). This we should fix before we merge it into master. ([Saved Variables after NUKE](https://gist.github.com/Muehe/6e6c52f4232e7ce22f951f923cccf7e0))

**Offtopic:** You got one of those nasty useless merge commits in your log, so I took the liberty of cherry-picking your actual commits into a clean master branch: https://github.com/Muehe/QuestieDev/commits/ZoeysNoteMerge

You can get it into a fresh clone of Questie like this (pushing this might require some fixing if you don't properly reset your forks master branch, or just delete and re-fork, as your current commits are already on that. If you have any problems shoot me a mention on Discord. I'll also upload an extended version of this explanation as the CONTRIBUTING.md later):

```
git clone https://github.com/AeroScripts/QuestieDev.git
cd QuestieDev
git remote add muehe https://github.com/Muehe/QuestieDev.git
git checkout -b YourNewLocalBranch
git fetch muehe
git merge muehe/ZoeysNoteMerge
```

Then you can work on this branch and assuming your remotes look like this (add your repo as a remote like above):
```
git remote -v
muehe	https://github.com/Muehe/QuestieDev.git (fetch)
muehe	https://github.com/Muehe/QuestieDev.git (push)
origin	https://github.com/AeroScripts/QuestieDev.git (fetch)
origin	https://github.com/AeroScripts/QuestieDev.git (push)
zoey 	https://github.com/ZoeyZolotova/QuestieDev.git (fetch)
zoey 	https://github.com/ZoeyZolotova/QuestieDev.git (push)
```

you can publish changes to a new branch on your fork like this

`git push zoey YourNewLocalBranch:YourNewRemoteBranch`

so others can pull your new changes into their branches. This way, once your changes get merged into the official repo you can just delete your local branches and the ones on your fork (use Github interface for this). To continue working on your next change you do:

```
git checkout master
git fetch
git pull
git checkout -b NextLocalFeatureBranch
```

Dyaxler:
> One little bug I found though: My test char was on the quests Sarkoth and Sting of the Scorpid. Sarkoth drops items for both quests, and on earlier versions his note would correctly merge to show both drops, however with your current version it shows only one drop, depending on which quest has been accepted first (or is found first in the quest log if you are already on them). This we should fix before we merge it into master. (Saved Variables after NUKE)> 

Uhhh... I think that bug is in my code. Quest Monsters and Quest Item tool tips are done by a separate function which is separate from Mini Map and World Map tool tips.

Dyaxler:
@ZoeyZolotova 

You left the rounding code in place? I thought this was obsolete now that you're using this new clustering method? Unless I'm reading through the PR wrong... which is entirely possible. :D

ZoeyZolotova:
In response to the above comments, 
@Muehe i've fixed the issue with quest objectives at the same coordinate not being combined. All that off-topic stuff went over my head. I just use SourceTree and commit as often as necessary to not lose progress in case of a hard drive failure.
@Dyaxler the rounding code was removed, I think you said you were reading through the PR wrong :)

PR #397, opened 2017-02-03
Add event handler for "QUEST_ACCEPTED" events

gabrielecimolino:
There has been an issue with Cartographer that requires the user to reload
the UI in order to see Questie's quest objective markers on the map. By
updating Questie when a quest is accepted this issue is fixed.


Dyaxler:
I've been meaning to look into this. It only happens some of the time though especially after all the tweaks I've made to my dev build with changes I haven't yet checked in. I'm being very careful especially with all the new clustering code that we've just checked in. This bug was uncovered when I began reducing Questies memory footprint and some of the functions started firing out of order or some faster than others causing a timing issue.

I'll add these changes to my dev build and give it a whirl...

PR #406, opened 2017-02-13
Added UI support for settings

### LINKED ISSUE #315: Questie needs a configuration UI (not just slash command)
The configuration could be much more intuitive. A UI can be generated from the same logic that handles the slash command.

ThiconZ:
Added /questie options to display the newly added user interface for
managing settings. All the old commands still work but now there is an
options for those who would like a simple interface for changing their
settings. The UI file is compatible with WoW UI Designer for easy
updating in the future.

Should resolve #315


Schaka:
Looks clean and simple. It was meant to only be implemented after 3.70, but by the looks of it, it's loosely coupled enough, that it can be merged even now. 

What's the WoW UI designer though?

!Questie/QuestieOptionsForm.lua
```diff
@@ -0,0 +1,127 @@
...
+
+    QuestieConfig.hideObjectives = Questie:toboolean(QO_hideobjectives:GetChecked())
+
+    QuestieConfig.corpseArrow = Questie:toboolean(QO_corpsearrow:GetChecked())
+
+    QuestieConfig.clusterQuests = Questie:toboolean(QO_clusterquests:GetChecked())
+
+    DEFAULT_CHAT_FRAME:AddMessage("Questie Options Applied", 1, 0.75, 0);
+
+    QuestieOptionsForm:Hide()
```
ZoeyZolotova: Does need to have 
```Questie:Toggle();
Questie:Toggle();``` at the end to actually apply the changes?

File: !Questie/QuestieOptionsForm.lua [hunk 2]
```diff
@@ -0,0 +1,127 @@
+function Questie:OptionsForm_Init()
+    QO_showlevel = getglobal("QuestieOptionsForm".."AlwaysShowLevelCheck")
```
ZoeyZolotova: maybe put all these `QuestieOptionsForm` into a variable in case the form gets renamed.

!Questie/QuestieOptionsForm.lua
```diff
@@ -121,7 +121,9 @@ function Questie:OptionsForm_ApplyOptions()
 
     QuestieConfig.clusterQuests = Questie:toboolean(QO_clusterquests:GetChecked())
 
-    DEFAULT_CHAT_FRAME:AddMessage("Questie Options Applied", 1, 0.75, 0);
+    DEFAULT_CHAT_FRAME:AddMessage("Questie Options Applied", 1, 0.75, 0)
+
+    Questie:Toggle()
```
ZoeyZolotova: It needs it twice to basically re-draw all the questie stuff. It's the same as pressing the Toggle Questie button twice to turn it off and on to apply all the settings. Most of the other slash commands do the double toggle.

ThiconZ:
@Schaka [WoW UI Designer](http://www.wowinterface.com/downloads/info4222-WoWUIDesigner.html) is a vanilla (and TBC?) application that allows you to design interfaces, and addons in general, through a visual studio like environment.

ThiconZ:
@ZoeyZolotova commited all your suggestions, I did't make the form name a global though - was unsure if it'd really need to be one.

ZoeyZolotova:
form name looks good :) is the screenshot in #315 from before you added the checkbox for clustering?

ThiconZ:
Awesome. Yeah I didn't see your clustering update until after I posted the screenshot but here is how it currently looks (adding more settings to it is pretty easy).
image

ZoeyZolotova:
Sweet. tbh if it were up to me i would make the command to open this just be `/questie` since that's what users are most likely to try first. Would need to get the others' inputs on that though.

Maybe make `/questie` show this menu and also write in the chat frame `/questie help -- View command line options`

ThiconZ:
It may be worth noting that when going through the UI it does not warn users that certain options would need a UI reload before they are applied, such as the ```Tracker Enabled``` and ```Show Tracker Header``` options. Would it be desired that both these warnings and the ReloadUI() on accept functionality is brought over to this?

ZoeyZolotova:
It's kind of weird that pressing No still applies the other changes. Maybe if the question was "Do you want to reload UI or revert these settings?" and have the buttons say "Reload UI" and "Revert". 

Also it's unfortunate that it doesn't displayed the detail as to why it needs to be reloaded, especially for things like the Tracker.

ThiconZ:
Added in your suggestions. If this looks good I'll commit it:
image

All of the settings that require a reload are now stored in a table that allows you, for the warning message, to define their display name and what 'description' they will display.

RBDaniel:
This is looking great. One less thing to do after releasing 3.70. :) 

Zoey and I have changed quite a bit. And some of the "option" names are going to change to better reflect what they actually do! Can we make the options alphabetical? Also, it would help to either have a tool tip or some kind of label to indicate what the default setting(s) were originally and to give a brief description on what the option does. 

Can we add another tab? Maybe (Options)/(Troubleshooting)/(FAQ)?
Options would be what you have there...

Troubleshooting can contain buttons for:
    /questie NUKE
    /questie clearconfig
    /questie cleartracker
    Maybe make a button to reset all the options back to default?

FAQ can contain a more truncated FAQ that what is currently on GitHub.

Dyaxler:
This is looking great. One less thing to do after releasing 3.70. :)

Zoey and I have changed quite a bit. And some of the "option" names are going to change to better reflect what they actually do! Can we make the options alphabetical? Also, it would help to either have a tool tip or some kind of label to indicate what the default setting(s) were originally and to give a brief description on what the option does.

Can we add another tab? Maybe (Options)/(Troubleshooting)/(FAQ)?
Options would be what you have there...

Troubleshooting can contain buttons for:
/questie NUKE
/questie clearconfig
/questie cleartracker
Maybe make a button to reset all the options back to default?

FAQ can contain a more truncated FAQ that what is currently on GitHub.

Dyaxler:
Also some of the text looks lop sided.... you can shorten some of the lines so they look cleaner.

For example, "The following settings will require a UI Reload to apply:" you can rewrite it to be shorter.

"Some of the options require ReloadUI to apply:"

If we include tool tips for the settings you won't need to summarize what each one does further reducing the amount of info. The final line can be summarized by just saying "Reload or Revert changes?"

ThiconZ:
I have made the settings in the UI all be alphabetical now from top to bottom, left to right. To keep things consistent I am going to reorder the /questie settings output alphabetically as well.

As for making another tab, the best I can do is use a dropdown and just switch between option visibility (don't look too bad on the users end however managing it will be a bit more cumbersome).

I am currently in the process of trying to get tooltips working for it.

ThiconZ:
@Dyaxler I just noticed the alwaysShowLevel setting was removed in the 3.69 release. Is this setting going to be coming back? If not, I'll be removing it from the UI settings list.

Dyaxler:
That was never an actual command. There were CVAR's for it but no slash handler. It was in there for MONTHS before I noticed it wasn't being used and removed it. I think its purpose was originally for something that never came to light when another feature made it obsolete before it was implemented. I went back to the very first build I released ever for Questie and it was in there back then too. I just missed it in all the code cleanups.

Dyaxler:
> As for making another tab, the best I can do is use a dropdown and just switch between option visibility (don't look too bad on the users end however managing it will be a bit more cumbersome).
> 

It's cool if you can't do tabs. This is awesome for a v1 Options Panel. :)

If you want to learn how to do tabs the best source is to look at the WoW Client UI files. Take a look at UIOptionsFrame.XML and UIOptionsFrame.lua... I find it's the best way to see how Blizzard does it rather than plow through forum posts at people's interpretation on how it should be done.

Muehe:
This is how shagu made tabs, all lua and pretty straightforward: https://github.com/shagu/shaguquest/blob/master/ShaguDB_GUI.lua#L188-L196

```lua
SDBG.buttonSpawn:SetScript("OnClick", function()
  SDBG.buttonSpawn:SetBackdropColor(1,1,1,.15)
  SDBG.buttonItem:SetBackdropColor(1,1,1,.05)
  SDBG.buttonQuest:SetBackdropColor(1,1,1,.05)

  SDBG.spawn:Show()
  SDBG.item:Hide()
  SDBG.quest:Hide()
end)
```

ThiconZ:
This is now how it currently looks. The Tool Tips move close to where the moused over setting was. Ones that require a ReloadUI also have a second line stating that so the user knows before they click Done that a reload is going to be needed.
image

ThiconZ:
@Dyaxler unsure if this would be necessary but perhaps the /help command and the tool tips could use the same descriptions so they are stored in a table once in the main Questie.lua file and then both the help command and the tool tip function just load it from there - to keep consistency.

Dyaxler:
Absolutely... that is an awesome idea! Matter of fact, the help menu was completely rewritten. But it'll have to be rewritten again before release since we're thinking about changing some of the names of the commands so they make more sense. Now that we have a Settings UI to consider... it needs to be done even more.

I think a single file containing all global strings should work. Both command names and UI Settings names along with their descriptions. That way were making changes in only one place. I think that would be better than just drawing the tool tips directly from the help menu.

ThiconZ:
Do you want to handle doing that or should I?

Dyaxler:
No it's fine the way it is. If you waited for us, this UI Panel would be blocked from being pushed and available to the general public. We're going to change the command names... just not until we get these other features we're working on stable.

ZoeyZolotova:
Would it be possible to replace the input fields with sliders? 0 to 1 for the tracker alpha, 0 to 60 for the min and max level filters, and some appropriate range for the tracker scale.

ThiconZ:
Would something like the following be what you had in mind?
image

ZoeyZolotova:
That's good :) can you make it more like this though?
image
Left number being min, right max, and current in the middle?
And I suppose call it Tracker Transparency going from 0% to 100% (but you'd have to invert it since 100% transparent means 0 alpha).

ZoeyZolotova:
And add a Reset Defaults button which calls `/questie clearconfig`

ThiconZ:
This is how it looks now (and is completely hooked up)
image
However, it may be better to name it Tracker Opacity to have the displayed range be 0% - 100% to remove the extra math required to convert to and from the displayed and internal values that are currently used to have it display one way but work another.

Also, I may have come up with a 'simple' to implement solution for tabs. I'll be looking into it more after the EditBoxes are all converted over to Sliders.

ZoeyZolotova:
Looks good, but I think the extra math is worth it for the user experience. It really should say Tracker Transparency with 0% on the left and 100% on the right.

ThiconZ:
This is the new look of the UI:
image
I'll be adding in the Reset button next and looking into the possibility of adding Tabs. (A commit of just the Reset button being added will be done first though)

Muehe:
Looking good! Btw. we have a Discord server, you are welcome to join us: https://discord.gg/phKnHVu

ZoeyZolotova:
How easy is it to select a specific max/min level filter level?

ThiconZ:
All sliders feel pretty easy to select a specific value.

Dyaxler:
I feel compelled to point out that "Resize Worldmap" is only used when you aren't using a MapMod like Cartographer or Metamap.

ThiconZ:
With the Reset Defaults button:
image

Dyaxler:
I had some time to finally test this... here are some observations.
- The Tracker Scale's max value is set way too high. Anything beyond 142% gets truncated and makes the tracker look squished.
image
- The Tracker Transparency doesn't need a ReloadUI (at least in my dev build). There is now a nice border so I see no reason to allow a Transparency of zero. Anything less than 10% isn't very noticeable.
- For me personally, it's kind of annoying to have to reopen the Settings UI after clicking Done. Perhaps we should consider changing it to an "Apply" button and unless a ReloadUI is necessary, keep the settings window open so a player can explore other settings freely.
- Bold Colors uses the function QuestieTracker:FillTrackingFrame() to apply it's color palette.
- The scale/size of the Options Panel could match the scale/size of the tracker. Then we could tie both the Tracker Scale and Options Panel Scale into one setting or have a separate setting. It's kind of big.
image
- The Options Panel should also be movable. It's super easy to do. Just add the following code:
image
- I don't think having Show Map Aids is necessary (see note below).

Now that I see all the different options in the Settings UI Panel I think some of the names needs some work - this would be a great place to have this discussion. Thoughts? Here is mine - go ahead and poke fun because I SUCK at naming things.
- Arrow Enabled and Corpse Arrow isn't very intuitive. Let's call them Quest Arrow and Corpse Arrow.
- Instead of Bold Colors I think we should use Alternate Colors or Optional Color?
- Instead of Cluster Quests (since it's functionality has been expanded to objectives) we could call it Clustered Icons? Or Cluster Icon? Or Cluster Markers?
- Hide Objectives - I've been thinking about changing how this works. Instead of hiding all Objectives everywhere (unless they are forcibly being tracked - Always Show Quests: Off and actively tracked on the Questie Tracker) we should only hide Objectives on the Minimap to help reduce clutter. We shouldn't touch the World Map. So this settings name should be Hide Minimap Objectives. Or leave it the way it is?

Note: Using Questie:Toggle() should be avoided since it's use has been reduced to only two functions --> Map Aids, which is a "backup" of sorts for the Questie Toggle button, and the Questie Toggle button that is located in the World Map frame.

ThiconZ:
That commit should fix the issues you mentioned @Dyaxler. However, for the Apply button will there still be a 'Done' button to close the form or will closing it be done via a notification prompt after the user clicks Apply each time. If there is a dedicated 'Done' button to close the form, will it prompt or auto-apply any changes made since the user last applied them or will it just simply close the form and discard any changes?

Options have not been renamed yet as a total list of new names should be decided upon before any have their names adjusted in the ui.

Dyaxler:
Hmmm... good point. There might be a way to make the "Apply" button light up vs being greyed out when there are changes to be applied? I don't know. I've never done something like this before. Then you could just simply have an X box in the upper right-hand corner.

haverdaden:
Would it be possible to scale the tracker down even more than 100%? Also a minimize button would be nice :)

ThiconZ:
Scaling support needs to be done by @Dyaxler, I can make the UI slider go to whatever limits he make it support.
@haverdaden what exactly do you mean by a minimize button? Like a button on the tracker itself to minimize the entire list or a button to minimize the options UI or something else entirely?

haverdaden:
Alright. I'm not familiar with where you are supposed to comment and not comment on github yet so maybe you could pass this scale request to him. 

I'm thinking a little button anchored to the top right of the tracker running the function "/questie mintracker".

SalviaSage:
haverdaden, you can already minimize all the quests that are displayed in the quest tracker by clicking on the "QuestLog Status". Quest tracker header must be enabled for this to work.

Dyaxler:
You can set the scaling lower than 100%... but keep the default at 100%.

And yes, there is already a minimize button. If you aren't using the header we'd prefer to not add an ugly button to a clean tracker. This is why the back ground and header aren't on by default. We present a very clean, minimalist tracker by default. However, there is a minimize toggle. `/questie mintracker` that you can use. It would be the only way because if you minimize the tracker without the header, you might forget that you have done that and your tracker will essentially be gone without any visual indication that it's minimized... this is why that button doesn't work unless the header is enabled. At least with the header visible you can visually see that the tracker is there, just minimized.

Muehe:
This PR has been handled [here](https://github.com/Dyaxler/QuestieDev/pull/1). Closing.

PR #463, opened 2017-11-05
Use a more convenient way to detect pfUI minimap

shagu:
The old function will result in nil errors for every addon that includes the pfUI API but isn't pfUI itself. This PR will use the pfUI.minimap field which is only existing if pfUI is active and has enabled minimap support.


Muehe:
Thanks :+1:

PR #495, opened 2018-03-11
Fix required quests for quest 794

krangerich:
* Quest 794 (Burning Blade Medallion) is only available after Quest 792 (Vile Familiars)


Muehe:
Merged, thank you! :+1:

PR #499, opened 2018-04-08
Added 'hide objectives' option into settings gui

llyama:
#Adds the hide objectives option into the setings GUI.

The setting is the one added in ee4e60999549c1e466fd59d13d01e854344e00ff

image

It's an enhancement to the existing gui issue #315

### DIFF
--- a/!Questie/Modules/QuestieInterfaceTexts.lua
+++ b/!Questie/Modules/QuestieInterfaceTexts.lua
@@ -100,6 +100,13 @@ QuestieInterfaceTexts = {
                 ["description"] = "This option toggles Questies world map and minimap notes on and off.",
             },
         },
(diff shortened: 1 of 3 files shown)


Dyaxler:
The Questie button on the world map hides all the objectives. Redundant feature.

llyama:
The Questie button on the world map toggles off **all** map icons when i press it. The quest icons (!,?) and the icons for monster locations.

The command line /questie hideobjectives that you added in ee4e60999549c1e466fd59d13d01e854344e00ff hides only the icons for monster locations and leaves the quest icons displayed.

My change is to expose this command line option in the settings gui.

Dyaxler:
Ahhhhh... I get it! OK. Nice one. Thanks!

PR #521, opened 2018-11-04
QuestieTBC: Small fixes to solve constant Lua errors

VideoPlayerCode:
The MBB_Exclude table only exists if the "MBB" addon is loaded.

I know that QuestieTBC is archived. But this is a really ugly error that happens on every startup of Questie (and shows the error _on screen_ if you have an error handler like "!Swatter"), so let's fix it! It was caused by trying to use the "MBB_Exclude" global table without first verifying if the "MBB" addon even exists!

This fix may also be necessary in the core Questie (master) branch. I didn't look.

PS: Could any of you please consider rebasing QuestieTBC on top of master, to get the latest 350 commits into QuestieTBC too? Or perhaps they don't matter?


VideoPlayerCode:
I've just been to Outland with QuestieTBC enabled... and... ew! I just added a small extra commit: "QuestieTBC: Fix invalid access to missing rq field". It takes care of the *constant* Lua "trying to access field rq on nil value" errors that you get whenever you turn in Burning Crusade quests while QuestieTBC is activated. Because Questie doesn't have a hash-map about the Outland quests, but tries to access them anyway, it was causing an error.

This error will also happen in Vanilla Questie if the user ever finds some quest that isn't included. So you guys should probably copy this code-fix to the `master` branch too!

VideoPlayerCode:
I added one final commit, "QuestieTBC: Silence errors about missing quests". The previous ones are meant to be merged manually by you into Vanilla (master) branch too, since I suspect those bugs exist there too.

Whereas this final one is _just_ intended for QuestieTBC. It turns off the chat-window spam "please report this missing quest to GitHub", since Questie spams that message about 10-15 tims in big, red letters every time you interact with an Outland quest NPC or pick up a quest item in Outland. It's hell to have QuestieTBC enabled while in Outland. This commit fixes that.

So... all 3 commits, in order, make life good for Outland users for now!

- Fixing the "MBB" addon dependency check, to ensure we check if MBB is even _loaded_ before we try to access MBB's global variable. (Needed to be merge to `master` too.)
- Fixing Questie's internal "get parent/pre-quest Id" code, to ensure it doesn't try to read the "rq" field from quests it doesn't know about. (Needed to be merged to `master` too.)
- Turning off "this quest is not know!" errors in QuestieTBC, since it's an archived/paused version, and that spam is making life hell in Outland... So let's turn it off so that any QuestieTBC users will get peace and quiet. :-)

The two first commits need to be manually merged into `master` by any team member who feels like it... In fact, I had a quick look and I see that a _partial_ fix for the 2nd issue has already been put into master. Open https://github.com/AeroScripts/QuestieDev/blob/master/!Questie/Modules/QuestieQuest.lua and search for the word `'rq'`. I say a _partial_ fix, because one line, 1072, still has the bug: `https://github.com/AeroScripts/QuestieDev/blob/master/!Questie/Modules/QuestieQuest.lua#L1072`. Just search for `'rq'` and do the same fix there as you've already done in two other code locations in that file on the master branch. :-) Or... adapt my "use a new `GetRequiredQuest(questHash)`" function to safely take care of it everywhere instead... Your choice!

VideoPlayerCode:
Alright, I've added 3 more bugfixes. Now QuestieTBC is finally bug-free enough to use 24/7.

Do whatever you want with this. I suggest merging it, just so that the archived QuestieTBC codebase is in a working state. If not, people should go to my fork and download it instead, since the official branch here has a couple of very game-breaking bugs that I've now fixed. ;-)

Take care!

My fork:

- https://github.com/VideoPlayerCode/QuestieDev/tree/QuestieTBC
- Download: 
https://github.com/VideoPlayerCode/QuestieDev/archive/QuestieTBC.zip

Muehe:
Thanks for your work, and sorry for my late reply. I'm going to try and test this when I have the time and then merge it.

But to be honest, QuestieTBC needs a rebase on master. The master branch is more or less stable now, with only the odd bug fix here and there, but QuestieTBC is still based on Questie 3.1, which is horribly outdated. I have tried to do a quick rebase of your branch onto master today, but there were many merge conflicts. Seems like it would take some tinkering to get right.

Muehe:
So I did a quick test and for some reason some notes appear on the minimap, but not on the world map. Is this the same for you?

Another thing is there is a constant error when the `questiedebug` chat window is created. This is something that should be solved before merge (might be a preexisting condition from before your PR).

VideoPlayerCode:
@Muehe You're welcome! And yeah I tried a rebase on master too, and got so many merge conflicts.

The saddest part is that the original QuestieTBC work was one big crazy commit, without being split into individual commits, so porting all of its changes will be tricky. It also had some pointless variable renamings and whitespace changes that makes merging even harder, since they show up as "changes". :\\

https://github.com/AeroScripts/QuestieDev/commit/f5a056a6e91ae3730b1b7eb06bfb336d23d68c76

That commit would at least be manageable if it had been split into one "upgrade external library versions" and one "tbc fixes" patch.

I tried to take a look at it manually and just ignoring all changes that happened in library-folders.

It looks like the only solution will be to take the latest Master, then look at that huge commit and manually read its changes (GitHub's diff view here is good) and applying relevant code changes manually. But even that's hard work... since it's impossible to know which changes belong together.

> So I did a quick test and for some reason some notes appear on the minimap, but not on the world map. Is this the same for you?

Hmm nope, from what I've seen I see all "!" quest npcs and "bag" collection icons on both the minimap and world map. What's the problem? I'll look specifically for it. :-)

> Another thing is there is a constant error when the `questiedebug` chat window is created. This is something that should be solved before merge (might be a preexisting condition from before your PR).

I haven't looked at the questiedebug feature. It's a dev mode, I assume? And yeah that's definitely a pre-existing bug, there were lots of those. :) None of the changes affect anything that could affect the debug mode.

Muehe:
> I haven't looked at the questiedebug feature. It's a dev mode, I assume?

It is just a chat window. When you manually add it next to the existing ones Questie uses it to print debug messages.

> Hmm nope, from what I've seen I see all "!" quest npcs and "bag" collection icons on both the minimap and world map. What's the problem?

The problem is that I don't see them all. :smile: Might be related to the debug thing though. It keeps happening although I removed the debug chat. I'll remove WTF and check again.

VideoPlayerCode:
Didn't have time to check this today and tomorrow will be tough as well. But I guess there's no urgency.

Muehe:
Ok, so I got a chance to test this some more. The problem is not that the notes are not there, they are in the wrong zone. And this does not affect quest start notes, only quest finish and quest objective notes. E.g. if I accept quest in the Orc starting zone, their objectives appear in Darnassus, and if accept a quest in Teldrassil, their objectives appear in Stonetalon Mountains.

So some zone IDs seem to be getting misplaced somewhere. What puzzles me is why you do not seem to have the same issue.

Muehe:
Well apart from some zones being wrong and the debug chat not working, the PR gets rid of the constant errors, so it's a definite improvement and I'm merging it.

If you are interested in fixing the zones or maybe even maintaining the TBC branch you can join us on our Discord, or just send new PRs. But I'm sure aero wouldn't mind giving you push access to the repo if you want it.

Thanks again for your work! :+1:

VideoPlayerCode:
> Well apart from some zones being wrong and the debug chat not working, the PR gets rid of the constant errors, so it's a definite improvement and I'm merging it.

Glad to hear it! Thanks to these fixes I've been able to run with Questie constantly enabled, even when in Outlands, which has been a nice feeling. ;-)

My weekend was really busy. I'm working on the debug chat bug right now. I'll make a separate PR for that fix when it's done.

> If you are interested in fixing the zones or maybe even maintaining the TBC branch you can join us on our Discord, or just send new PRs. But I'm sure aero wouldn't mind giving you push access to the repo if you want it.

Unfortunately I have wayyyyy too much to do in terms of other obligations, otherwise I'd gladly join the team.

As for fixing zones, I don't know how that would be done. Is there some automatic way of extracting zone/subzone IDs or is it a tedious, manual process of running all over the world and capturing the data manually? ;-)

VideoPlayerCode:
I see the problem with the debug printing now. It's using `function Questie:debug_Print(...) for i = 1, arg.n, 1 do` which errors with "attempt to index global 'arg' (a nil value)".

The `arg.n` seems to be some old technique of getting the argument count for vararg (...) functions.

From tbc onwards, I think the correct method is `select('#', ...)` to get the count. I'll check for the correct new method and fix this. New PR incoming soon for that final fix! :-)

VideoPlayerCode:
It's up. All done! :-)

https://github.com/AeroScripts/QuestieDev/pull/523

Apart from the potential zone-ID issues you've mentioned, I don't see any other problems with QuestieTBC at the moment. Running smoothly without errors! ;-)

// Johnny

Muehe:
Nice, and having the debug channel working should help with taking care of the zone problem.

About the zones. Like I said before, the quest start notes are working fine in the affected zones, so it's some difference between those and objective/finishing notes being placed on the map. I'll try and merge your new PR today or tomorrow so I can take a closer look at this.

VideoPlayerCode:
Ahh, I see! I missed the part about _"And this does not affect quest start notes, only quest finish and quest objective notes."_

I think I've seen such misplaced "quest finishing/objective" notes on the world map. I never really paid attention to that when questing (I mainly need Questie to be able to see the "!" questgivers, so that I don't miss out on any quest starters).

That's strange... I wonder what could cause the finishing-notes to move to other locations. Unfortunately I don't have time to help debug that, but hopefully it's something really trivial.

VideoPlayerCode:
Btw, regarding the idea of one day merging the vanilla and TBC code into a single tree, with some per-client checks at runtime (to call different APIs if needed), you can simply do this:

```lua
tonumber(GetBuildInfo():sub(1,1)) -- 2 for TBC, I guess 1 for Vanilla (but haven't tested in Vanilla)
```

Issue #522, opened 2018-11-19
Astrolabe.lua


AMGarkin:
**astrolabe/Astrolabe.lua**

_lines 184-201:_
```lua
    if (not WorldMapFrame:IsVisible() == nil) then
        return
    else
        return Astrolabe_LastC, Astrolabe_LastZ, Astrolabe_LastX, Astrolabe_LastY
    end
    if (WorldMapFrame:IsVisible() == nil) then
        SetMapToCurrentZone()
        x, y = GetPlayerMapPosition("player")
        if (x <= 0 and y <= 0) then
            SetMapZoom(GetCurrentMapContinent())
            x, y = GetPlayerMapPosition("player")
            if (x <= 0 and y <= 0) then
                return
            end
        end
    else
        return Astrolabe_LastC, Astrolabe_LastZ, Astrolabe_LastX, Astrolabe_LastY
    end
```
These 2 conditions one after another doesn't make sense. Function will always end in the first one.
The first condition returns nil when coords are less then 0 when world map is open and last coordinates when map is closed. The second one tries to update player coordinates when map is closed and returns last coordinates when map is open. I believe that there should be just the second condition and the first should be deleted or commented out.
And maybe use `if (not WorldMapFrame:IsVisible()) then` instead of `if (WorldMapFrame:IsVisible() == nil) then` because return value from IsVisible() should be boolean.

_lines 275, 276:_
```lua
    Astrolabe.ForceNextUpdate = true;
    Astrolabe:UpdateMinimapIconPositions();
```
When you call UpdateMinimapIconPositions(), you should also reset update timer, otherwise you can have 2 updates in a row. My suggestion is:
```lua
    self.ForceNextUpdate = true
    self.UpdateTimer = self.MinimapUpdateTime
    self:UpdateMinimapIconPositions();
```

_lines 528-531:_
```lua
    for Z in ipairs(zones) do
        SetMapZoom(C, Z);
        zones[Z] = {mapFile = GetMapInfo(), mapName = N}
    end
```
There is missing `N` in for loop, so `mapName` will be always nil. Correct is: `for Z, N in ipairs(zones) do`

AMGarkin:
[...]

PR #524, opened 2018-11-20
Astrolabe, issue #522

AMGarkin:
- bumped version to 19 (here is 17, TomTom Vanilla has 18)
- removed unecesssary condition in function Astrolabe:GetCurrentPlayerPosition()
- update timer reset in function Astrolabe:PlaceIconOnMinimap(...) in order to prevent excessive updating of minimap icons
- fixed issue that Astrolabe.ContinentList[continent][zone].mapName was always nil
- reduce rate of memory consumption (clearing arrays, removed unpack() - by Road-block, see https://github.com/cralor/TomTomVanilla/commit/b76ae2c9da1564ffe3218ef611ff8c315833f03f)

### DIFF
--- a/!Questie/astrolabe/Astrolabe.lua
+++ b/!Questie/astrolabe/Astrolabe.lua
@@ -1,6 +1,6 @@
 --[[
 Name: Astrolabe
-Revision: $Rev: 17 $
+Revision: $Rev: 19 $
 $Date: 2006-11-26 09:36:31 +0100 (So, 26 Nov 2006) $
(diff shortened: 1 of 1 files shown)


Muehe:
@AMGarkin Merged, thank you! :+1:

The minimap icons have a noticeable lag at riding speed or above, but I guess the saved memory and all the other fixes are worth it.

Just one thing for next time (if there is any) please make more commits. The diff was a bit hard to digest. Things like separating indentation fixes and code changes, one commit per change, etc. would be nice.

PR #552, opened 2019-05-26
Questie classic

sinkda:
Saw in the Questie Discord a request for placing a quest market on the mobs nameplates that are objectives for quests.  Took a shot at it and working on my Lua / addon skills.  Uses the QuestieTooltip data to add a small ! icon to the left of the nameplates for all creatures that are objectives. 

Current FramePool didn't seem to support so wrote a quick new one.  Some cleanup required.

### DIFF
--- a/Modules/QuestieNameplate.lua
+++ b/Modules/QuestieNameplate.lua
@@ -4,6 +4,8 @@ local activeGUIDs = {};
 local npFrames = {};
 local npFramesCount = 0;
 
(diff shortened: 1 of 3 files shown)


sinkda:
image

Preview image

AeroScripts:
Looks good, I'll test shortly and pull this. We're going to be slightly redoing the tooltip logic soon so it might benefit from that later, along with a way to easily get the note's texture, but not 100% sure when that will happen

Thanks for the PR 👍

PR #582, opened 2019-06-03
Fade over player minimap

sinkda:
Updated logic to be able to fade the icons over the player on the minimap too to avoid hiding in-game nodes (such as if an herb or ore node is under an objective icon).    This can be enabled / disabled from the options. 

image


AeroScripts:
I like it! Though, we should probably also figure out a way to make notes render underneath base game icons like player / group etc. I think that's possible, but this is definitely better for now

PR #589, opened 2019-06-06
Update to Config Window and Options

sinkda:
Updated the config window and options. This commit includes:

1.) Added remap function for "Icon of Player" fade smoothing
2.) Added ability to configure distance and fade amount of icons over player
3a.) Update the minimap icon tooltip and toggling
3b.) Updated AceConfigDialog library to include an onclose callback setter to updating the toggle variable to avoid toggle inconsistencies. 
4.) Updated config window to have better defined category tabs and spacing.


sinkda:
Fix to icon overlapping with nameplate icons:

PR #598, opened 2019-06-07
Ability to set map coordinate precision

sinkda:
As suggested by user Shelfen on Discord.  Ability to choose the decimal precision for the coordinates of player and cursor on the map (not minimap due to space limitations)

### DIFF
--- a/Questie.lua
+++ b/Questie.lua
@@ -126,6 +126,7 @@ _QuestieOptions.defaults = {
 	  nameplateEnabled = true,
 	  minimapCoordinatesEnabled = true,
 	  mapCoordinatesEnabled = true,
+	  mapCoordinatePrecision = 1,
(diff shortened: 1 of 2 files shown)


Questie.lua
```diff
@@ -623,10 +638,10 @@ local options = {
 					name = "Reset Questie",
 				},
 				Spacer_B = _QuestieOptions:Spacer(12),
-				github_text = {
+				reset_text = {
```
wallrik: Is this change intended for this pull request?

sinkda: Yes this change is intended as it's a copy/paste error from the original commit.  The table key for "GitHub_text" is already used below so it was ts that texnot showing this text.  by changing it to "reset_text" it let show and not be overwritten.

PR #610, opened 2019-06-17
Journey System (And Quest Searching)

sinkda:
Initial drop of the Journey system and quest searching / sorting.  Some Localization fixes and a few QoL methods implemented for UI styling and coloring.


Questie.lua
```diff
@@ -872,10 +876,22 @@ local minimapIconLDB = LibStub("LibDataBroker-1.1"):NewDataObject("MinimapIcon",
 			end
 
 			_QuestieOptions.OpenConfigWindow()
+
+			if QuestieJourney:JounreyWinodwShown() then
```
Schaka: type-o!

PR #612, opened 2019-06-19
Updates to Journey System

sinkda:
Fixed Type-o in function name
Added Minimap toggle slash command: /questie minimap
Added dungeon quests (that start *INSIDE* a dungeon, not all quests for that dungeon. This also fixed NPC searches for mobs located inside dungeons
Now clicking on a quest name inside of the "Also Starts / Also Finishes" areas will bring you to the search for that quest and auto-select that quest for you.

### DIFF
--- a/Modules/QuestieJourney.lua
+++ b/Modules/QuestieJourney.lua
@@ -4,7 +4,7 @@ local AceGUI = LibStub("AceGUI-3.0");
 local journeyFrame = {};
 local isWindowShown = false;
 local lastOpenWindow = "journey";
-local containerCache = nil;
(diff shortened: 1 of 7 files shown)


Modules/QuestieJourney.lua
```diff
@@ -1379,7 +1379,7 @@ function QuestieJourney:toggleJourneyWindow()
     end
 end
 
-function QuestieJourney:JounreyWinodwShown()
+function QuestieJourney:JounreyWindowShown()
```
Schaka: Journey, not Jounrey, right? :P

PR #634, opened 2019-08-11
Streamline german locale config with english version

Hamdor:
There where some typos and missing entries in the german locale config.
For example 'LOWLEVEL_ABOVE' was wrong. For now, this at least shows the correct english texts in the UI.


AeroScripts:
Hey, thanks for noticing this. We should probably copy in the latest version to all the other languages too, I don't think any translations have been done yet

Muehe:
FYI, we will look into all open PRs once the servers are up again.

PR #637, opened 2019-08-11
Fix other language missing quest

bmagic:
When nothing is found in DB with ObjectiveDesc, fill it with current Objective from ObjectiveData. 
Maybe the order is wrong bug it will work.

### DIFF
--- a/Modules/QuestieQuest.lua
+++ b/Modules/QuestieQuest.lua
@@ -744,6 +744,9 @@ function QuestieQuest:GetAllQuestObjectives(Quest)
                             Quest.Objectives[i].Id = v.Id
                             Quest.Objectives[i].Coordinates = v.Coordinates
                             v.ObjectiveRef = Quest.Objectives[i]
+                        else
+                            -- When nothing is found (other languages) fill it.
+                            Quest.Objectives[i].Id = Quest.ObjectiveData[i].Id
                         end
                     end
                 end


AeroScripts:
hey thanks for this, working good for you? I'm unable to test right now

bmagic:
I am playing with this change in French. I think this will not impact English players but at least allows compatibility with other languages.

AeroScripts:
I will be merging this and other PRs today, since beta is back up I can properly test myself. Thanks again for the contribution ❤️

bmagic:
I think to be sure that it does not have an impact we would have to add this:
Quest.ObjectiveData[i] ~= nil 

Do you want a new PR ?

AeroScripts:
Yeah thats probably a good idea to add. You can create a new PR if you prefer or add it to this one, ill merge tonight and test, sorry for the delay once again, been a crazy couple days

AeroScripts:
Just did some testing, and there's a few cases where this will cause issues in it's current form

Basically, sometimes the first objective doesn't match correctly on some quests (normal behavior, its rare) so the logic needs to be applied only when no match has been found (after the ObjectiveData loop is finished)

Something like this
```Lua
-- try to find npc/item/event ID
for k, v in pairs(Quest.ObjectiveData) do
    if objectiveType == v.Type then
        -- TODO: use string distance to find closest, dont rely on exact match
        if ((v.Name == nil or objectiveDesc == nil) and v.Type ~= "item" and v.Type ~= "monster") or (v.Name and ((string.lower(objectiveDesc) == string.lower(v.Name))) or (v.Text and (string.lower(objectiveDesc) == string.lower(v.Text)))) then
            Quest.Objectives[i].Id = v.Id
            Quest.Objectives[i].Coordinates = v.Coordinates
            v.ObjectiveRef = Quest.Objectives[i]
        end
    end
end

-- 2nd pass (fix for missing language data)
if Quest.Objectives[i].Id == nil then
    for k,v in pairs(Quest.ObjectiveData) do
        if objectiveType == v.Type then
            -- When nothing is found (other languages) fill it.
            Quest.Objectives[i].Id = v.Id
        end
    end
end
```
should do the trick. I'm going to test with that code shortly, if you don't mind modifying the PR I'll merge.

bmagic:
I will create a new one with your code. I cannot test it because I haven't access to the beta.

PR #644, opened 2019-08-16
Quick fix for other languages

bmagic:
See #637 
Fix for other languages. When the ObjectiveData doesn't exist, fill it with default objectiveData.


AeroScripts:
Looks good <3

AeroScripts:
I tested this myself last night and it seemed fine, but we should still keep an eye on objective updates

Issue #562, opened 2019-05-30
No Quest at marker for Guarded Thunder Ale Barrel


Lickwid:
Searched for a good 15 minutes, there no quest at the quest marker. Maybe I just missed it, but was standing on the marker for 5 minutes. Tried getting to it from the outside and you can't climb the mountains.

Edit: I just got the quest Bitter Rivals (https://www.wowhead.com/quest=310/bitter-rivals), and it appears to be on the same marker as Guarded Thunder Ale Barrel, possibly the start marker is wrong?

sinkda:
Just to confirm, you saw the quest to get the Guarded Thunder Ale Barrel but you didn't have the quest for Bitter Rivals yet?  I know this quest makes you buy the thunder ale for the dwarf in the inn's basement and he walks away. then you interact with the barrel behind him but in-game there is no quest marker on the barrel itself.  you have to mouse over it and it highlights then you can interact with that quest.

Lickwid:
Yep. I walked into town and started picking up quests, couldn't get that one. Didn't realize what it was until I was completing Bitter Rivals, and then I put two and two together.

PR #650, opened 2019-08-24
Fix #641

### LINKED ISSUE #641: Questie for 1.13 client quest issue: Guarded Thunder Ale Barrel
Guarded Thunder Ale Barrel is a quest in Dwarf starting zone that is not being treated properly by Questie. 

It is a optional quest which allows you to complete another quest: Unguarded Thunder Ale Barrel.

If someone has already complete the Guarded Thunder Ale Barrel quest recently, you are able to complete the Unguarded quest without doing the Guarded quest. After doing so, you are not able to do the Guarded quest, but it remains on the map and can't be removed.

I think Guarded Thunder Ale Barrel should be auto completed if Unguarded Thunder Ale Barrel is completed in order to avoid this issue.

BreakBB:
The [Guarded Thunderbrew Barrel](https://classic.wowhead.com/quest=403/guarded-thunderbrew-barrel) is an optional quest which can't be done if the [Return to Marleth](https://classic.wowhead.com/quest=311/return-to-marleth) quest has been done before.

This PR adds the quest IDs to the questExclusiveGroupFixes.

### DIFF
--- a/Database/corrections.lua
+++ b/Database/corrections.lua
@@ -17,7 +17,8 @@ questExclusiveGroupFixes = {
     [1943] = {1944}, -- mage robe breadcrumb
     [526] = {322,324}, -- not 100% sure on this one but it seems lightforge ingots is optional, block it after completing subsequent steps (#587)
     [3765] = {1275}, -- corruption abroad breadcrumb
-    [164] = {95} -- deliveries to sven is a breadcrumb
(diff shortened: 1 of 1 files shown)


AeroScripts:
Looks good <3 ill merge once classic is up

BreakBB:
This might fix #562 as well.

PR #665, opened 2019-08-28
Use upstream HereBeDragons

Nevcairiel:
Upstream supports Classic now, and this avoids duplicating resources.
HereBeDragons-Migrate is left untouched, due to the data changes made in it, and since I was not sure if you still need those. Ideally it should not be loaded when its not needed, as it just consumes resources.


AeroScripts:
I'll test this shortly and then merge, thanks again for the help :D 

I'll have to test with other HBD addons though, one of the issues we had in beta was a broken copy of HBD loading instead of the version questie came with. Shouldn't be a problem now that there's an official version though

AeroScripts:
Merging this now, As for the changes in HBD-Migrate, it was quite a while ago I'll have to look through the code but I'm sure there's a way we can make it work with the unmodified one

Nevcairiel:
I saw you rolled this back on some claim of conflict with Zygor.

But, I downloaded the latest version of Zygor and they use renamed HBD libraries and renamed pin templates. They should absolutely not conflict with anything. What exactly are you seeing?

AeroScripts:
Several users reported not having any world map icons (minimap icons worked fine)

I got in contact with one of them on discord and they sent me their copy of zygor's and I was able to reproduce the issue. I'm not sure what is happening, but the issue seems to be having the same pin template name, despite different ace library names

Zygor (the copy I have) uses HereBeDragons-ZYG for the lib name, but the regular HereBeDragonsPinTemplate for notes

Do you have a discord? add aero#1357 I can send you the copy I have. At first I thought it was nothing, or just some thing 1 person had but a bunch of people reported it.

AeroScripts:
Maybe something to do with RemoveAllPinsByTemplate / EnumerateAllPinsByTemplate? Something is making zygor think questie pins belong to it (That's my guess anyway)

Nevcairiel:
I checked it out just now from the Zygor client, so either they fixed it in the last day or so, or I don't know. Definitely had different template names as well.

Nevcairiel:
In fact I checked out the Retail version, and that still has the original pin name in it, so they must've fixed it quite recently.

AeroScripts:
Hmm, well maybe best bet to leave it like this for now because it does prevent issues, until things are sorted out

I'd prefer to return to the standardized names in the long term, going to keep it like this for now as a lot of people seem to have issues (It may be other addons messing with it also)

If you're curious, here's the HereBeDragons-Pins-2.0.lua from the copy of zygor's that user sent me:
https://hastebin.com/iwesotihac.lua

PR #670, opened 2019-08-28
Fix russian locale

243083df:


AeroScripts:
Can't verify, but assuming this is correct, thanks for the fixes!

PR #680, opened 2019-08-29
[WIP] Add language lookup data

BreakBB:
This PR adds the first lookup tables for languages different to English. This lookup tables will allow to show tooltips and target icons for more languages than just English.

Languages included are:

- English
- German
- French
- Spanish
- Russian
- Chinese

Those tables include language specific data about the NPC names and quest data.

NPC table:

```lua
LangNameLookup = {
  [id] = {"NPC name"},
  ...
}
```

Quest table:

```lua
LangQuestLookup = {
  [id] = {"Quest title", "Quest description", "Quest objective"},
  ...
}
```

Some quests have their description and/or objective missing, because the Wowhead data isn't complete yet.


bmagic:
Maybe for enUs you need to push to empty array ?
LangNameLookup = {}
LangQuestLookup = {}

BreakBB:
Since I am new to addon development I can't even tell how to make use of the data. You think the English data is not needed? 🤔 Well it is inside the questDB.lua file, but not sure if it clever to have the stuff at different places!

Will update the files later this day and might include Chinese as well. Found some great improvements after a talk with Muehe yesterday 👍

bmagic:
I just finished 50% of the job to use your data. I'm probably going to do a PR tonight.

BreakBB:
Sounds good. I need some more time though to update the data. The quest data in this PR is severe corrupted. A lot of stuff slipped through my filters and the new filters are not messing with the separation of description and objective.
Hopefully I will make it tomorrow!

bmagic:
WoWScrnShot_083019_212119

Sound good !!! 
Is it possible not to include data when there is no translation?

bmagic:
First returns:
- I do not need the long description (LangQuestLookup [2]) or I do not see where it is used.
- Is it possible to have the translation of the items? https://github.com/bmagic/QuestieDev/blob/master/Database/TEMP_questie4items.lua
- Is it possible to have the translation of objects? https://github.com/bmagic/QuestieDev/blob/master/Database/objectDB.lua

The ideal would be not to put the entry at all if he does not have a translation.

Muehe:
@bmagic 

* Description can be used to identify quests that have the same title.
* Item translations are possible, but this PR only includes quests and NPCs due to the scrapper being WIP.
* Same.

@BreakBB I agree that untranslated texts should default to `nil`. Maybe the Gatherer object we discussed can be helpful with this (it had a `name_dede` field, quests with missing translations might not have it).

bmagic:
I have 3 fields in the lookup.lua file.
```
[5] = {
'L\'estomac qui gargouille',
'Je suis resté caché dans cette ville fantôme depuis des semaines. Apportez-moi de quoi festoyer et je vous payerai généreusement. J\'ai entendu parler d\'un certain chef Grual qui fait de succulents gâteaux au crabe brun, à la taverne du Corbeau écarlate à Darkshire...',
'Parler au chef Grual.'
},
```
[1] : Name 
[2]: LongDescription
[3]: Description 

I don't see where the field LongDescription is used in current Questie code.
Name & Description it's ok for me.

bmagic:
I will do a PR with the data as is. But for the moment I play without worries in French (except for object & item missing data)

Muehe:
Ok, but if you don't mind, please remove the line breaks. One quest/npc/etc. per line.

bmagic:
Yes it was just to be clearer, I use the datas made by @BreakBB

BreakBB:
Sounds good.

I need some more time though to update the data. The quest data in this PR is severe corrupted. A lot of stuff slipped through my filters and the new filters are not messing with the separation of description and objective.

Hopefully I will make it tomorrow!

BreakBB:
As I just arrived  at home I will have a short look at the code, but I will add my progress tomorrow!

BreakBB:
The last commit should remove at least most of the corruptions.

Moreover I made two changes to it:

1. Missing descriptions are now `nil` instead of an empty string
2. Quests which haven't been translated from English are `nil` as well now instead of the English fallback

bmagic:
Can not you just not put them in order to save space? 

Can you name the variable in each lockup.lua like this : LangNameLookup['frFR'] or LangQuestLookup['deDE'] ?

BreakBB:
Sorry but I don't understand your first sentence 🤔 

And yes I can rename them!

bmagic:
No, but forget my first sentence. In the end it's perfect with the "nil" in the tables.

BreakBB:
So the changes are good for now? :)

BreakBB:
@bmagic Btw did you add a fallback to English for data that is `nil`? Didn't watched over your changes (and can't even tell what I understand from it) :D

bmagic:
Yes, if there is no translation it uses the original db.
But as a result there is a good chance that without translation this will not appear in Questie (on tooltip).

BreakBB:
You mean because the original db isn't complete?

bmagic:
No, it's because the tooltips are based on the mob's names. As a result, this will not be displayed because the localized client displays a different name from the DB.

BreakBB:
Oh ye that totally makes sense

merasil:
am i reading this right? The current master-branch does include a fallback to the english version of the tooltip whenever a different language than english is used in wow client?

BreakBB:
@merasil There will be a release very soon which includes this translation data. The latest version only supports English WoW clients towards tooltip and nameplate information.

With this PR and #733 the languages listed above will be supported as well.

PR #692, opened 2019-08-30
Updated ENABLE_AVAILABLE_DESC 

mattstruble:
Updated ENABLE_AVAILABLE_DESC  text to use "available quest" instead of "quest turn-in"


AeroScripts:
Nice catch!

PR #733, opened 2019-09-01
Add locale data for Quest & NPC.

bmagic:


BreakBB:
#724 can be closed with this one open, right?

PR #801, opened 2019-09-04
more translation stuff for german

vriic:
Found more stuff in the german translation. Should be nice for now :)


BreakBB:
Vielen Dank dafür 🎉

Issue #653, opened 2019-08-27
The Hunter's Path quest showing despite being completed


Painstormx:
This quest is still showing on the map on 2 trainers in Darnassus, even though the quest has already been completed in Doolanar.

BreakBB:
From the Wowhead data it looks like Blizzard changed some quests from vanilla to Classic.

[6072](https://classic.wowhead.com/quest=6072/the-hunters-path) is listed to start in IF.
[6073](https://classic.wowhead.com/quest=6073/the-hunters-path) is listed to start in SW.

In our data those quests both start on Teldrassil since those lead to Dolanaar.

My guess is that Blizzard changed the breadcrumb quests to start for each Hunter race in a capital city leading them to their starting place. Meaning nightelfes getting this quest in SW still have to go to Teldrassil.

Since multiple users stated those quests don't match with our data I think we should trust Wowhead on this, shouldn't we?

It will take some time though to go through all the 15 starting quests.

BreakBB:
This has been tested and is fixed in the current version. If something similar comes up please create a new issue.

PR #808, opened 2019-09-04
[WIP] Add possible fix for #653

BreakBB:
The "exclusiveTo" field already had the quests for once guiding to the same trainer, but missing the ones at a different capital. This PRs adds all available quests for one faction onto each quest.

I wasn't sure if only the missing IDs should be added or all of them.


Muehe:
Not sure if this should be fixed with adding the quests as exclusive or by changing the race requirements of the quest, but I suspect it's the latter. We need to check if the quests from other capitals are available until the one in the own capital is finished, or if they are never available in the first place.

BreakBB:
Right now we can't do much on this since not even the Wowhead data is correct where you can accept which quest. I suspect Blizzard changed things on Classic for these quests

Muehe:
Most of the mangos/wowhead data is from TBC times, and it's possible they made class quests available to other races only after Classic as a QOL change for people leveling together. We could level a hunter and check...

Muehe:
Ok, so I have a Orc hunter on level 10 now, he isn't able to accept quests for the Tauren pet quest, but one NPC in Thunderbluff offers a breadcrumb quest which isn't shown by Questie:


Screenshot_20191008_180438

BreakBB:
Closing since it is obsolete.

PR #851, opened 2019-09-07
Update lookup data + ptBR

BreakBB:
This PR updates the lookup data for the localization of most of the languages. Only Chinese had no new data to this point.

Also Brazilian Portuguese is introduced for the first time.


AeroScripts:
more data 🎊

PR #864, opened 2019-09-07
Fix off-by-one issue level filter.

stijnherreman:
When the player level is 10 and the minimum level filter is 1, quests with level 9 should be visible. Similarly, when the maximum level filter is 1, quests with level 11 should be visible.

I've tested this with [a level 43 quest](https://classic.wowhead.com/quest=2342/reclaimed-treasures), a level 33 character and the maximum level filter ("how many levels above your character to show") set to 10. Before the change, the quest doesn't show up. After the change, it does.


AeroScripts:
Looks good, thanks for debugging that 👍

PR #866, opened 2019-09-07
different objective color for different quests

siikamiika:
Just a proof of concept/RFC, still missing menu option to turn off. Related to https://github.com/AeroScripts/QuestieDev/issues/604. Another useful feature would be to highlight the objectives of the same quest on the map when mouse is hovered over one objective.
sbOUnxkQcd


AeroScripts:
I like it, ill merge after 4.0.18 is up (unless you want to make some more changes first)
Shouldn't be hard to add it to the config menu

siikamiika:
@AeroScripts I read the code a bit, and realized that there was already an implementation using [fastrandom](https://wow.gamepedia.com/API_fastrandom). I'll change the PR to use that instead and add a toggle button, but I might make some changes to how the colors are generated

siikamiika:
Alright, I think it's done for the world map now except for localization

AeroScripts:
Looks good 👍 thanks for your work

Issue #890, opened 2019-09-08
Soothing Turtle Bisque quest


ShabbyAbby:
## Description
Questie is showing that you have to obtaine Soothing Spieces from Tarren Mill even if you play as Alliance

## Wowhead link
https://www.wowhead.com/quest=555/deprecated-soothing-turtle-bisque#comments

## Questie version
4.0.18

haggen:
Also it's not showing the turtles (Snapjaw, ID 555) locations. even though it's in the database.

AeroScripts:
Hmm something must have messed up with the logic for this one. It was a bit of a weird fix compared to the other quests (adding spices to the nearby vendors for each faction) 
The correction code was recently refactored so something could have gone wrong with it (its currently the only quest in the game that has a faction-specific vendor drop like that)

haggen:
I just checked:

- In the quest database the item ID is correct for Turtle Meat (3712).
- In the item database Snapjaws (2408) do drop Turtle Meat, albeit the drop % is incorrect, it says 100% when this not really the case.
- In the NPC database Snapjaws are correctly located in Alterac Mountains and Hillsbrad Foothills.

I should note that the icons for Turtle Meat do show in come location over in Alterac Mountains, but they don't show for the spawns in Hillsbrad.

haggen:
Switching the lines https://github.com/AeroScripts/QuestieDev/blob/master/Database/corrections.lua#L83 with 89 fixed the issue.

PR #897, opened 2019-09-09
Switch All./Horde corrections of Soothing Turtle Bisque #890

haggen:


AeroScripts:
Looks good 👍 that was pretty dumb of me

PR #910, opened 2019-09-10
Update config.lua

Pgmbru:
Update new features for the tracker

With new layout of the translation


AeroScripts:
Looks good thanks again for the translations <3

PR #926, opened 2019-09-11
Fix description in comment

gyunter:
`=` instead of `-` for unknown faction


AeroScripts:
Nice catch, we really gotta add an option to show these spawns

PR #986, opened 2019-09-15
Add locals and title case global functions

BreakBB:
I did some first cleanups and added locals where appropriate.


BreakBB:
Switching to LogonDevelop

Issue #697, opened 2019-08-30
Being a GuildMaster and setting ranks


MaxHughes:
Getting an issue with Questie when setting ranks in my guild, it creates an error and if i click ignore, then it sets any changes to the guild master. I have to disable questie to make any changes to the guild ranks.

Muehe:
Seems familiar. Please make sure that:

1. you are using the [latest version](https://github.com/AeroScripts/QuestieDev/releases/latest) of Questie (4.0.12 currently).
2. Questie is the only addon enabled when this bug occurs.

Assuming you already created your guild we will try to reproduce on our own.

AeroScripts:
Someone else had this issue when trying to leave a group, but it stopped happening after /reload

Not sure what causes this to happen as questie doesn't touch the guild or group UI

BreakBB:
This issue was been fixed with 1f33a36d7e9c76bd7889d8a1dc3b09abc2bf804b .

PR #1009, opened 2019-09-16
Tracker improvements & fixes

Demoniko:
I noticed that context menu for quest tracker uses self-implemented EasyMenu which seems to be exact copy of API already implemented by Blizzard in FrameXML, so I replaced usage to built-in one. Reason - it works with addons that change appearance of context menus, for example ElvUI and it doesn't break anything, since API is exactly the same.

I also noticed that Drag event incorrectly checks if tracker is locked (uses logical 'or' instead of 'and', so frame is never locked) and fixed it.

Also improved context menu further - added 'notCheckable' where appropriate, replaced different text depending on state into checkbox (easier to read in my opinion) and made TomTom button only appear if you have TomTom and made menu 2 level - objectives added directly to menu (reason is that with 3 level menus 3rd level menu will overlap with 1st level menu, so it's easy to accidentally close submenu by moving mouse a bit too far and hovering over some 1st level menu button, also less mouse movement to get to objectives).

### DIFF
--- a/Locale/enUS/config.lua
+++ b/Locale/enUS/config.lua
@@ -188,6 +188,7 @@ QuestieLocale.locale['enUS'] = {
     ['TRACKER_FOCUS_OBJECTIVE'] = "Focus Objective",
     ['TRACKER_FOCUS_QUEST'] = "Focus Quest",
     ['TRACKER_CANCEL'] = "Cancel",
+    ['TRACKER_TRACK'] = "Track Quest",
     ['TRACKER_UNTRACK'] = "Untrack Quest",
     ['TRACKER_SHOW_QUESTLOG'] = "Show in Quest Log",
     ['TRACKER_SET_TOMTOM'] = "Set TomTom Target",
--- a/Modules/QuestieTracker.lua
+++ b/Modules/QuestieTracker.lua
@@ -70,7 +70,7 @@ function _QuestieTracker:StartFadeTicker()
 end
 
 local function _OnDragStart(self, button)
-    if IsControlKeyDown() or not Questie.db.global.trackerLocked then
+    if IsControlKeyDown() and not Questie.db.global.trackerLocked then
         _QuestieTracker.baseFrame:StartMoving()
     else
         if not IsMouselooking() then-- this is a HORRIBLE solution, why does MouselookStart have to break OnMouseUp (is there a MOUSE_RELEASED event that always fires?)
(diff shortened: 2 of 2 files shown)


AeroScripts:
There's a big reason why we use LibUIDropDownMenu. Without it, blizzard UIs like the guild permisions UI become tainted.

The other changes look good but we definitely cant use the base UIDropDown.

AeroScripts:
"or" instead of "and" is also intentional in that case. Unlocking the tracker allows you to move it without any keybind. Holding control lets you move it while its locked

Demoniko:
On "Lock" - this is a bit weird. If I want to move my quest tracker I can just unlock it and if I locked it - it means I don't want to move it, so why allow it at all? I have some ctrl hotkeys, so there is a decent chance I'll just accidentally move it. And I might do another pull request on this, but right now there is no way to truly reset tracker position, clicking reset button just changes setting, not removes it (which makes a difference, when you change default tracker position with different addon).

AeroScripts:
It was a compromise between the 1.12 questie way of ctrl+shift drag, and the more modern way of locking/unlocking. I don't think its a huge issue if people can drag it by holding a key while its locked, and that way it works the way people coming from 1.12 expect

AeroScripts:
As for resetting the tracker position, I Agree we need to fix how that behaves, to fully reset you need to /reload

Demoniko:
I don't think /reload works. I'm talking about situation where I changed default tracker position through other addon (for example ElvUI) and want my tracker to be at default position, but resetting doesn't truly "reset" it, it just changes position, so tracker position is not tied to default position. The difference is that before I moved tracker first time I can change tracker position using other addons (I prefer to setup layout with ElvUI), but if I ever moved it - I can no longer do that, even if I reset position. The only way to truly reset it is to remove globals from SavedVariables related to tracker position.

Demoniko:
Also can't people who expect old behavior just unlock tracker and move it with ctrl+click? It just seems very counterintuitive that locking frame doesn't really lock it and instead changes how it's moved?

AeroScripts:
```Lua
function QuestieTracker:ResetLocation()
    Questie.db.char.TrackerLocation = nil
    if _QuestieTracker.baseFrame then
        _QuestieTracker.baseFrame:SetPoint("CENTER",0,0)
        _QuestieTracker.baseFrame:Show()
    end
end
```
The only thing that saves the position is Questie.db.char.TrackerLocation, which is reset when you click the button

It's written again if the tracker is dragged after that though. Also, reset location sets it to the center of the screen (when working properly) instead of default location, because default location might not be safe. It sets it to the position of the blizzard tracker, and if something goes wrong with that there would be no getting the tracker back, center of the screen should always be safe

Not an ideal solution though I agree

Demoniko:
I see on location reset, I don't know why my TrackerLocation variable didn't reset last time I tried, I probably moved it accidentally.

Demoniko:
Also about FrameXML EasyMenu - what do you mean blizzard UI becomes tainted? I haven't encountered any issues so far after I changed it and it seems weird that using that API will break something. I did not see anything in documentation to indicate that. Unless it's a bug specific to current WoW version?

AeroScripts:
Yeah it's super strange, blizzard has an issue with it currently any addon that uses UIDropDownMenu (including AceGUI/AceConfig, thats why we have a custom copy of AceConfig also)
can cause taint, which causes errors/problems when blizzard restricted UIs use a dropdown
See #697 and https://us.forums.blizzard.com/en/wow/t/guild-master-cant-edit-guild-permissions/277494

https://www.wowace.com/projects/libuidropdownmenu

The LibUIDropDownMenu page also has some info about the bug

Demoniko:
Wait, this addon is 2 years old, do they really have this issue for 2 years and do absolutely nothing about it?

AeroScripts:
The UIDropDown bug has been around on retail for ages yeah, there are tonnes of forum posts about it too, not sure why blizzard doesnt fix it. Just using the api functions completely normally still causes taint, all we do is standard AceGUI stuff, no weird hooks or anything like that.

Wasn't an issue in 1.12 afaik but a lot of other stuff was, that bug is one of the few that were brought in with the modern api

Demoniko:
Was there reason to have LQuestie prefix instead of L? I looked into ElvUI and it seems it skins LibUIDropDownMenu and NoTaint_UIDropDownMenu.

AeroScripts:
The comments section on the mod page, some people were having issues due to older/bad copies of the lib being loaded (from a different addon). Renaming it just makes sure it always loads our copy of the code

There's probably a better way of doing it though or at least a way we can make it work with things like elvui

Demoniko:
Ok, so I reverted change to Lock tracker and LibUIDropDownMenu -> UIDropDownMenu. I still propose changing lock function to actually lock frame or at least add an option for that.

Demoniko:
For now I guess I'll just add `if not S.LQuestie_UIDropDownMenuSkinned then S:SkinLibDropDownMenu('LQuestie') end -- LibUIDropDownMenuQuestie` to my local ElvUI until there is a better solution.

BreakBB:
Closing since this PR is outdated.

PR #1052, opened 2019-09-18
Added localization for Germany for the seach 'Quest per Zone'

LasterOfDesaster:
Since not knowing the german name of the zone I want to look for non-completet quests is kinda frustrating, I inserted the german names of the zones. 

May my first pull request contribution be something helpful! :D

### DIFF
--- a/Modules/QuestieJourney.lua
+++ b/Modules/QuestieJourney.lua
@@ -523,6 +523,82 @@ local zoneTable = {
     }
 };
 
+if (GetLocale() == "deDE") then
+	zoneTable = {
+		[1] = {
+			[36] = "Alteracgebirge",
+			[45] = "Arathihochland",
+			[3] = "Verwüstet Lande",
+			[4] = "Verwüstee Lande",
+			[46] = "Brennende Steppe",
+			[41] = "Gebirgspass der Totenwinde",
+			[2257] = "Tiefenbahn",
+			[1] = "Dun Morogh",
+			[10] = "Dämmerwald",
+			[139] = "Östliche Pestländer",
+			[12] = "Wald von Elwynn",
+			[267] = "Vorgebirge von Hillsbrad",
+			[1537] = "Ironforge",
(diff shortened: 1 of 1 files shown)


Modules/QuestieJourney.lua
```diff
@@ -523,6 +523,82 @@ local zoneTable = {
     }
 };
 
+if (GetLocale() == "deDE") then
+	zoneTable = {
+		[1] = {
+			[36] = "Alteracgebirge",
+			[45] = "Arathihochland",
+			[3] = "Verwüstet Lande",
+			[4] = "Verwüstee Lande",
```
BreakBB:
```suggestion
			[4] = "Verwüstete Lande",
```

Modules/QuestieJourney.lua
```diff
@@ -523,6 +523,82 @@ local zoneTable = {
     }
 };
 
+if (GetLocale() == "deDE") then
+	zoneTable = {
+		[1] = {
+			[36] = "Alteracgebirge",
+			[45] = "Arathihochland",
+			[3] = "Verwüstet Lande",
```
BreakBB:
```suggestion
			[3] = "Ödland",
```

Modules/QuestieJourney.lua
```diff
@@ -523,6 +523,82 @@ local zoneTable = {
...
+			[141] = "Teldrassil",
+			[17] = "Brachland",
+			[400] = "Tausend Nadeln",
+			[1638] = "Thunder Bluff",
+			[490] = "Un'Goro Krater",
+			[618] = "Winterquell"
+		},
+		[3] = {
+			[2437] = "Flammenschlund",
+			[1581] = "Die Todesmienen",
```
BreakBB:
```suggestion
			[1581] = "Die Todesminen",
```

BreakBB:
Thanks, we should really improve our journey and this is a great start :+1:

PR #1101, opened 2019-09-22
Performance Optimization. Reduce CPU usage from 100ms per sec to 10 ms per sec

euloanty:
Thank you!

### DIFF
--- a/Modules/QuestieCoordinates.lua
+++ b/Modules/QuestieCoordinates.lua
@@ -88,6 +88,7 @@ function QuestieCoords:Initialize()
 end
 
 function QuestieCoords:Update(elapsed)
+	Questie:SendMessage("QUESITE_ICON_ON_UPDATE")
     if (Questie.db.global.minimapCoordinatesEnabled) or
         (Questie.db.global.mapCoordinatesEnabled) then
 
--- a/Modules/QuestieNameplate.lua
+++ b/Modules/QuestieNameplate.lua
@@ -5,14 +5,9 @@ local npFrames = {};
 local npUnusedFrames = {};
 local npFramesCount = 0;
 
-QuestieNameplate.GlobalFrame = nil;
-
 -- Initializer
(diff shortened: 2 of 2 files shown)


Modules/QuestieCoordinates.lua
```diff
@@ -88,6 +88,7 @@ function QuestieCoords:Initialize()
 end
 
 function QuestieCoords:Update(elapsed)
+	Questie:SendMessage("QUESITE_ICON_ON_UPDATE")
```
BreakBB:
```suggestion
	Questie:SendMessage("QUESTIE_ICON_ON_UPDATE")
```

Modules/QuestieNameplate.lua
```diff
@@ -5,14 +5,9 @@ local npFrames = {};
...
 
-QuestieNameplate.GlobalFrame = nil;
-
 -- Initializer
 function QuestieNameplate:Initialize()
-    if QuestieNameplate.GlobalFrame == nil then
-        QuestieNameplate.GlobalFrame = CreateFrame("Frame");
-        QuestieNameplate.GlobalFrame:SetScript("OnUpdate", QuestieNameplate.UpdateNameplate);
-    end
+	Questie:RegisterMessage("QUESITE_ICON_ON_UPDATE",QuestieNameplate.UpdateNameplate)
```
BreakBB:
```suggestion
	Questie:RegisterMessage("QUESTIE_ICON_ON_UPDATE",QuestieNameplate.UpdateNameplate)
```

euloanty: No GlobalFrame in the code anymore. this needs to be removed.

Logonz:
Hey!
This is already fixed in the v4.1.2 branch.
Also we don't really want to put a message from QuestieComms -> QuestieNameplates as that would create a dependency between them.

Thanks for the interest!

Issue #1033, opened 2019-09-18
Quest Log doesn't close when pressing the escape key


DookieSkid:
## Bug description

Opening the quest log with the tracker shortcut works, but it doesn't close when pressing the escape key. Also other panels such as the character info sheet and talents page when opened cover the quest log.

## Screenshots

![WoWScrnShot_091719_214221](https://user-images.githubusercontent.com/905921/65110201-efeaae80-d995-11e9-9bb9-afa2fc57b9d0.jpg)
![WoWScrnShot_091719_214336](https://user-images.githubusercontent.com/905921/65110211-f1b47200-d995-11e9-96be-86b7e6ba1ae6.jpg)

## Questie version

4.1.1

drejjmit:
I don't have this bug on 4.1.1. Did you do a full client restart after updating?

DookieSkid:
Yes. It only happens when I shift + left click the tracker to open the quest log. Otherwise opening the quest log normally with a keyboard shortcut or the micro menu doesn't cause the problem at all.
I've also tried disabling all other addons but Questie, and the issue still occurs.

drejjmit:
Tested shift clicking the tracker and still can't reproduce this bug. Maybe there's an addon compability issue. Can you try with just Questie activated? If it works, try enabling a few addons and see if you can find what addon cause problems.

DookieSkid:
Odd. I've tried with only Questie enabled and got the same issues. I tried with a fresh install of Classic, only enabling Questie and I still got the same problems.

BreakBB:
Yes this is an issue. After using either the "Show in questlog" option or "Show on map" the escape button does not work anymore. Moreover the questlog can overlap with other frames like stated above.

We're investigating this.

BreakBB:
Fixed in the release of v4.1.2 https://github.com/AeroScripts/QuestieDev/releases/tag/v4.1.2

PR #1106, opened 2019-09-22
Fixed quest log not closing on escape press.

Spencer0:
### Issue #1033
Currently, when a user opens the quest log via shift clicking the QuestieTracker, they are unable to close it using the escape button after. This pull request address that.

### DIFF
--- a/Modules/QuestieTracker.lua
+++ b/Modules/QuestieTracker.lua
@@ -200,7 +200,7 @@ local function _ShowQuestLog(Quest)
             QuestLogExFrameMaximizeButton:GetScript("OnClick")(QuestLogExFrameMaximizeButton)
         end
     else
-       ToggleQuestLog();
+        ToggleQuestLog();
     end    
     SelectQuestLogEntry(GetQuestLogIndexByID(Quest.Id))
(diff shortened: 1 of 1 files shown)


Modules/QuestieTracker.lua
```diff
@@ -200,7 +200,7 @@ local function _ShowQuestLog(Quest)
             QuestLogExFrameMaximizeButton:GetScript("OnClick")(QuestLogExFrameMaximizeButton)
         end
     else
-        QuestLogFrame:Show()
+       ToggleQuestLog();
```
BreakBB:
```suggestion
        ToggleQuestLog();
```

BreakBB: Such an easy fix, thanks for this!

Spencer0: Anytime! Glad to contribute to something so useful, i'll be around more now that i finally hit 60.

BreakBB: Haha, sounds good! We appreciate any help we get. @AeroScripts and I are around level 30 😅

PR #1107, opened 2019-09-22
the icons in the names of plates are not shown, and the mission information when hovering over the target is not displayed, so changes for the esMX language are added

icomputo:
the icons in the names of plates are not shown, and the mission information when hovering over the target is not displayed, so changes for the esMX language are added, test in wow Classic.


BreakBB:
Thanks for your effort! We will look into this shortly. Just curious: Where did you get the localization data from 😁

icomputo:
it's in the wow classic directories

BreakBB:
@icomputo Could you please update this PR so we can merge it? :)

PR #1121, opened 2019-09-24
mark quest "[90] Seasoned Wolf Kabobs" as a cooking quest

liweinan0423:
Reference see https://wow.gamepedia.com/Seasoned_Wolf_Kabobs


BreakBB:
Hey thanks for this fix! We're indeed missing this information.

But we don't edit the DB files directly but add changes to the `corrections.lua`. Please move your fix there :smiley:

liweinan0423:
submitted another PR #1123

PR #1157, opened 2019-09-27
Bungle in the Jungle / March of the Silithid pre req

lphua:
First timer, but I believe this fixes
https://github.com/AeroScripts/QuestieDev/issues/942

not:
https://github.com/AeroScripts/QuestieDev/commit/23f0f916ea77ffcc2c5cf9e4f93df2c09eef8fc4


lphua:
[4496] = {"Bungle in the Jungle",
[4493] = {"March of the Silithid",
[4494] = {"March of the Silithid",

4493 (alliance version) or 4494 (horde version) are pre-reqs to 4496

BreakBB:
Actually this is not as easy to fix as other quests because you have to finish quest 4493 OR 4494 and we don't have a database field for this yet.

preQuestSinlge and preQuestGroup are both for AND connections between preQuests.

Database/corrections.lua
```diff
@@ -326,6 +326,9 @@ QuestieCorrections.questFixes = {
     [1085] = {
         [QuestieDB.questKeys.preQuestSingle] = {1070},
     },
+    [4496] = {
+        [QuestieDB.questKeys.preQuestGroup] = {4493, 4494},
```
BreakBB:
```suggestion
        [QuestieDB.questKeys.preQuestSingle] = {4493,4494},
```

I've updated our requirement logic so that the preQuestSingle field is actually used for what it is supposed to.

BreakBB:
@lphua Could you have a look at the changes I suggested so we can merge this?

BreakBB:
I added your changes in 83fc2af1d014c8f46ccd20f22b1703596a4b99be

PR #1031, opened 2019-09-17
Update config.lua

Pgmbru:
Big update of the file ptBR

PR #1185, opened 2019-09-30
Update config.lua

Pgmbru:
Add latest commands more translation


BreakBB:
Thanks for this! Could you also have a look at #1031 to add the changes requested there? :)

PR #1212, opened 2019-10-02
Questie dev xp

LasterOfDesaster:
Added database for Quest and XP from WoWHead and print them on tracker, minimap and worldmap showing how much XP this quest rewards.


Muehe:
Ok first off thank you for trying to improve Questie, it's very much appreciated. :+1: 

That being said, there are a couple of problems with the PR, I wished you would have contacted us first.

1. Changes to `questDB.lua`. This is an auto-generated file using https://github.com/Muehe/sqlua so it should never be touched. All changes there will be overwritten on the next DB extraction. Same for npcs/objects/items, which is why we use `Database/corrections.lua`.
2. Changes in `_GetColoredQuestName(self)` should be wrapped in a new config option. We aren't certain we have XP data on every existing quest (I guess), so it should also be `nil`-checked.
3. There isn't any kind of logic regarding player level. Quests give different XP depending on player level.
4. We would prefer getting XP values from the API. We are looking into that, but apparently it's only possible for quests that are already in the log.
5. There is already https://github.com/BreakBB/wowhead_scraper which we use to get localisation data. **If** we decide to use wowhead data, that project should be extended.
6. Nobody is really sure how the XP system works in Classic and the data on wowhead is probably from TBC times, so wrong in many cases.

In summary, a lot more research is needed before we progress on this feature.

I hope my reply doesn't discourage you though, the first three points are easily fixed. But maybe discuss other changes here or on Discord first.

LasterOfDesaster:
Hi @Muehe!

I was on vacation, sooo my answer has delayed.

To your points:
1.) I used `questDB.lua` for the data since I wasn't able to include a extra file in the .toc file. I always had an error that told me that the variable `QuestieDB.questXpData` is a `nil`-value. This was because I din't knew I had to restart the WoW client when making a change to the .toc file. So I will create a new PR where I store the data in a seperate file.
2.) That is a greate idea. My fault of not thinking of this. I will change my code to do this.
3.) I am confused. I wrote in my PR that the raw-data amount of XP gets displayed. If my PR gets accespted I will further create logic to increase or decrease the displayed amount based on the relation of player level and quest level.
4.) That was my problem too. There are some API functions where I input a Quest ID and get information about the quest such as title, description, text, level, ... But no API gives me the XP. There IS one function that tells me the xp of the last quest I accepted. This way the quest Addon `QuestXP` collectes data about each quest the player accepts and can show you the xp in the quest log. But when logging off and on again and applying the function, I get no "real" result (nil?).
5.) My favorite point :D I made a [PR](https://github.com/BreakBB/wowhead_scraper/pull/4/) on @BreakBB 's repo where I added a crawler for the raw Xp and Quest ID, but there was no reaction on it.
6.) Your point about the XP system in Classic and the (probably) wrong data from WoWHead is another headscratch for me. I crawl the data from the [classic.wowhead.com](https://classic.wowhead.com/). The "wrong" data could be fixed with the `correction.lua` file.

Thank you for your fast response and you commitmend for this project. Its a pleasure that even within WoW you can find some dedicated programmer :D 
I'm sure this litle step-stone does not kill my motivation to contribute to this.


PS: You mentioned that I should've contaced you earlier. Why?

Muehe:
> You mentioned that I should've contaced you earlier. Why?

So we could have discussed all these points before. :wink: 

First of, summoning @Logonz because he [made a commit](https://github.com/AeroScripts/QuestieDev/commit/37c64670c6332ab0dab42c3f3a88b4782ef896b6) that added LibQuestXP on his branch (not sure what that is).

As for the points:

1.  Another file is OK, I guess, since we do not already have a place for added data from Wowhead except for translations. So if you don't mind this format would be nice for including it into our existing infrastructure:
    ```lua
    QuestieDB.extraQuestKeys = {
        ['xp'] = 1,
    }
    QuestieDB.extraQuestData = {
        [questID] = {xpAmount}, -- e.g.:
        [1234] = {100},
    }
    ```
    The thought behind putting it in a table like this is that we can later add other fields of data without changing the logic that reads it.
2.  :+1: 
3.  Just wanted to point the missing logic out, in case you weren't aware of it.
4.  As I mentioned, Logon has started looking into this, so if anybody knows more it's him.
5.  Summoning @BreakBB because he overlooked/ignored your PR. :smile: 
6. The point here was that the data available from Wowhead (as well as private servers) is mostly from TBC onwards, since there was much less data-mining during vanilla times. Just another point to be aware of.

Another point I didn't mention in my other message is what is know today about how quest XP works. There is a [QuestXP.dbc](https://wow.tools/dbc/?dbc=questxp&build=1.13.2.31882#search=&page=1) file in the 1.13 client that defines the XP a quest can get. The actual amount is [apparently](https://github.com/cmangos/issues/wiki/Quest_template#rewxpid) selected by a flag on the quest, which we do not know, so there goes that line of thinking.

Btw. if you have any questions feel free to join us on our Discord server (link in the readme).

BreakBB:
Sorry about the late response on my PR! I just accepted it. Notifications and emails of summonings don't reach me every time ^^

LasterOfDesaster:
Hi @Muehe 
thank you again for your answer. I will proceed to implement your (the project) needs into my branch.

1.) 👍 
2.) 👍 
3.) 👍 
4.) I have already made a comment on this PR
5.) I didn't mind to bother BreakBB  (:D) for this
6.) As I said, IMHO wowhead is the most trustworthy source, so I thought this would be good for using.

(7.) Your point about the flag that tells how to calculate the amount of rewarded XP: IMHO this is a problem. But as a software developer I can assure your that this will be fixed (soon) ;D 

For upcomming features I will contact you via Discord :D

Logonz:
Sorry for the late response!
The honest answer is that LibQuestXP is just a lazy way of getting the data in a good way and have it try to calculate the exp rewarded. Even if it is not 100% correct hopefully its "correct enough" especially because we use it for smaller features.

BreakBB:
Closing this since we decided to use LibQuestXP.

PR #1225, opened 2019-10-05
DBM HudMap Integration

MysticalOS:
Complete configuration already done, if user disables it, it does nothing, if DBM is not installed it does nothing (although GUI options still exist if DBM is not enabled, I hope you have an ideal solution for that maybe hide tab if DBM not installed, or gray out options)

Option defaults were chosen to be for optimal performance and minimal clutter (in other words hud has loot and slay icons off by default, leaving it's default settings to quests and interactables. Users can then elect into the higher cpu/clutter options on their own via the config)

As for feature set, well it's just hard to describe unless you try it out :D


MysticalOS:
WoWScrnShot_100519_134113
Screen Shot 2019-10-05 at 1 40 12 PM

Issue #1250, opened 2019-10-07
Map marker for [They Call Him Smiling Jim] forever stuck on the map


AnnaManiac:
## Description
Alliance starts with [James Hyal] and the followup [James Hyal]  OR  [They Call Him Smiling Jim] (you can do only one of the prerequisites).  If you did the [James Hyal] prerequisite, then the map marker for [They Call Him Smiling Jim] is stuck on the map forever.

## Wowhead link
https://classic.wowhead.com/quest=1282/they-call-him-smiling-jim

## Screenshots
![WoWScrnShot_100719_174809](https://user-images.githubusercontent.com/54857214/66351107-ff24a280-e92a-11e9-9820-52502a8f69b9.jpg)

## Questie version
Version: 4.1.2 BETA

Ben1220:
I am having this issue as well

lifeingote:
I am also having this issue.

BreakBB:
Actually as stated in #875 it should be possible to complete both quests. Can you ensure this is not the case?

BreakBB:
Fixed with 1698b260632ea658f040b25656c2b896fe798c22

Issue #875, opened 2019-09-07
They Call Him Smiling Jim (Theramore) - Not shown on map


Painstormx:
Quest is missing on map.
/reload doesn't help.

BreakBB:
Related to #815

BreakBB:
@Painstormx Could you run the following commands to check something for us? [They Call Him Smiling Jim](https://classic.wowhead.com/quest=1282/they-call-him-smiling-jim) is actually exclusive to [James Hyal](https://classic.wowhead.com/quest=1302/james-hyal) according to our data and Wowhead. So you might have completed the other one. But maybe in Classic this changed.

`/dump IsQuestFlaggedCompleted(1282)`
`/dump IsQuestFlaggedCompleted(1301)`
`/dump IsQuestFlaggedCompleted(1302)`

Painstormx:
I have already completed They Call Him Smiling Jim, but here are the results:
1282=true
1301=false
1302=true

PR #1271, opened 2019-10-11
Issue #1250, Exclusivity Issue with [They Call Him Sterling Jim]

Spencer0:
Responding to 
https://github.com/AeroScripts/QuestieDev/issues/1250

Currently, completing either #1282[They Call Him Sterling Jim] isn't properly disabling the map icon for #1302[James Hayl], nor vice versa. Updated the quest correction file.


BreakBB:
Actually as stated in #875 it should be possible to complete both quests. But if you're sure you can't you simply have to remove the corrections for boths quests. The data in `questDB.lua` is correct then.

BreakBB:
This is fixed with 1698b260632ea658f040b25656c2b896fe798c22. Thanks for your help 👍

PR #1308, opened 2019-10-13
Improvements to HUD integration

MysticalOS:
- Changed default refresh rate from 0.035 to 0.03 since it has far less jitter at 60fps with minimal cpu increase
- While at it, added a new slider that allows users to adjust refresh rate to their liking, some may need to decrease it further (higher than 60fps) or lower refresh rate for lower than 60fps situations to resolve jitter.
- Tweaked the chat print to only run once.
- Cleaned up comments


MysticalOS:
An important note is that the refresh changes won't work without current DBM-Classic dev, but the code is already inteligent that if that code isn't present the GUI just grays the option out.

PR #1378, opened 2019-10-20
Update config.lua (french)

Hantunaar:
New features Auto accept quest and Auto complete are missing in locales files. DBM HUD part isn't done because I'm not using DBM for now, still I added it for later.


Logonz:
Hey!
There are a few new strings added to Questie 5.0, if it wouldn't be too much to ask could you also update those and then we can accept the changes! :+1:

Issue #1357, opened 2019-10-18
Make Zone selector dropdown list alphabetical order for Quests by Zone at Journey


arphox:
## Description
Go to **Journey**, then click the **Quests by Zone** tab.
Select any continent, and the right dropdown list becomes active.
I would like that dropdown list to be in **alphabetical order** to be able to find a zone faster.

Screenshot of current state:
![kép](https://user-images.githubusercontent.com/17320094/67125321-d470fe80-f1f4-11e9-89d1-97cbd860e6b6.png)

_(Also I am interested in implementing this, if you have any idea how to do it, please let me know and I try to implement it by myself.)_

Spencer0:
I just saw your comment about wanting to know how to implement this, i submitted a pull request for this feature (https://github.com/AeroScripts/QuestieDev/pull/1382) so i can speak a little bit to it. 

Since we wanted to sort a collection on screen, i just went over to stack overflow and searched for sorting lua tables. From there, i looked into how that SetList function works from ACE GUI. These two links were all i really needed. 

https://stackoverflow.com/questions/2038418/associatively-sorting-a-table-by-value-in-lua

https://www.wowace.com/projects/ace3/pages/ace-gui-3-0-widgets

PR #1383, opened 2019-10-21
Feature request #1357, Make zone selector dropdown list alphabetical

Spencer0:
Implementing FR #1357, using a helper function to get a list of keys corresponding to alphabetical order, and then passing that list of keys as an optional parameter to ACE GUIs SetList function.


Spencer0:
I've written (one) unit test for this, i'm not sure if any of the devs are interested, but i merely showed the correctness of the helper function [here](https://github.com/Spencer0/QuestieUnitTests/blob/master/QuestieLuaUnitTests.lua)

PR #1403, opened 2019-10-23
Update koKR locale

39hn:
😄


BreakBB:
@39hn Thanks for this massive update to the Korean localization! :raised_hands:

Issue #1413, opened 2019-10-23
Map objective markers for "Kim'jael Indeed!" quest are not accurate


arphox:
## Description
Map objective markers (I was watching them on minimap) for the quest [**Kim'jael Indeed!**](https://classic.wowhead.com/quest=3601) are not accurate, and there are too few.
Considering the map on the wowhead page of the [**Kim'jael's Equipment**](https://classic.wowhead.com/object=153123) object, that seems a lot more accurate, but as i saw, questie does not use these locations.

## Questie version
5.0.1 2 e2e80b7

## Note
To solve this, I think the the coordinates for the `Kim'jael's Equipment` object (id=153123) should be modified in the `objectDB.lua` file.

AndrewVBragg:
I updated the list of coordinates in objectDB.lua and made a PR here: https://github.com/AeroScripts/QuestieDev/pull/1414
Just linking here so hopefully, nobody invests time solving a solved issue, :)

PR #1414, opened 2019-10-23
Issue #1413 Fix - Objective Markers for quest Kim'jael's Equipment updated!

AndrewVBragg:
I grabbed the exact coordinates from [wowhead's page](https://classic.wowhead.com/object=153123/kimjaels-equipment#map) on this quest. I just inspected the pin elements in their little image to find the values they used and popped those into the ObjectDB.lua. 
I noticed the old values were accurate to 2 decimal places, the values from wowhead only went to 1. If someone knows where I could get more accurate pinpoint locations for the map, I could easily update these values again.
_(First PR for questie btw)_


AndrewVBragg:
Oh, and just for convenience, here's a link to the issue mentioned:
https://github.com/AeroScripts/QuestieDev/issues/1413

AndrewVBragg:
@BreakBB Thanks for your comment! I updated the appropriate correction file and made a new PR.

Issue #1396, opened 2019-10-22
Seeping Corruption | Azshara | Horde


jesse-williams:
## Description
Tide pool locations not marked on either map in questie

## Wowhead link
https://classic.wowhead.com/quest=3568/seeping-corruption#comments

Coords for all 4 pools are first comment

## Questie version
4.3.0

imclerran:
PR fixing this issue [here](https://github.com/AeroScripts/QuestieDev/pull/1421#issuecomment-545860601)!

PR #1421, opened 2019-10-24
Issue #1396 Fix - Added objective markers for the Seeping Corruption quest items

imclerran:
The [quest](https://classic.wowhead.com/quest=3568) has four objectives - four items acquired at different locations in Azshara (zone 16). These items have ids [10691](https://classic.wowhead.com/item=10691), [10692](https://classic.wowhead.com/item=10692), [10693](https://classic.wowhead.com/item=10693), and [10694](https://classic.wowhead.com/item=10694) according to wowhead.

The questDB entry for quest 3568 did list these 4 objective ids, but the items were not present in the objectDB. I added the 4 items to QuestieObjectFixes, and included the `name`, `spawns`, and `zoneID` keys. Spawn locations are integer (not float) values, as this is as much detail as was available in the wowhead page [comments](https://classic.wowhead.com/quest=3568/seeping-corruption#comments).


BreakBB:
@imclerran Thanks for this fix! I currently can't check it since I am not at home. Just to make sure I understand you correctly: The objects are not at all in the `objectDB`?

imclerran:
Yep, that is correct. Anything else I should do for a new insertion to the objectDB?

BreakBB:
I had a look at the quests and I have to tell you the fix actutally belongs to the `items` and not `objects`.
Please check the `CHANGEME_<something>` item DB for those item IDs.

BreakBB:
I had a short look at our data and it looks like the items you already fixed (in the wrong place ;) ) are missing AND the objects where you get them. Those objects and their ID could be hard to find though. They should be named something like "tide pool" from what I've found out.

imclerran:
Thanks for the help -- I found the correct objects, and they are indeed in the DB. I'll update my PR, and change those to insert them in the itemDB instead.

Quick question though, I looking at the itemDB, there are 6 keys, the last of which is the `name` key. However in `QuestieItemFixes`, there are only 4 keys, the first of which appears to be `name`. Can you explain the discrepancy?  (Thanks for your help btw!)

imclerran:
Okay, I see that the item fixes appear to be inserting in an old version of the DB, which has only 4 keys. However I do not see any description of what the 4 keys are (looking at `TEMP_questie4items.lua`).

So to confirm that I'm interpreting the DB correctly, `questie4items` has the keys:
```lua
-- key names are descriptive and not useful for code purpose:
['name']
['quests']
['npc_drops']
['object_drops']
```
So if I were to add one of the items to the DB using QuestieItemFixes I would do the following:
```lua
[10691] = {"Filled Vial Labeled #1",{3568},{},{152598}} -- 3568 = quest | 152598 = tide pool 1
```

imclerran:
Closing this PR due to accidentally committing from my master branch.

BreakBB:
@imclerran You can always revert commits and mess around with them. We'll squash them into one commit before adding them to the project anyways :)

If you have another look at the file I linked above you'll see the itemDB we're currently using has only 4 keys. Those are explained in the corrections file for them as well.

And the code snipped you posted looks totally fine! :+1: It would be great if you can test the quest with those changes but I guess you already finished it.

imclerran:
@BreakBB Thanks again for the help! This is my first time contributing to open source, so thanks for the patience while I work out the kinks. I've opened a new PR [here](https://github.com/AeroScripts/QuestieDev/pull/1428), which should be all correct. 

And yeah, unfortunately I can't test the quest, but I can confirm the quest has the correct items as objectives, the items should now reference the tide pools as place to gather, and the tide pools DO have the correct coordinates.

PR #1422, opened 2019-10-24
Tooltips Update: Fixed data segmenting issue in comms code

SDPhantom:
Fix data segmenting issue that originally caused pairs() at line 150 to fail


Logonz:
This seems like a good change to instead not create an empty dataobject!
Will merge after i test it later!

Logonz:
This could potentially create issues where .title isn't set correctly for quests that have no objectives.
I don't think this solution is functional and the previous one is more inclusive while it requires more memory it won't matter most of the time because the fix just fixed an edge-case anyway.

PR #1440, opened 2019-10-27
Issue 1427 - Added dependencies for "The Binding" quest

NguyenMartin:
"The Binding" is still showing from Gan'rul Bloodeye, even if the general questline is finished 

Wowhead links
quest: https://classic.wowhead.com/quest=1471
general "the binding" quest: https://classic.wowhead.com/quest=1504

solved by adding dependencies


imclerran:
I believe this correction should be made in `Corrections/QuestieQuestFixes.lua` as mentioned [here](https://github.com/AeroScripts/QuestieDev/pull/1414).

> We don't update the DBs directly, because they're generated and using the correction files it is easier for us to upstream the changes to cmangos (the initial DB).

BreakBB:
@NguyenMartin Please use the QuestFixes file in the Corrections folder instead of changing the DB file directly.

All the DBs are generated and therefore corrections will be overwriten. Moreover with the corrections we can upstream the changes more easily to other places.

Issue #1406, opened 2019-10-23
Repeatable quests show/hide filter


QuadrozZz:
Please make the filter for repeatable quests (show and hide).

reaper666735:
this as an extra toggle option would be a nice addition indeed, even if the repeatable's already follow the "level range" filter :)

imclerran:
PR adding this feature [here](https://github.com/AeroScripts/QuestieDev/pull/1442)!

PR #1442, opened 2019-10-27
Issue #1406 Fix -- Added feature allowing repeatable quests to be toggled on or off.

imclerran:
This feature consists of two parts -- an addition to Questie's general options tab, and an adjustment to how `QuestieQuest:CalculateAvailableQuests()` works.

The addition to options is fairly trivial -- I made the decision to put it on the general tab, because it applies to both the map, and the minimap, and because there are other similar options on this tab.

The change to `CalculateAvailableQuests()` is more involved, but primarily consists of checking if a quest is repeatable, and if displaying repeatable quests is enabled. If both of these checks are true for a given quest, then the method will not filter out the quest for being hidden or completed.

Additionally `IsDoable()` previously returned false if a quest was hidden. This behavior is now slightly moddified, so that it returns false if it is BOTH hidden AND NOT repeatable. Repeatability should not affect doability.

The only potential issue here that I see is if a repeatable quest is on the blacklist for more than one reason, since repeatable quests are currently hidden by blacklisting them. In this case, the quest would be shown anyway, if the option to show repeated quests is enabled.


imclerran:
An ideal solution would be where repeatable quests are filtered from the map ONLY based on their `Repeatable` property, which is added to quest objects by `QuestieDB:GetQuest()`, based on `questDB.queatData`'s specialFlags bitmask value, and to remove repeatable quests from the blacklist. 

Since quests have this repeatable property, if that is the only reason the quest is blacklisted, it could be filtered out based on it's repeatable property, without the need to blacklist it. 

However sifting through the blacklist to determine the reason each quest has been blacklisted would be a monumental undertaking.

BreakBB:
@imclerran Thanks for this PR. The feature is missing for quite some time and should finally be added. Could you resolve the conflicts so we can test this and see if everything fits right in?

imclerran:
I'll resolve the merge conflicts when I get home. Also before testing the changes, I need to add one more commit to the PR. I'll comment here when I've added that commit.

imclerran:
Alright, should be all ready for testing!

imclerran:
Just a heads up, I set repeatable quests to be disabled by default, so you'll have to enable it in options for it to work.

Locale/frFR/config.lua
```diff
@@ -1,144 +1,146 @@
...
-    ['ENABLE_MAP_ICONS_DESC_MINIMALISTIC'] = "Active l'affichage d'icônes minimalistes.",
-    ['ENABLE_MINIMAP_ICONS'] = "Activer les icônes sur la mini-carte",
-    ['ENABLE_MINIMAP_ICONS_DESC'] = "Active l'affichage de toutes les icônes sur la mini-carte.",
-    ['HIDE_UNEXPLORED_ICONS'] = "Cacher les icônes des zones inexplorées",
-    ['HIDE_UNEXPLORED_ICONS_DESC'] = "Désactive l'affichage des icônes dans les zones inexplorées.",
-    ['ENABLE_ICON_LIMIT'] = "Limiter le nombre d'icônes",
-    ['ENABLE_ICON_LIMIT_DESC'] = "Active la limite au nombre d'icônes affichées d'un même type.",
-    ['ICON_LIMIT'] = "Limite d'icônes",
-    ['ICON_LIMIT_DESC'] = "Limite le nombre d'icônes affichées d'un même type. (Par défaut : %s)",
+		-- Onglet Options
```
BreakBB: Please fix the indentation of this file.

Locale/ptBR/config.lua
```diff
@@ -42,104 +42,106 @@ QuestieLocale.locale['ptBR'] = {
     ['ENABLE_ICON_LIMIT_DESC'] = "Enable the limit of icons drawn per type.",
     ['ICON_LIMIT'] = "Icon Limit",
     ['ICON_LIMIT_DESC'] = "Limits the amount of icons drawn per type. ( Default: %s )",
+    ['ENABLE_REPEATABLE_QUEST_ICONS'] = "Show repeatable quests on map",
+		['ENABLE_REPEATABLE_QUEST_ICONS_DESC'] = "Choose whether repeatable quests should be visible on the map and minimap.",
```
BreakBB: Please fix the indentation of this file.

Modules/Options/GeneralTab/QuestieOptionsGeneral.lua
```diff
@@ -138,14 +138,26 @@ function QuestieOptions.tabs.general:Initialize()
...
+                        width = 1.5,
                         disabled = function() return (not Questie.db.char.enabled); end,
                         get = function () return Questie.db.global.enableAvailable; end,
                         set = function (info, value)
                             Questie.db.global.enableAvailable = value
                             QuestieQuest:UpdateHiddenNotes();
                         end,
-                    }
+                    },
+                    mapRepeatableQuests = {
```
BreakBB:
```suggestion
                    showRepeatableQuests = {
```

Modules/Options/QuestieOptionsDefaults.lua
```diff
@@ -69,7 +69,8 @@ function QuestieOptionsDefaults:Load()
             trackerbindSetTomTom = 'ctrlleft',
             iconFadeLevel = 0.3,
             trackerLocked = true,
-            hideUnexploredMapIcons = false
+            hideUnexploredMapIcons = false,
+            mapRepeatableQuests = false,
```
BreakBB:
```suggestion
            showRepeatableQuests = true,
```

If we hide them by default users will be confused because it changes the current behavior!

Modules/QuestieQuest.lua
```diff
@@ -1779,22 +1781,24 @@ end
 --TODO Check that this function does what it is supposed to...
 function QuestieQuest:CalculateAvailableQuests()
     local playerLevel = QuestiePlayer:GetPlayerLevel()
-    
     local minLevel = playerLevel - Questie.db.global.minLevelFilter
     local maxLevel = playerLevel + Questie.db.global.maxLevelFilter
+    local mapRepeatable = Questie.db.global.mapRepeatableQuests
```
BreakBB:
```suggestion
    local mapRepeatable = Questie.db.global.showRepeatableQuests
```

Modules/QuestieQuest.lua
```diff
@@ -1674,7 +1674,9 @@ function _QuestieQuest:IsDoable(questObject)
     if not questObject then
         return false;
     end
-    if questObject.Hidden then
+    if questObject.Hidden and not (questObject.Repeatable and Questie.db.global.mapRepeatableQuests) then
```
BreakBB:
```suggestion
    if questObject.Hidden and not (questObject.Repeatable and Questie.db.global.showRepeatableQuests) then
```

Modules/QuestieQuest.lua
```diff
@@ -1779,22 +1781,24 @@ end
...
     if(not Questie.db.char.manualMinLevelOffset) then
         minLevel = playerLevel - GetQuestGreenRange();
     end
 
     QuestieQuest.availableQuests = {}
 
     for questID, v in pairs(QuestieDB.questData) do
-        --Check if we've already completed the quest and that it is not "manually" hidden and that the quest is not currently in the questlog.
+        local quest = QuestieDB:GetQuest(questID)
+        local questShowHiddenAndComplete = quest.Repeatable and mapRepeatable
```
BreakBB:
```suggestion
        local questShowHiddenAndComplete = quest.Repeatable and showRepeatableQuests
```

BreakBB:
@imclerran I finally had the time to review your files and I added some comments. Please update your PR accordingly so we can merge it 🙌

BreakBB:
@imclerran Any news on this? :)

imclerran:
@BreakBB Thanks for taking the time to review! Been real busy and haven't gotten around to making the fixes yet, but should have time to fix tomorrow.

BreakBB:
@imclerran Thanks for letting us know. But don't stress yourself :)

imclerran:
@BreakBB all conflicts resolved, should be ready for merge!

imclerran:
Ah yeah, I left the indentation as I found it -- those two had single tabs instead of the 4 spaces to indent. I can replace the tabs with spaces though.

BreakBB:
Thanks I wasn't able to see that tabs whwre used in the diff view :thinking:

Issue #1427, opened 2019-10-25
"The Binding" showing, but completed race-specific quest


stevestock:
## Description
"The Binding" is still showing from Gan'rul Bloodeye, even though I already completed it in undercity (I'm an Undead Warlock)

## Wowhead link
I completed undead "the binding" quest: https://classic.wowhead.com/quest=1471
But still showing general horde "the binding" quest: https://classic.wowhead.com/quest=1504

## Screenshots
Can upload if needed

## Questie version
5.0.1 2 e2e80b7

PR #1444, opened 2019-10-27
Issue 1427 - Added dependencies for "The Binding" quest #1440

NguyenMartin:
"The Binding" is still showing from Gan'rul Bloodeye, even if the general questline is finished

Wowhead links
quest: https://classic.wowhead.com/quest=1471
general "the binding" quest: https://classic.wowhead.com/quest=1504

Solved #1427 by adding "exclusiveTo" flags for every Undead quest step to its counterpart.


BreakBB:
It should only be nessesary to add the exclusiveTo to the first quest in the chain. All the other quests should have a pre quest listed and will therefore not been shown if the first one is exclusive to another one.

dimsno:
Thanks! I've gone ahead and removed the redundant exclusivity additions and made only the first quests in the chains to be exclusive.

NguyenMartin:
Thanks! I've gone ahead and removed the redundant exclusivity additions and made only the first quests in the chains to be exclusive.

Issue #1379, opened 2019-10-20
Stonegear's Search


Zardilann:
## Description
Quest can start from two different NPCs. The one in Ironforge doesn't display the quest.

## Wowhead link
https://classic.wowhead.com/quest=467/stonegears-search

## Questie version
4.3

imclerran:
PR fixing this issue [here](https://github.com/AeroScripts/QuestieDev/pull/1455)!

PR #1455, opened 2019-10-28
Issue #1379 Fix -- Added missing "startedBy" NPC for quest 467

imclerran:
[Quest 467](https://classic.wowhead.com/quest=467) can be started by two NPCs -- [1340](https://classic.wowhead.com/npc=1340), and [2092](https://classic.wowhead.com/npc=2092). Only 1340 was present in the DB.

Added via QuestieQuestFixes.


BreakBB:
Thanks for another PR from you! 👍

PR #1486, opened 2019-11-05
update for zhCN config translation

Qcat0214:


BreakBB:
@Qcat0214 Thanks for this! Adding translations is not an easy task for us so we heavily rely on such PRs 🎉

PR #1489, opened 2019-11-05
update config.lua & lookupZones.lua (zhCN)

Qcat0214:


BreakBB:
Thanks for another PR 👍

PR #1528, opened 2019-11-11
Update config.lua

Hantunaar:


BreakBB:
@Hantunaar Thank you very much for another update to the translations 🎉 Keep up the good work 👍

PR #1538, opened 2019-11-14
Fixes tracking of repeatable quests

### LINKED ISSUE #1433: Tracking of repeatable quest (Winterfall Intrusion) not being updated
## Bug description
Tracking of the quest Winterfall Intrusion is not updated after the first completion after the game has loaded (a reload makes it work again for one turn-in).
Stopping and starting tracking of the quest updates the tracjer to the current progress but new progress is not tracked.

## Screenshots
image

## Questie version
5.0.1 2 e2e80b7

scleal:
Fixes #1433

Either of these changes in isolation would fix the issue.

Either repeatable quests should be considered `complete` or we should be laxer about updating them. 

I'm not entirely sure of the structure of the `Questie.db` object or what the `Questie.db.char.complete` map is meant to track, so this seems kind of hack-ish to me. Open to any feedback here.

### DIFF
--- a/Modules/Quest/QuestieQuest.lua
+++ b/Modules/Quest/QuestieQuest.lua
@@ -365,7 +365,7 @@ end
 function QuestieQuest:CompleteQuest(quest)
     local questId = quest.Id
     QuestiePlayer.currentQuestlog[questId] = nil;
-    Questie.db.char.complete[questId] = quest and not quest.Repeatable
+    Questie.db.char.complete[questId] = not quest.Repeatable
     QuestieHash:RemoveQuestHash(questId)
 
     --This should probably be done first, because DrawAllAvailableQuests looks at QuestieMap.questIdFrames[QuestId] to add available
--- a/Modules/Quest/QuestieQuest.lua
+++ b/Modules/Quest/QuestieQuest.lua
@@ -362,9 +362,10 @@ function QuestieQuest:AcceptQuest(questId)
 
 end
 
-function QuestieQuest:CompleteQuest(questId)
+function QuestieQuest:CompleteQuest(quest)
+    local questId = quest.Id
     QuestiePlayer.currentQuestlog[questId] = nil;
-    Questie.db.char.complete[questId] = true --can we use some other relevant info here?
+    Questie.db.char.complete[questId] = quest and not quest.Repeatable
     QuestieHash:RemoveQuestHash(questId)
 
     --This should probably be done first, because DrawAllAvailableQuests looks at QuestieMap.questIdFrames[QuestId] to add available
--- a/Modules/Quest/QuestieQuest.lua
+++ b/Modules/Quest/QuestieQuest.lua
@@ -365,6 +365,7 @@ end
 function QuestieQuest:CompleteQuest(quest)
     local questId = quest.Id
     QuestiePlayer.currentQuestlog[questId] = nil;
+    -- Only quests that aren't repeatable should be marked complete, otherwise objectives for repeatable quests won't track correctly - #1433
     Questie.db.char.complete[questId] = not quest.Repeatable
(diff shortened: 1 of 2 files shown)


Modules/QuestieQuest.lua
```diff
@@ -423,7 +424,7 @@ end
 
 function QuestieQuest:UpdateQuest(QuestId)
     local quest = QuestieDB:GetQuest(QuestId);
-    if quest and not Questie.db.char.complete[QuestId] then
+    if quest and (not Questie.db.char.complete[QuestId] or quest.Repeatable) then
```
BreakBB: I prefer this solution as it saves one call to the database and won't mess with the `complete` state elsewhere.

scleal: Yeah, just not sure if the `complete` state is ever really accurate for repeatable quests. If we opt for this change, after looking at the code again, I think we'd also need to add checks for `quest.Repeatable` in the `CalculateAvailableQuests()` and `AddFinisher()` methods where we check the `complete` state.

BreakBB: I also had another look at the code to check the use of `Questie.db.char.complete` and now I prefer the other solution. Thing is currently we mark repeatable quests as complete even though on the next reset/reload they will be unmarked. The API function `GetQuestsCompleted` (and similar functions) always return `false` for repeatable quests.

So we are actually storing it wrong at this point.

Modules/QuestieQuest.lua
```diff
@@ -363,7 +363,8 @@ end
 
 function QuestieQuest:CompleteQuest(questId)
     QuestiePlayer.currentQuestlog[questId] = nil;
-    Questie.db.char.complete[questId] = true --can we use some other relevant info here?
+    local quest = QuestieDB:GetQuest(questID)
```
BreakBB: Instead of calling `QuestieDB:GetQuest(questID)` we could hand the `quest` from one layer above and store the Id in a local variable. So we save the call here.

scleal: It should be cached based on the implementation of `QuestieDB:GetQuest(questID)`, but I get what you're saying. I updated it to the way you described. Let me know what you think.

BreakBB:
This looks good imo. Will let the other devs have a look since I might be missing something. This should be tested though so we don't introduce sideeffects

scleal:
Rebased to fix merge conflict and squash commits

Modules/Quest/QuestieQuest.lua
```diff
@@ -362,9 +362,10 @@ function QuestieQuest:AcceptQuest(questId)
 
 end
 
-function QuestieQuest:CompleteQuest(questId)
+function QuestieQuest:CompleteQuest(quest)
+    local questId = quest.Id
     QuestiePlayer.currentQuestlog[questId] = nil;
-    Questie.db.char.complete[questId] = true --can we use some other relevant info here?
+    Questie.db.char.complete[questId] = quest and not quest.Repeatable
```
BreakBB: We don't need to check for `quest ~= nil`, since it is already used before. Instead we just need to check for the `Repeatable` property:

```suggestion
    Questie.db.char.complete[questId] = not quest.Repeatable
```

BreakBB: Maybe you could add a small comment to that line as well to explain it for later viewers :ok_hand: 

Also this needs to be tested with regular repeatable quests and with repeatable quests in a child/parent releation (meaning sub quests).

BreakBB:
@scleal Thank you very much for your patience! This PR looks good 👍

scleal:
> @scleal Thank you very much for your patience! This PR looks good 👍

No problem, my pleasure to help 😄

PR #1539, opened 2019-11-14
[#1540] Solve the problem that other addons could not find Questie minimap icon

wolfcon:
eg: 
- Minimap Button Frame
- SexyMap
- ...

#1540


BreakBB:
Thanks for this. It totally makes sense that Questie isn't listed correctly in other addons!

Issue #1594, opened 2019-11-25
Neeka Bloodscar breadcrumb quest


Seprum:
The Neeka Bloodscar (ID: 1418) is not available after completion one of two quests: Coyote Thieves (ID: 1419), Report to Helgrum (ID: 1420)

Seprum:
Just opened #1595 that fixes this breadcrumb quest.

BreakBB:
Fixed with #1595

PR #1595, opened 2019-11-25
[quest] Neeka Bloodscar

Seprum:
Fixes breadcrumb quest (#1594)


BreakBB:
Thanks for this @Seprum I checked the Wowhead links on this and you're right. The fix looks good 👍

PR #1600, opened 2019-11-29
Update config.lua

iron217:


BreakBB:
@iron217 Thank you very much for always taking your time to update the Russian translations ❤️

PR #1616, opened 2019-12-09
Update QuestieQuestFixes.lua

Gogo1951:
Adding a some quests that start from items looted or items picked up.

Screenshots.

I don't think they look too bad.

WoWScrnShot_120919_132943

WoWScrnShot_120919_164035

WoWScrnShot_120919_222307

I don't think this quest is actually repeatable... only weird thing I saw.

WoWScrnShot_120919_133302

### DIFF
--- a/Database/Corrections/QuestieQuestFixes.lua
+++ b/Database/Corrections/QuestieQuestFixes.lua
@@ -14,8 +14,17 @@ function QuestieQuestFixes:Load()
     table.insert(QuestieDB.questData, 7670, {}) -- Add missing quest index #1432
 
     return {
+        [2] = {
+            [QuestieDB.questKeys.startedBy] = {{12676},nil,{16305}},
+        },
         [5] = {
             [QuestieDB.questKeys.preQuestSingle] = {}, -- #1198
+        },
+        [23] = {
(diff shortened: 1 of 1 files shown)


Gogo1951:
I take it back, The Lost Supplies is repeatable. News to me and I've done it a handful of times on different toons. (=

Yeah, looks good!

Gogo1951:
Hey Guys, anything needed from me to move this into master?

BreakBB:
Hey @Gogo1951 I somehow forgot about this. Actually we devs shortly talked about this yesterday and think this would be more like a band aid fix instead of a really good solution. We keep you updated.

Gogo1951:
Hey @BreakBB ,

I'd love a better solution for some of the zone quests, but to be clear everything in this batch is something that starts from a limited number of locations (under 5), or a unique mob that drops an item. Similar to the quests in Booty Bay with the giant in the harbor. No reason not to get these in while you're sorting out a longer-term solution.

I've got another list of about 25 quests when we have a better solution for zone-wide drops / multiple mob dropping items (like the chicken escort beacon quests). Cheers!

Gogo1951:
Hey @BreakBB ,

What's the verdict?

PR #1629, opened 2019-12-12
Update QuestieItemFixes.lua

Teppic1:
Quest: https://classic.wowhead.com/quest=418/thelsamar-blood-sausages

Remove tons of incorrect drops for bear meat.


BreakBB:
@Teppic1 Thank you very much for this! It is not easy for us to test everything, so we appreciate every help we get 👍

PR #1657, opened 2019-12-23
Add some translation text for Korean.

neatnet:


BreakBB:
@neatnet Thank you very much for these additions!

PR #1670, opened 2019-12-30
Thick Yeti Fur does not drop from Ice Thistle Matriarch/Patriarch

gouletr:
As the title says, `Thick Yeti Fur` does **not** drop from `Ice Thistle Matriarch` or `Ice Thistle Patriarch`, so remove these two from the list.


BreakBB:
@gouletr Good catch, thanks for this!

PR #1685, opened 2020-01-04
Fix bug with Questie and Coordinates addons

Ragnoroct:
Both Questie and Coordinates addon's try to write map coordinates.

This change just lets the Coordinates addon do it's thing if it's enabled.


BreakBB:
@Ragnoroct Thanks for this small adjustment. We're always trying to keep up the compatibility with other addons!

PR #1697, opened 2020-01-10
Update lookupNpcs.lua

woopydalan:


BreakBB:
@woopydalan Thank you so much for taking the time here to update this data! 🎉

PR #1711, opened 2020-01-15
Update QuestieQuestFixes.lua

Gogo1951:
Added https://classic.wowhead.com/quest=1148/parts-of-the-swarm


Gogo1951:
This one technically drops off one of 4 mobs in the same area. I picked the highest drop-rate mob and listed it. I don't have a horde toon that I can test this on at the moment, but I don't think it'll be terrible.

Gogo1951:
This one sucks. Please don't merge it into master. I just had a friend test it. Eww. We need to add this one to the zone-drop feature, it doesn't work this way.

PR #1746, opened 2020-01-31
Fix parent quest for quest 254

### LINKED ISSUE #1725: Digging Through the Dirt(254) listed as a available in quests by zone.
This is a quest which only involves interacting with a grave to summon an elite for Bride of the Embalmer(253). Having completed that quest already, I cannot interact with the grave and therefore cannot begin/finish this quest, but it is still being listed as available.

v5.5.1

Doxxer:
Also, it fixes #1725


BreakBB:
Good catch, thanks!

Issue #1807, opened 2020-03-06
Encoded Fragments - wrong npc


bhaap:
## Description
Points to wrong mobs. should point to https://classic.wowhead.com/npc=8766/forest-ooze

## Wowhead link
https://classic.wowhead.com/quest=8235

## Screenshots
<img width="341" alt="Skärmavbild 2020-03-06 kl  20 27 09" src="https://user-images.githubusercontent.com/2168003/76115730-01c61980-5fe9-11ea-92ba-87adca995700.png">

## Questie version
5.7.2

PR #1813, opened 2020-03-10
Fix for #1807

Gogo1951:
Fix for https://github.com/AeroScripts/QuestieDev/issues/1807

Also sorted by Item ID, I know how you guys like to keep things clean. (=

However... I see this "-- AUTO GENERATED FILE! DO NOT EDIT!" and I'm not sure if I should be editing this now. Let me know. This was the only reference I could find that had the incorrect quest mobs in it. Thanks!


BreakBB:
@Gogo1951 Fixes for the `itemDB.lua` go into the `Corrections\QuestieItemFixes.lua`. In there you'll find the same pattern as you'll find in the `itemDB.lua`, but also the possibility to change just single values of each database entry.

Gogo1951:
fixed here https://github.com/AeroScripts/QuestieDev/pull/1814

Closing this one.

PR #1830, opened 2020-03-18
Update lookupZones.lua

woopydalan:


BreakBB:
@woopydalan As Drejjmit already mentioned it would be great if you could change the translation for

```lua
[722] = "Zahúrda Rajacieno"
```

as well to

```lua
[722] = "Zahúrda Rojocieno"
```

woopydalan:
@BreakBB I just wanted to point out that the Rojocieno is actually a typo by Blizzard. If you read any quest text in esES related to the zone, it is spelled as Rajacieno. Not sure how you guys want to proceed with that information.

BreakBB:
@woopydalan Well if it is wrong on Blizzard's side I am totally fine with keeping it correct on our side 👍  Thanks for the hint!

PR #1833, opened 2020-03-20
Update lookupZones.lua

woopydalan:
@BreakBB The latest patch removed the (english name) from all the zones in the frFR locale. Ashenvale was changed to it's retail name Orneval, and Feralas was changed to Féralas


BreakBB:
Oh wow, that is a big (and very late) change :O Thanks for pointing that out. We should change it then as well. Thanks for all the contributions, we really appreciate it 🙏

Issue #1886, opened 2020-04-22
Mazrzanache


Gogo1951:
## Description

Quest item doesn't drop off all planestriders, only Adults and Elders.

https://classic.wowhead.com/item=4806/plainstrider-scale

## Wowhead link

https://classic.wowhead.com/quest=766/mazzranache

PR #1887, opened 2020-04-22
Fix for "Mazrzanache"

Gogo1951:
Fixes for #1886.


Gogo1951:
Please add comment in patch notes "Thanks Blinded from Countdown."

PR #1910, opened 2020-04-29
QuestieTrackerV2

Dyaxler:
* Objective Sorting by Zone: This allows sorting the Tracker first by Zone, then by Quest Level. (Default: By %% Completed)
* Tracker Backdrop and Border: Comes with an on/off Toggle and an Alpha Slider (Default: Off)
* Tracker SetPoint: 4 settings and a detailed tooltip (Default: TOPRIGHT)
* Tracker Minimize Buttons: Changed minimize buttons for the Active Quests header and zone headers into Clickable Buttons that do the same thing.
* Tracker SetClampedToScreen: Keeps it from being accidently dragged off screen forcing a player to reset the tracker position.
* Padding Between Quests: Now uniformly adds space between line elements to make it easier to interact with specific features.
* Numerous Bug Fixes to the Tracker and added a few missing DB entries.


Dyaxler:
image

image

Dyaxler:
With the Minimize Buttons visible and the Dire Maul zone minimized.

image

Dyaxler:
What it looks like with all Quest Objectives minimized... you'll notice that there aren't any minimize buttons for Quests with a Quest Item attached. There was no clean way to place the minimize button so I left them out. But if you minimize the zone they are in the Quest Item buttons will disappear and then reappear when you maximize the zone.

image

Dyaxler:
And of course to make all the purists happy and those that hate changes... :)

The DEFAULT state.

image

Dyaxler:
Sorry... quest objectives are white by default. :D

This is the REAL DEFAULT state... out of box.

image

Issue #1954, opened 2020-05-19
[15] Mountaineer Stormpike's Task


drejjmit:
## Description
Picking up [14] Stormpike's Order (1338) disables [15] Mountaineer Stormpike's Task (1339).
Breadcrumb data seems fine. 
When i pick up [14] Stormpike's Order (1338) then [15] Mountaineer Stormpike's Task (1339) is still showing as available but can't be picked up from Mountaineer Kadrell.

## Wowhead link
https://classic.wowhead.com/quest=1339/mountaineer-stormpikes-task
https://classic.wowhead.com/quest=1338/stormpikes-order

## Questie version
v5.10.3

Gogo1951:
Spoke with @BreakBB and the data for this one looks OK as is... something else is going on. Requires investigation.

drejjmit:
Fixed with https://github.com/AeroScripts/QuestieDev/commit/ee93b7dca09de68c8549056caf607aaebb2f1953

PR #1960, opened 2020-05-19
Fixes for #1954

Gogo1951:
I think the exclusivity was switched around in the original fix.


Gogo1951:
Closing after talk with @BreakBB - I did this wrong.

PR #2096, opened 2020-06-28
Fix for Noboru the Cudgel

drejjmit:
Confirmed by ~20 people on wowhead 
https://classic.wowhead.com/npc=5477/noboru-the-cudgel#comments:id=3159709

(won't tag it [quest] since we already have Noboru the Cudgel in change log)


drejjmit:
Dupe #2102

PR #2119, opened 2020-07-02
Get version string without parsing title

SimonSK:
I think it's better to grab the Version field defined in toc, rather than parsing the title.

### DIFF
--- a/Database/compiler.lua
+++ b/Database/compiler.lua
@@ -980,13 +980,6 @@ function QuestieDBCompiler:GetDBHandle(data, pointers, skipmap, keyToRootIndex, 
     return handle
 end
 
-function QuestieDBCompiler:GetVersionString() -- todo: better place
-    local _,ver = GetAddOnInfo("Questie")
-    if not ver then
-        _,ver = GetAddOnInfo("QuestieDev-master")
-    end
-    -- todo: better regex for this
-    ver = string.sub(ver, 32)
-    ver = string.sub(ver, 0, string.find(ver, "|")-1)
-    return ver
+function QuestieDBCompiler:GetVersionString()
+    return GetAddOnMetadata("Questie", "Version")
 end
--- a/Database/compiler.lua
+++ b/Database/compiler.lua
@@ -981,7 +981,3 @@ function QuestieDBCompiler:GetDBHandle(data, pointers, skipmap, keyToRootIndex, 
 
     return handle
 end
-
-function QuestieDBCompiler:GetVersionString()
-    return GetAddOnMetadata("Questie", "Version")
-end
--- a/Database/compiler.lua
+++ b/Database/compiler.lua
@@ -8,6 +8,8 @@ local QuestieStream = QuestieLoader:ImportModule("QuestieStreamLib"):GetStream("
 local QuestieDB = QuestieLoader:ImportModule("QuestieDB")
 ---@type QuestieSerializer
 local serial = QuestieLoader:ImportModule("QuestieSerializer")
+---@type QuestieLib
+local QuestieLib = QuestieLoader:ImportModule("QuestieLib")
 
 serial.enableObjectLimit = false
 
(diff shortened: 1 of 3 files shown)


SimonSK:
I assume the version info function exists in two different places because of the overall structure.
It would be great to have only one to avoid inconsistencies.
For example, prior to this PR, the two functions were using different regex patterns to extract version numbers from title.

Modules/Libs/QuestieLib.lua
```diff
@@ -338,14 +338,19 @@ end
 local cachedTitle = nil
 -- Move to Questie.lua after QuestieOptions move.
 function QuestieLib:GetAddonVersionInfo()
+    local name = "Questie"
+    local version = GetAddOnMetadata(name, "Version")
+    -- %d = digit, %p = punctuation character, %x = hexadecimal digits.
+    local major, minor, patch = string.match(version, "(%d+)%p(%d+)%p(%d+)")
+
     if (not cachedTitle) then
+        --[[ not really sure why QuestiDev-master is checked
```
BreakBB, line 347: This is because the TOC and the repo folder is named this way. Not ideal and might be changed at some point. Currently we need this.

Database/compiler.lua
```diff
@@ -980,13 +980,6 @@ function QuestieDBCompiler:GetDBHandle(data, pointers, skipmap, keyToRootIndex,
...
-function QuestieDBCompiler:GetVersionString() -- todo: better place
-    local _,ver = GetAddOnInfo("Questie")
-    if not ver then
-        _,ver = GetAddOnInfo("QuestieDev-master")
-    end
-    -- todo: better regex for this
-    ver = string.sub(ver, 32)
-    ver = string.sub(ver, 0, string.find(ver, "|")-1)
-    return ver
+function QuestieDBCompiler:GetVersionString()
```
BreakBB, line 983: This can completely be replaced with `QuestieLib:GetAddonVersionString()`. QuestieLib needs to be loaded on top of the file though.

BreakBB:
@SimonSK Thanks for this PR. Good catch of the VersionString 👍

Database/compiler.lua
SimonSK: `QuestieDBCompiler:GetVersionString()` returns `major.minor.patch` (originally parsed from the title, now just grabbed from the Version defined in toc).
`QuestieLib:GetAddonVersionString()` however returns `"v" .. tostring(major) .. "." .. tostring(minor) .. "." .. tostring(patch) .. buildType`.

Though `QuestieDBCompiler:GetVersionString()` is only used in
https://github.com/AeroScripts/QuestieDev/blob/ed756defc2a8eaaa81fa75274945bc9d69497fcb/Database/compiler.lua#L634
and
https://github.com/AeroScripts/QuestieDev/blob/ed756defc2a8eaaa81fa75274945bc9d69497fcb/Modules/QuestieEventHandler.lua#L147

Does the version string format not matter as long as it is consistent across those two lines?

BreakBB: Yeah as long as these lines are consistent it doesn't matter 👍

Modules/Libs/QuestieLib.lua
```diff
@@ -338,14 +338,16 @@ end
 local cachedTitle = nil
 -- Move to Questie.lua after QuestieOptions move.
 function QuestieLib:GetAddonVersionInfo()
+    local name = "Questie"
+    local version = GetAddOnMetadata(name, "Version")
```
BreakBB, line 342: Also this should be something like

```suggestion
    local version = GetAddOnMetadata("Questie", "Version") or GetAddOnMetadata("QuestieDev-master", "Version")
```

to also cover the dev version.

SimonSK: Fixed. While at it, I added `cachedVersion`.

PR #2129, opened 2020-07-04
Adding Pat for Boss Tho'grun

Gogo1951:
* Boss Tho'grun - NPC - World of Warcraft || https://classic.wowhead.com/npc=2944/boss-thogrun


drejjmit:
This is way too many spawns. Will look awful.
image

Just use the nice spawns that's already present? Just change ``npcKeys.waypoints`` to ``npcKeys.spawns``
image

PR #2137, opened 2020-07-05
[db] Fixed spawn of "Aean Swiftriver"

drejjmit:
Will move her out of "Camp Taurajo" and on to the road where she paths. Also: 
https://classic.wowhead.com/npc=5797/aean-swiftriver#comments:id=2825660


BreakBB:
@drejjmit Good catch, damn 😅

PR #2145, opened 2020-07-06
[quest] Updated objective locations for "Dearest Natalia"

drejjmit:
Setting objective to the actual NPC (for mouseover) instead of an invisible trigger.


BreakBB:
Ah I see. Yeah seems like a good idea 👍

PR #2148, opened 2020-07-06
Update config.lua

anon1231823:


BreakBB:
@anon1231823 Thanks for helping with the translations 🙏

PR #2181, opened 2020-07-18
[quest] Marked "Neeka Bloodscar" as breadcrumb quest

drejjmit:


BreakBB:
@drejjmit 1420 doesn't seem to have a preQuestSingle field in our DB, so the correction has no effect. You want to mark 1418 as exclusiveTo 1420 I guess.

Issue #2182, opened 2020-07-18
[[50] The Final Message to the Wildhammer (7843)] - Missing objective location


Gogo1951:
## Bug description

Quest location marker not on map.

Will have to add a fake "object" I think to make this work.

## Screenshots

![image](https://user-images.githubusercontent.com/55365231/87856676-2ced3d80-c8e6-11ea-92b1-fa01ac33a581.png)

## Questie version

6.0.2

PR #2183, opened 2020-07-18
[quest] Correctly mark "Message to the Wildhammer" chain as Normal quests

drejjmit:
No PvP involved. Icons doesn't show for people with PvP icons disabled.
#2182


BreakBB:
@drejjmit Well the quest is marked as PvP by Blizzard so we shouldn't unmark it as such, should we? Also this doesn't fix the issue from what I see

drejjmit:
I spoke to Gogo in PM about this and he couldn't see any objectives of these quests (7841,7842 and 7843). Then i saw the PvP tag and he got them showing by enable PvP icons. 
image

These tree and ``The Last Barov (5344, 5342) are the only quests outside Battlegrounds that are tagged PvP:
https://classic.wowhead.com/quests/type:41

I don't see a reason for them to be PvP tagged and include them in PvP filtering since they don't involve any PvP.

BreakBB:
Hmmm... I think we shouldn't filter objectives at all. When a quest is in your quest log then the objectives for those should show. Don't you think that will be the better solution?

drejjmit:
This fix requires no coding ^^ And it will still hide the available icon for these three + barov quests if you fix it like that.

BreakBB:
Alright, alright. For now you win 😛

PR #2184, opened 2020-07-18
[quest] Correctly mark "The Last Barov" as Normal quest

drejjmit:
No PvP needed.


BreakBB:
Same as with #2183 . If it is marked as PvP quest it should show up as such

PR #2186, opened 2020-07-19
[db] Add spawns for "The Crimson Courier"

drejjmit:
Not sure why this was changed from clean pathing from spawnDB to some mish mash.

Before:
image

spawnDB (this commit):
image


BreakBB:
@drejjmit Thanks, must have slipped in some way, sorry about that and thanks for the fix 🙏

PR #2357, opened 2020-09-18
[feature] Show objective icons on targets of UnitFrame addons

### LINKED ISSUE #2269: Custom Target Frame Quest Objectives marker
Since I use the addon Shadowed Unit Frames instead of the default Blizzard Unit Frames, I am unable to see any Mob Objectives marks on my Target Frame. 

So to combat this, can you implement the feature to make a custom moveable Target Frame Quest Objectives marker that I can manually position next to the Target Frame? This will solve any issues regarding other addons that replace the default Blizzard Unit Frames.

BreakBB:
Fixes #2269


BreakBB (commented):
Rechecked these changes. Going for testing on master.

PR #2441, opened 2020-10-31
Remove questie from quest announce text

wallrik:
The quest announcing in party chat is sometimes really long, and it does not really need to say "Questie" every time.

If you don't want to merge this PR, I suggest that you correct the grammar - there is not supposed to be a space before the colon. That's a French grammar thing, not English.

Thanks 😊


BreakBB:
Hey @wallrik Thanks for this PR and sorry for this very late response 🙈 

We haven't really decided yet if we want to keep things as they are because of the "branding" and for marketing. I will leave this PR open, maybe we will remove the stuff later on and you should get the credits for it 👍

Gogo1951:
By design.

PR #2445, opened 2020-11-04
Update config.lua

nanjuekaien1:


BreakBB (approved):
Thanks once again for keeping the translations just fine @nanjuekaien1 ❤️

PR #2546, opened 2021-02-19
Add fast skill key/value reference

fathomcodes:


AeroScripts:
This data is available in the table "QuestieProfessions.professionKeys", you shouldn't directly include the values in case they change in the API. always better to do for example QuestieProfessions.professionKeys.BLACKSMITHING

BreakBB:
I agree with aero, so keeping the values on the `QuestieProfessions.professionKeys` and use it as enum wil be safer. Closing this for now :)

Issue #2551, opened 2021-02-27
Darkmist Recluse do not drop Acidic Venom Sac


Bennus:
Quest: https://classic.wowhead.com/quest=1322/the-black-shield

You can see which spiders drop the sac, but ingame the addon tells you to kill Darkmist Recluse spiders, these do not drop it. Only the Darkfang spiders.

(I checked wowhead after killing about 60 spiders without getting a single sac)

raethkcj:
Has this change not made it into a release yet? I'm using v6.2.6 and it's still showing this dropping off the Darkmist spiders after doing Recompile Database.

BreakBB:
Hey @raethkcj yes, this fix is not released for Classic. We've been focussing on a stable TBC version and v6.2.6 didn't include any data fixes.

PR #2552, opened 2021-02-28
fix: The Black Shield - Acidic Venom Sac drops

kristinclemens:
https://github.com/Questie/Questie/issues/2551


AeroScripts:
NPC 4411 and 4413 don't have any spawns in the database but the rest are fine

Shouldn't cause any problems and we'll likely get the missing spawn data in the future

Thanks for the contribution!

PR #2554, opened 2021-03-02
Fixes for 7636 Stave of the Ancients

Faeryqueenaeval:


AeroScripts:
Looks good, seems originally it only referenced the script-spawned version of these NPCs, which have no spawn points in the database

Thanks for the PR!

Issue #2489, opened 2021-01-15
Quest not showing up in NPC tooltip (Feathermoon Stronghold)


rbpaiett:
## Description
The quest Feathermoon Stronghold is available with the NPC Herald Moonstalker in Darnassus.  I have enabled Tooltips > Show quests in NPC tooltips.  The quest doesn't appear in her tooltip.  However for an example to the contrary, the quest Morrowgrain to Darnassus does appear in the tooltip for Mathrengyl Bearwalker in Darnassus.

## Wowhead link
https://classic.wowhead.com/quest=7494/feathermoon-stronghold
https://classic.wowhead.com/quest=3803/morrowgrain-to-darnassus

## Screenshots
<img width="565" alt="World of Warcraft 1_14_2021 10_15_33 PM" src="https://user-images.githubusercontent.com/1621701/104688863-61d02680-56b6-11eb-9d6c-0600be120cb3.png">

<img width="549" alt="World of Warcraft 1_14_2021 10_08_56 PM" src="https://user-images.githubusercontent.com/1621701/104688926-82987c00-56b6-11eb-8db2-e8e35e56ad33.png">

## Questie version
6.2.4

PR #2555, opened 2021-03-02
Feathermoon Stronghold Quest Start Corrections

kristinclemens:
https://github.com/Questie/Questie/issues/2489


AeroScripts:
Not an issue since the other fixes follow this format, but you don't need to include the trailing nils
We should probably clean that up in the future

Thanks for the PR, looks good!

Issue #2516, opened 2021-01-25
Rabine Saturna (1123)


Gogo1951:
## Description

This quest https://classic.wowhead.com/quest=1123/rabine-saturna shows up before it can be picked up; it needs the "New Frontier" quest to be completed before it's available.

PR #2558, opened 2021-03-02
Rabine Saturna Prereq Correction

kristinclemens:
https://github.com/Questie/Questie/issues/2516


AeroScripts:
It looks like this is also an issue with the alliance variant (quest 6762)
Can you add those too? Thanks for the PR!

Issue #2509, opened 2021-01-25
Improvements to Auto-Dialog


Gogo1951:
## Objective

Need to add an option settings to make it so when you talk to an NPC it automatically advanced the dialogue. Very useful for quests like the NPC in Uldaman, or the guy at the tower in Kagrath, or many others, who make you talk to them before they give you a quest.

## Notes

Leatrix Plus has this feature, and we need an option to turn it on by default. Just have it auto-chat with every NPC when yo open the the interact window.

https://www.curseforge.com/wow/addons/leatrix-plus-classic

> Automate gossip - Automatically select a single gossip option by holding down the alt key when opening a gossip window.  Banker, taxi, trainer and vendor gossip windows are automatically skipped with no alt key requirement.

## Testing Feedback

- [ ] Needs to check for available quests in menu before advancing.
- [ ] Shouldn't use Auto Dialog on City Guards.

Gogo1951:
Hey @kristinclemens ,

Can you make this so it doesn't fire on city guards? Came into an issue last night where it was hard for Rawg to get instructions from the guard since it would auto advance. Pretty sure guard was an option, right?

- [ ] Don't auto dialog on City Guards.

kristinclemens:
The way it's coded, it only continues if the only gossip types found are allowed gossip types, i.e. quest or gossip. If any disallowed type is found, it doesn't continue. Here are the [gossip types](https://wowwiki-archive.fandom.com/wiki/API_GetGossipOptions): `banker, battlemaster, binder, gossip, healer, petition, tabard, taxi, trainer, unlearn, or vendor`

kristinclemens:
Although, `quest` isn't even on that list. Weird.

BreakBB:
@kristinclemens You should be able to add the guard NPC IDs to the `_QuestieAuto.disallowedNPC` table. Not really easy to find all of them, but most likely the savest way to prevent auto from triggering on guards

BreakBB:
Since this is already party done in master we should improve this even further. Currently the auto dialog makes it harder to pickup quests, when the NPC also has a dialog option. Same goes for guards are written above.

kristinclemens:
Assuming I’m not working till 8pm again, I should have time to do this tomorrow evening. I’ll be busy from 8-10 but should hopefully have time before that.

Gogo1951:
Moving this back to planning; out of 7.0 tasks. If we get to it great, but need to focus on what we can do for 7.0.

[...]

PR #2566, opened 2021-03-03
feat: enable automatic NPC dialog

kristinclemens:
resolves https://github.com/Questie/Questie/issues/2509

- Advances NPC dialog automatically only if the only gossip types are either `taxi` or `gossip`. This would be easy to change in the future, such as if we want to allow auto-dialog on `bank` and `trainer`.
- Added a helper method to convert an array to a dict.


Locale/enUS/config.lua
```diff
@@ -62,6 +62,8 @@ QuestieLocale.locale['enUS'] = {
     ['ENABLE_AUTO_ACCEPT_QUESTS_DESC'] = "Enable or disable Questie auto-accepting quests.",
     ['ENABLE_AUTO_COMPLETE'] = "Auto Complete",
     ['ENABLE_AUTO_COMPLETE_DESC'] = "Enable or disable Questie auto-complete quests.",
+    ['ENABLE_AUTO_DIALOG'] = "Enable Automatic NPC Dialog",
+    ['ENABLE_AUTO_DIALOG_DESC'] = "Enable or disable Questie automatically advancing through NPC dialog.",
```
BreakBB, line 66: It would be great to copy paste these lines into all the other localization files. That is the current way to keep the translations in a shape which allows to find missing translations.

Modules/Libs/QuestieLib.lua
```diff
@@ -548,3 +548,18 @@ function QuestieLib:UnpackBinary(val)
     end
     return ret
 end
+
+function QuestieLib:ArrayToDict(arr)
+    local t = {}
+    if arr == nil then
+        Questie:Debug(DEBUG_DEVELOP, "ArrayToDict parameter is nil")
+    elseif #arr % 2 ~= 0 then
+        Questie:Debug(DEBUG_DEVELOP, "ArrayToDict parameter is empty")
```
BreakBB, line 557: Since these paths only lead to log messages: Do we need them? Especially in a lib method, that could clutter the log quite easily 🤔

Modules/Options/GeneralTab/QuestieOptionsGeneral.lua
```diff
@@ -285,32 +297,33 @@ function QuestieOptions.tabs.general:Initialize()
...
                 type = "select",
-                order = 7,
+                order = 8,
                 values = _GetShortcuts(),
                 style = 'dropdown',
                 name = function() return QuestieLocale:GetUIString('AUTO_MODIFIER') end,
                 desc = function() return QuestieLocale:GetUIString('AUTO_MODIFIER_DESC'); end,
-                disabled = function() return (not Questie.db.char.autocomplete) and (not Questie.db.char.autoaccept) end,
+                width = 1.5,
+                disabled = function() return not (Questie.db.char.autocomplete or Questie.db.char.autoaccept or Questie.db.char.autoDialog) end,
```
BreakBB, line 320: Good catch on the bool chaining 👍

BreakBB (changes requested):
Hey @kristinclemens thanks for this PR. Great addition to Questie 👌 Please check the commentes below so we can merge this 😁

Modules/Libs/QuestieLib.lua
kristinclemens: I added them while debugging the method itself, but you're right, they're not strictly necessary. I could remove them, or we could put them at the SPAM level. I'm open to either. Do you have a preference?

Modules/Options/GeneralTab/QuestieOptionsGeneral.lua
kristinclemens: Thanks. :D

Locale/enUS/config.lua
kristinclemens: Done: https://github.com/Questie/Questie/pull/2566/commits/cdd3d561ef9b7c178a466dce9d0fc4030644ad1a

Modules/Libs/QuestieLib.lua
kristinclemens: I ended up putting them at spam and updated the method a bit:
https://github.com/Questie/Questie/pull/2566/commits/6e68f63e3db271dea5e4931db556b4eb830852ed

AeroScripts:
Thanks for reviewing this @BreakBB, everything seems to check out now as far as I can tell.

Thanks for the PR!

Issue #1093, opened 2019-09-21
Hover details need line wrapping


BreakBB:
## Bug description

When viewing the details of a quest the text is not wrapped correctly and can extend the size of your screen,

## Screenshots

![image](https://user-images.githubusercontent.com/33514570/65376318-c20baf80-dc9e-11e9-87dc-bfcb062be800.png)

## Questie version

4.1.1

BreakBB:
Muehe wrote [this function](https://github.com/Muehe/sqlua/blob/master/QuestList.py#L662) to reformat the objectives. If we use it to transform our objective data this issue should be fixed.

kristinclemens:
So, I figured out how to do this, but I'm not sure there's a way to actually set the desired width since this seems to be built in to the `GameTooltip:AddLine` API. Here are some examples. What do y'all think?

![image](https://user-images.githubusercontent.com/4911629/110219241-af316e80-7e83-11eb-92b9-c5dd725b9f0b.png)

![image](https://user-images.githubusercontent.com/4911629/110219248-b9536d00-7e83-11eb-8fe0-d4e221b137c9.png)

![image](https://user-images.githubusercontent.com/4911629/110219278-e3a52a80-7e83-11eb-9ab5-5f245b9d9240.png)

![image](https://user-images.githubusercontent.com/4911629/110219312-0b948e00-7e84-11eb-9e44-4f75e99e050d.png)

PR #2571, opened 2021-03-06
feat: wrap quest details text in map tooltip

kristinclemens:
resolves https://github.com/Questie/Questie/issues/1093

Adds text wrapping for quest details. See issue for example images.


AeroScripts:
That's neat! I didn't know there was a flag for text wrapping

We wanted to keep the original text formatting to match the quest log, but that doesn't work for all cases. I believe the quest log wraps in the same way so it should be good

Thanks for the PR!

BreakBB:
We talked about this solution a few month ago and tried to find a better way. We didn't find one (or at least found none as easy as this one) but also totally forgot about this one!

Thanks for bringing it back up :grin:

PR #2589, opened 2021-03-18
Prepare corrections to be done for classic and TBC

BreakBB:


Database/QuestieDB.lua
```diff
@@ -57,6 +57,27 @@ local questTagCorrections = {
     [8408] = {41, "PvP"},
 }
 
+-- race bitmask data, for easy access
+local VANILLA = true -- TODO: Check if we are on TBC
```
BreakBB, line 61:
```suggestion
local VANILLA = string.byte(GetBuildInfo(), 1) == 49
```

PR #2600, opened 2021-03-23
Tooltip Removal Fix.

Logonz:
Before we removed the tooltip key wholesale, now doing a more specifc removal not to lose the whole object!

The fix has not been tested fully but the code looks sane enough.

### DIFF
--- a/Modules/Tooltips/Tooltip.lua
+++ b/Modules/Tooltips/Tooltip.lua
@@ -69,8 +69,24 @@ function QuestieTooltips:RemoveQuest(questId)
         return
     end
 
-    for _, key in pairs(QuestieTooltips.lookupKeysByQuestId[questId]) do
-        QuestieTooltips.lookupByKey[key] = nil
+    for _, key in pairs(QuestieTooltips.lookupKeysByQuestId[questId] or {}) do
+        --Count to see if we should remove the main object
+        local totalCount = 0
+        local totalRemoved = 0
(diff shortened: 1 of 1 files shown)


Modules/Tooltips/Tooltip.lua
```diff
@@ -69,8 +69,24 @@ function QuestieTooltips:RemoveQuest(questId)
         return
     end
 
-    for _, key in pairs(QuestieTooltips.lookupKeysByQuestId[questId]) do
-        QuestieTooltips.lookupByKey[key] = nil
+    for _, key in pairs(QuestieTooltips.lookupKeysByQuestId[questId] or {}) do
+        --Count to see if we should remove the main object
+        local totalCount = 0
+        local totalRemoved = 0
+        for _, tooltipData in pairs(QuestieTooltips.lookupByKey[key] or {}) do
```
BreakBB, line 76: I removed the `or {}` in this line and the loop above, because if this erros out it means we have a data leak, doesn't it? It also errored out for me, didn't find the cause for it though.

BreakBB: The debug stack says it is `RemoveQuest` which is called "every now and then" which will then lead to `QuestieTooltips.lookupByKey[key]` being nil. When you accept a quest and then remove it, it always errors so it should be easy to investigate.

Maybe it is not really worth investigation though, but I think there is a data leak somewhere.

Issue #2599, opened 2021-03-23
Warlock succubus quest 


drejjmit:
## Description
From Discord:
**I've just completed the Lv20 succubuss quest as an orc, and even after a /reload, i still have both "[20] Hearts of the Pure (1476)" & "[20] The Binding (1474)" in my journey, they should also be marked as completed (its the UC version)**

## Wowhead link
https://classic.wowhead.com/quest=1513/the-binding

## Questie version
v6.2.5

Gogo1951:
Suggested a fix for one quest but likely others in the chain still have to be fixed.

drejjmit:
From the original reporter:
**it appears that you can do it even if u did the orc version, you do recieve exp reward and the 12slot bag, but not the actual sucubuss pet**

Zedseayou:
Not sure if I should open a new issue for this, but I don't think the above is actually true anymore. Having done the orc version, the Undercity version continues to show on Questie but no quest is offered if I try to talk to Godrick Farsan.

PR #2602, opened 2021-03-25
Fixes for Devourer of Souls

Gogo1951:
https://github.com/Questie/Questie/issues/2599


drejjmit:
Original reporter double checked this and said:
**it appears that you can do it even if u did the orc version, you do recieve exp reward and the 12slot bag, but not the actual sucubuss pet**
#2599

PR #2604, opened 2021-03-27
WIP: Add TBC databases

BreakBB:
This PR will be used to import the new TBC databases. It also moves the Classic databases into the new `Database\Classic` folder to have them separat from the TBC ones.

Things left to do:

- [ ] Import TBC quest database
- [ ] Import TBC NPC database
- [ ] Import TBC item database
- [ ] Import TBC object database
- [ ] Find a way to load corrections only for the specific database (Should we split the corrections like we split the databases and move the "old" ones to the `Classic` folder and load them conditionally?)


AeroScripts:
Closing this as it will be merged with #2607

Issue #2563, opened 2021-03-02
Update Dependencies


Gogo1951:
https://github.com/Questie/Questie/tree/master/Libs

We have a number of dependencies that seem to be out of date. Let's push to latest and see what breaks? Good to be on latest, even if it means putting time into refactoring to make them work.

According to the Here Be Dragons folks they solved the zoom issue on the map with one of their updates.

Gogo1951:
Broad ask... I think just see what we can update. Hopefully updates fix a few issues. (=

PR #2607, opened 2021-03-29
TBC Beta Support

AeroScripts:
Current state of this PR:

- Pull in new Ace libraries
- Pull in new LibUIDropDownMenu
- Implement SetBackdrop mixins where needed
- Create proxy functions where needed, some of the new API calls return data that's formatted differently from the current classic. Once we know they are updating classic codebase to the new API we should refactor the code that uses these.
- Create compatibility locals when needed
- Pull in new TBC database data

All of these changes should work on both current classic and tbc beta.

Things left to do:

- [x] Replace `GetQuestTimers` with something that works (https://wowpedia.fandom.com/wiki/API_C_QuestLog.GetTimeAllowed)
- [x] Check if `AutoQuestWatch_Insert` will exist in TBC as it does not in Shadowlands --> It does exist!
- [ ] Find out why `zoneMap[QuestieDB.sortKeys.SPECIAL]` does seem to not exist
- [ ] Find out if `QuestieProfessions` needs to be updated, since the required functions were Classic specific and removed prior to Shadowlands
- [ ] Find a solution for `Questie\Libs\LibUIDropDownMenu\LibUIDropDownMenu.lua:1586: hooksecurefunc(): UIDropDownMenu_HandleGlobalMouseEvent is not a function` (A fix is already provided in [this issue](https://www.curseforge.com/wow/addons/libuidropdownmenu/issues/11))
- [x] Import TBC quest database
- [x] Import TBC NPC database
- [x] Import TBC item database
- [x] Import TBC object database
- [x] Find a way to load corrections only for the specific database (Should we split the corrections like we split the databases and move the "old" ones to the `Classic` folder and load them conditionally?)
- [x] Update HBD (for #2563 )
- [ ] Update HBD again when it's updated or tempfix (for #2625 )
- [x] Update Ace (for #2563 )
- [x] Fix "Hide Questie" button position on the map (might be a Blizzard problem as the map is not working correctly)
- [x] Update QuestieEvent to hide the new event quests


AeroScripts:
We need to find an alternative to "LQuestie_Create_UIDropDownMenu"

AeroScripts:
GetQuestLogTitle function is not just renamed, it also returns a table with named keys now. We'll need to create a wrapper function to support both versions

AeroScripts:
AutoQuestWatch_Insert doesn't seem to exist anymore. I've only tested this on shadowlands though, so hopefully it's still in TBC.

AeroScripts:
The quest timer API has changed significantly. I'm not sure we can support both versions easily:
https://wowpedia.fandom.com/wiki/API_C_QuestLog.GetTimeAllowed

This is the only functionality to replace "GetQuestTimers"

AeroScripts:
zoneMap[QuestieDB.sortKeys.SPECIAL] seems to not exist for some reason. We'll have to dig deeper and find out why (seems timed-event related)

AeroScripts:
QuestieProfessions may need to be updated, I'm not sure yet though as the functions it uses are classic-specific and were removed from retail earlier than 9.x

AeroScripts:
Just for reference, this also requires #2604 to fully function.

embeds.xml
```diff
@@ -11,10 +11,9 @@
     <Include file="Libs\AceGUI-3.0\AceGUI-3.0.xml"/>
     <Include file="Libs\LibSharedMedia-3.0\lib.xml"/>
     <Include file="Libs\AceGUI-3.0-SharedMediaWidgets\widget.xml"/>
-    <Include file="Libs\AceConfigQuestie-3.0\AceConfig-3.0.xml"/>
+    <Include file="Libs\AceConfig-3.0\AceConfig-3.0.xml"/>
     <Include file="Libs\AceComm-3.0\AceComm-3.0.xml"/>
     <Include file="Libs\AceSerializer-3.0\AceSerializer-3.0.xml"/>
-    <Include file="Libs\LibUIDropDownMenuQuestie\LibUIDropDownMenu.xml"/>
```
BreakBB, line 17: Don't we have to load the Lib on our side to make sure it is available at all (loading just Questie) and in the version we want?

AeroScripts: Not anymore I don't think, I and our custom fork is really broken with 9.x api

Originally we had to rename a lot because it didnt use LibStub (all globals) the new one is properly contained

BreakBB: But if we don't load it at all and no other addon provides the lib then it will not be loaded, will it?

BreakBB: Oh nevermind I just realised it is loaded in the TOC file

PR #2609, opened 2021-03-30
fix: ArrayToDict checks for empty and odd-sized arrays

kristinclemens:
Adds a check to ensure that we're only transforming the array to a dict when the array has an even number of entries.


BreakBB (approved):
Looks fine to me 😁 Can you explain why this is needed?

kristinclemens:
> Looks fine to me 😁 Can you explain why this is needed?

It's bugged, by otherwise not _needed_ currently. It'll just prevent someone in the future from using it and expecting it to work as one would expect, when in reality it breaks on empty arrays _and_ the error message is wrong, hah.

PR #2615, opened 2021-04-02
Fix [10864] A Burden of Souls

kristinclemens:
Add real mobs to creature objective.

https://tbc.wowhead.com/quest=10864/a-burden-of-souls doesn't have any real mobs associated with it, but https://tbc.wowhead.com/npc=16878/shattered-hand-berserker, https://tbc.wowhead.com/npc=19414/shattered-hand-guard, and https://tbc.wowhead.com/npc=19415/shattered-hand-acolyte give credit.


BreakBB:
These type of quests will be solved using a new objectiveType, so we don't have to solve them all by hand.

PR #2618, opened 2021-04-03
Fix [10896] The Infested Protectors

kristinclemens:
* Adds two dummy creature objectives


BreakBB:
These type of quests will be solved using a new objectiveType, so we don't have to solve them all by hand.

PR #2619, opened 2021-04-03
Create dummy creature objectives using corrected quest data

kristinclemens:
Initial work for using creature objective corrections to populate dummy objectives and register dummy objective tooltips.
* `QuestieLib:PrintTable` was created for initial discovery and I decided others might  find it useful as well.
* `QuestieLib:DeepCopy` was created to avoid contaminating table references.

See discussion beginning here:
https://discord.com/channels/263036731165638656/263040777658171392/827884666085965854

See images from solution beginning here: https://discord.com/channels/263036731165638656/263040777658171392/828000955579498496

### DIFF
--- a/Modules/Libs/QuestieLib.lua
+++ b/Modules/Libs/QuestieLib.lua
@@ -575,3 +575,39 @@ function QuestieLib:UnpackBinary(val)
     end
     return ret
 end
+
+function QuestieLib:PrintTable(keyToPrint, value, debugLevel, seen)
+    if seen and seen[value] then return end
+    -- New table; mark it as seen and iterate recursively
+    local s = seen or {}
+    local d = debugLevel or DEBUG_SPAM
+    s[value] = true
+
+    if (type(value) ~= "table") then
+        Questie:Debug(d, keyToPrint .. ": " .. tostring(value))
+    else
+        for k, v in pairs(value) do
+            local key = keyToPrint .. "." .. k
+            QuestieLib:PrintTable(key, v, d, s)
+        end
+    end
+end
+
+-- Source: https://gist.github.com/tylerneylon/81333721109155b2d244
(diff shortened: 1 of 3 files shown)


BreakBB:
Hey @kristinclemens I am not sure if this is the way we want to tackle the problem. Atm the PR changes the logic of PopulateObjective so our database will have more objectives than the API returnes. But when exporting the `KillCredit` information we will be able to leave the objective amount as they are, but what we acutally want (imo) is to alter `objective.spawnList` in a way it will hold the KillCredit data if the list itself is empty. At the beginning of `QuestieQuest:PopulateObjective` there should be check like:

```lua
if (not next(objective.spawnList)) then
    objective.spawnList = _QuestieQuest:GetSpawnListFromKillCredit(objective.Id)
end

...

function _QuestieQuest:GetSpawnListFromKillCredit(npcId)
    -- query database for KillCredit entry of npcId
    -- transform the query result into the objective.spawnList format
    -- return the transformed table
end
```

Thing is I am currently not sure how `objective.spawnlist` and also `objective` really looks like for one of the "generic NPCs". You did look into the logic and the data though, I didn't ^^

kristinclemens:
@BreakBB  I will definitely take a look at that option! This is a naive implementation that I wasn’t even positive would work, so I’m happy to try a different method. :) The main thing I want to ensure we accommodate is tooltips _and_ icons populating. Once we have the data in our DB associating killable NPC IDs with generic NPC IDs, this all becomes more straightforward IMO.

BreakBB:
Sounds good! Let me know if you need more help :) This fix/solution should be priortized imo as it will fix many quest issues. You are on a good way though 👍

BreakBB:
This has been implemented with a9d3c37d93271320bcdc65c39e4d25d76805506f and is now called `extraObjectives`.

PR #2628, opened 2021-04-14
Fixes for 1188 & 1187

Gogo1951:
Quests

https://tbc.wowhead.com/quest=1188/safety-first
https://tbc.wowhead.com/quest=1187/razzerics-tweaking

Object Container
https://tbc.wowhead.com/object=20727/gizmorium-shipping-crate

Objective Item
https://tbc.wowhead.com/item=5862/seaforium-booster


BreakBB:
Hey Gogo, can you explain what the issue with these quests is and what you are trying to fix?

The object you are fixing with this already has the spawn data just right as you can see [in objectDB.lua - line 1426](https://github.com/Questie/Questie/blob/f3e4e532150f595dfbe8be0870d4e40f79aefe3e/Database/objectDB.lua#L1426). So this correction does not really change anything.

PR #2585, opened 2021-03-14
Localization refactoring

BreakBB:
Refactoring the localization module and everything related to it.

Major Changes:

- `QuestieLocale` has been renamed to `l10n` (which stands for "localization")
- The `l10n`-Module is callable meaning to retrieve translations you do `l10n("someString")`
- All the existing translations have a new format which eases the search for missing translations (Missing translations can be found by globally searching for `["<locale>"] = false` (e.g.  `["deDE"] = false`))
- Instead of some less descriptive key (`JOURNEY_TITLE`) the key is now the English text (`Journey`)

Checkout the #2372 issue for more information.

Things left to do:

- [x] Fix #2245
- [x] Look for other hardcoded parts which should be translated
- [x] Fix map button translation call
- [x] Fix minimap icon profession names
- [x] Make some debug messages hardcoded instead of translatable
- [x] Make tracker "complete" and "failed" message translateable (`line.label:SetText("|cFF40C040" .. _G["QUEST_COMPLETE"] .. "!|r")`)
- [x] Reduce translated strings for objective completion and quest item pickup

PR #2632, opened 2021-04-18
Create ci.yml

QartemisT:


BreakBB:
Hey @QartemisT looks like you read our minds! I was planning to add CI to Questie this week 😄 Thanks for taking the time! I will merge this as soon as #2585 is merged 👍🏻

.github/workflows/ci.yml
```diff
@@ -0,0 +1,46 @@
...
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v1
+        with:
+          fetch-depth: 100
+
+      - name: Install and run Luacheck
+        uses: nebularg/actions-luacheck@v1
```
BreakBB, line 23: There is also a `luacheck@v2` version available already which we should use.

Moreover will this also run luacheck and not just install it? 🤔 For my other addon I added:

```yml
      - uses: leafo/gh-actions-lua@v8.0.0
        with:
          luaVersion: "5.1"

      - uses: leafo/gh-actions-luarocks@v4.0.0

      - name: build
        run: |
          luarocks install luacheck
          luacheck .
```

to install lua + luarocks and run the luacheck command.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,46 @@
...
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v1
```
BreakBB, line 18: Oh and there is also a v2 of this action.

.github/workflows/ci.yml
QartemisT: Yeh it does both installing and running.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,46 @@
...
+
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 100
+
+      - name: Install and run Luacheck
+        uses: nebularg/actions-luacheck@v2
+        with:
+          args: "--no-color -q"
```
BreakBB, line 25:
```suggestion
          args: "Database Localization --no-color -q"
```

BreakBB (approved):

PR #2643, opened 2021-04-23
[locale][ptBR] Update more translations

lucasfevi:
A few more translations added and also fixed a few others.


BreakBB:
Thank you very much @lucasfevi for improveing the Brazilian translations! 🙏🏻

PR #2660, opened 2021-04-28
Update tbcItemFixes.lua

GillsITWorld:
Fix for Quest 11885

### DIFF
--- a/Database/Corrections/TBC/tbcItemFixes.lua
+++ b/Database/Corrections/TBC/tbcItemFixes.lua
@@ -103,6 +103,9 @@ function QuestieTBCItemFixes:Load()
         [30783] = {
             [itemKeys.objectDrops] = {184867},
         },
+        [32742] = {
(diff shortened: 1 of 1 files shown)


Database/Corrections/TBC/tbcItemFixes.lua
```diff
@@ -103,6 +103,9 @@ function QuestieTBCItemFixes:Load()
         [30783] = {
             [itemKeys.objectDrops] = {184867},
         },
+        [32742] = {
+			[itemKeys.npcDrops] = {23363}
+		},
```
BreakBB, line 108: Good catch on the `vendors` not being the `npcDrops` 👍🏻 

Could you fix the format of these two lines though?

BreakBB (changes requested):

BreakBB:
Also I think this fixes [A Shabby Disguise](https://tbc.wowhead.com/quest=11029/a-shabby-disguise) not 11885, right?

PR #2662, opened 2021-04-28
Update tbcQuestFixes.lua

GillsITWorld:
Add Extra Objectives for Quest 11885 (Fix 2 of 2)
Tested on Beta

PR #2663, opened 2021-04-28
Update tbcNPCFixes.lua

GillsITWorld:
Add Kill Objective Spawn/Summon Locations for Quest 11885 (Fix 1 of 2)
Tested in Beta


BreakBB:
Hey @GillsITWorld actually these NPCs don't have spawn points, but are summoned by using the "Time-Lost Scrolls". So we don't want to add spawns for them, as it implies that the NPCs will spawn "if you just wait long enough".

To fix the quest the `extraObjective` you used in #2662 is correct, but not entirely (check the comment on that PR).

GillsITWorld:
Yes, but since you have to use the scrolls at certain points, shouldn't we mark those summon points? (This was the "patch" method Aero suggested)

GillsITWorld:
Never mind read the other comment

PR #2664, opened 2021-04-29
Update tbcQuestFixes.lua

GillsITWorld:
Replace preQuestSingle with 10299 (Wrong Pre Quest currently entered)


BreakBB (approved):
Looks good 👍🏻

PR #2693, opened 2021-04-30
Fix for Quest 10263 Added PreQuest

GillsITWorld:


Database/Corrections/TBC/tbcQuestFixes.lua
```diff
@@ -739,6 +739,9 @@ function QuestieTBCQuestFixes:Load()
         [10246] = {
             [questKeys.preQuestSingle] = {10299},
         },
+        [10263] = {
+            [questkeys.preQuestSingle] = {10551},
```
BreakBB, line 743: I think the better fix is to add reputation requirements to The Aldor and change the prequest to be either 10551 or 10552.

Technically you could side with the Sycther, immediately change your mind and work your way towards the Aldor. Then you could still complete 10263 even though you didn't do 10551 but 10552.

BreakBB (changes requested):

Database/Corrections/TBC/tbcQuestFixes.lua
BreakBB, line 743:
```suggestion
            [questKeys.preQuestSingle] = {10551},
```

Database/Corrections/TBC/tbcQuestFixes.lua
```diff
@@ -739,6 +739,9 @@ function QuestieTBCQuestFixes:Load()
         [10246] = {
             [questKeys.preQuestSingle] = {10299},
         },
+        [10263] = {
+            [questKeys.preQuestSingle] = {10551},
```
BreakBB, line 743:
```suggestion
            [questKeys.preQuestSingle] = {10551,10552},
            [questKeys.requiredMinRep] = {932,3000},
```

BreakBB (approved):

PR #2694, opened 2021-04-30
Fix for Quest 10264 added PreQuest

GillsITWorld:


Database/Corrections/TBC/tbcQuestFixes.lua
```diff
@@ -729,6 +729,9 @@ function QuestieTBCQuestFixes:Load()
         [10231] = {
             [questKeys.triggerEnd] = {"Beat Down \"Dirty\" Larry and Get Information", {[zoneIDs.SHATTRATH_CITY]={{43.86,27.97},},}},
         },
+        [10264] = {
+            [questkeys.preQuestSingle] = {10552},
```
BreakBB, line 733: I think the better fix is to add reputation requirements to The Scryers and change the prequest to be either 10551 or 10552.

Technically you could side with the Aldor, immediately change your mind and work your way towards the Scryers. Then you could still complete 10264 even though you didn't do 10552 but 10551.

BreakBB (changes requested):

Database/Corrections/TBC/tbcQuestFixes.lua
BreakBB, line 733: Noted this too late for the merged PRs, but the variable is called `questKeys` not `questkeys`. You should grab a better IDE to show you these mistakes 😉 

```suggestion
            [questKeys.preQuestSingle] = {10552},
```

Database/Corrections/TBC/tbcQuestFixes.lua
```diff
@@ -842,6 +842,9 @@ function QuestieTBCQuestFixes:Load()
         [10246] = {
             [questKeys.preQuestSingle] = {10299},
         },
+        [10264] = {
+            [questKeys.preQuestSingle] = {10552},
```
BreakBB, line 846:
```suggestion
            [questKeys.preQuestSingle] = {10551,10552},
            [questKeys.requiredMinRep] = {934,3000},
```

BreakBB (approved):

PR #2695, opened 2021-04-30
Fix for Quest 10265 Added PreQuests

GillsITWorld:


Database/Corrections/TBC/tbcQuestFixes.lua
```diff
@@ -729,6 +729,9 @@ function QuestieTBCQuestFixes:Load()
         [10231] = {
             [questKeys.triggerEnd] = {"Beat Down \"Dirty\" Larry and Get Information", {[zoneIDs.SHATTRATH_CITY]={{43.86,27.97},},}},
         },
+        [10265] = {
+            [questKeys.preQuestSingle] = {10263,10264,10551,10552},
```
BreakBB, line 733: The "Assisting the Consortium" quests are breadcrumb quests I assume. Then it would be better to just add the Aldor/Scryer quests as pre quests and mark 10263 and 10264 as exclusive to 10265

BreakBB (changes requested):

Database/Corrections/TBC/tbcQuestFixes.lua
GillsITWorld: Ok, so then if we do this we would also have to mark 10551 and 10552 also exclusive to 10265, correct?

BreakBB: 10551 and 10552 are already marked as exclusive to each other. So you need to complete one of them to get 10265 and therefore no further exclusive marking is required :)

drejjmit:
I can pick this up with no quests done and 0 rep with The Consortium

PR #2698, opened 2021-05-01
Korean Updated Map

HoYeonJO:


BreakBB (approved):
Looks good as well 👍🏻 

Also: If you are changing multiple files it is totally fine to make a single PR for it. You can even add more and more changes to a single PR if you keep pushing commits on the existing branch.

If you are not done with the changes just add a `WIP: ` prefix to the PR title and I will skip the PR as long as the WIP is in the title 👍🏻

PR #2764, opened 2021-05-19
Database Update

Muehe:
* Lots of zone fixes and removed duplicates
* triggerEnd field in quest DB has data again
* Remove tagged NPCs
* Remove waypoints of NPCs with more than 2 spawns


BreakBB (changes requested):
Great PR!

There are even more triggerEnd included than we got fixed in the tbcQuestFixes.

It might be worth to remove the triggerEnd from corrections with this PR as well. But don't do it for every single one: 10710, 10711 and 10712 got edited on purpose.

PR #2784, opened 2021-05-20
Database continued

Logonz:
More info in the commit messages


AeroScripts:
The database tables should still have "TBC" added to them, any reason for removing that?

There are a few places that need them to be separate to work (And the current logic relies on that)

AeroScripts:
Nice job on the fixes 👍

PR #2789, opened 2021-05-20

Iowerth:
DebugMessages.lua and some corrections


AeroScripts:
Thanks again for the translations!

PR #2800, opened 2021-05-21
[Locale][zhCN] Readded/corrected TBC quest/iitem/npc zhCN locale data

qqytqqyt:


AeroScripts:
Thanks again for the PR, looks good!

PR #2824, opened 2021-05-23

Iowerth:
World event string in DebugMessages.lua


Iowerth:
Sorry, I was in a hurry.

PR #2827, opened 2021-05-23

Iowerth:
World event string in DebugMessages.lua and some corrections.


Iowerth:
Added missed translations for Townsfolk.lua

BreakBB (approved):

PR #2838, opened 2021-05-25
[fix] Unnecessarily short waypoints removed

Logonz:


drejjmit:
+ PH with spawn points added back. OLD stripped from quest name

PR #2849, opened 2021-05-27

Iowerth:
New string in DebugMessages.lua


Muehe:
Cherry-picked as 8339f605f48f47ee98dc76854f6275482799adbc , closing here. Thank you. :+1: 

Just FYI, the reason I cherry picked your commit instead of merging the PR is so we don't get all those merge commits you see in this PR into our history. I think I'm gonna write a wiki page later about the git workflow, so you don't have to do those merges anymore.

Thanks again for all your contributions. :smiley:

PR #2945, opened 2021-06-13
Wrong Korean translate updated Townsfolk.lua

HoYeonJO:


HoYeonJO:
Miss one, I will update with another one

PR #3033, opened 2021-07-17
Update the LDB display text when Quest object text is displayed

Jakobud:
Questie's use of the LibDataBroker display is... very minimal. It's pretty much just an icon that is used to access the addon settings, etc. The LDB Display Text is just the static word "Questie". Not very useful.

I thought I could make it a little more useful. You know whenever you get the yellow text on the screen that shows when a quest objective has been updated? For example:

Muck Core: 2/5
or
Gan'arg Engineer slain: 6/16

I added some code that updates the LDB display text to the latest quest object message. 

For example: 

Here is Questie's current bland use of LDB. I'm using the LDB display addon "Bazooka". It's just a bar at the top of the screen. It shows the Questie LDB icon as well as some others like Attune, Bulletin Board, ItemRack, etc.
WoWScrnShot_071721_164112

Now here it is after my update. You can see that when a quest objective message happens on the screen, the LDB display text also updates.
WoWScrnShot_071721_164238

Here it is with the LDB "Questie" label turned on (which is simply togglable in Bazooka)
WoWScrnShot_071721_164746

It uses the UI_INFO_MESSAGE event and looks for one of several different global strings that are used for Quest objective messages.

This functionality is really useful because it allows you to quickly and easily see how many more "something" you have to kill or gather or whatever. Like if you are mid-way through a quest and have to get up from your desk for a few minutes, when you come back you don't have to think about how many more things you have to kill and you don't have to search through your Quest Log or the Questie Tracker to find out. Simply look at this LDB icon and it tells you what the last thing was that you did.

### DIFF
--- a/Modules/MinimapIcon.lua
+++ b/Modules/MinimapIcon.lua
@@ -26,7 +26,7 @@ function MinimapIcon:Init()
 end
 
 function _MinimapIcon:CreateDataBrokerObject()
-    return LibStub("LibDataBroker-1.1"):NewDataObject("Questie", {
+    local LDBDataObject = LibStub("LibDataBroker-1.1"):NewDataObject("Questie", {
         type = "data source",
         text = "Questie",
         icon = "Interface\\Addons\\Questie\\Icons\\complete.blp",
@@ -80,5 +80,14 @@ function _MinimapIcon:CreateDataBrokerObject()
             tooltip:AddLine(Questie:Colorize(l10n('Ctrl + Left Click'),   'gray') .. ": ".. l10n('Reload Questie'));
         end,
     });
+
+    self.LDBDataObject = LDBDataObject
+
+    return LDBDataObject
+end
+
+--- Update the LibDataBroker text
+function MinimapIcon:UpdateText(text)
+    _MinimapIcon.LDBDataObject.text = text
 end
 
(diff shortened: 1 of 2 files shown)


BreakBB:
Hey @Jakobud Thanks for this very slim and nice PR!

Jakobud:
Your welcome!

Now that I'm playing with it a bit more, one potential improvement I can see would be saving the last message in the profile database and then restoring the text to that value next time you log in.

So instead of logging in and the display says "Questie" it would say "Gan'arg Engineer slain: 6/15" still. I think that would be worth adding.

BreakBB:
Sure go ahead and add that as well. I will mark your PR as "WIP" for now then :) Feel free to unmark it as such, once you are done 👍🏻

Jakobud:
Okay. Is `Questie.db.char` the best place to save a value like that?

BreakBB:
Yes, that is the best place for such information

Jakobud:
All done. Pretty simple!

BreakBB:
This looks great @Jakobud I am happy to merge it 👍🏻

Jakobud:
Awesome thanks

BreakBB:
Also this will be included in the next release which we are planning to drop in the next couple of days (maybe today).

PR #3042, opened 2021-07-21
Add support for release-specific directories (also fix incorrect XP values)

ainslievaux:
This PR adds support to the `build.py` script for release-specific directories; that is, separate directories for TBC or Classic-specific items. Add a top-level directory named `<dirname>-only-<release>` and when `build.py` is invoked with the `release` argument it will copy the aforementioned directory to `<dirname>`, e.g.:

    $ mkdir foo-only-tbc
    $ python3 build.py tbc
    $ ls -d releases/*tbc*/Questie/foo
    releases/v6.4.1-tbc-5b936ce1/Questie/foo
    $ python3 build.py classic
    $ ls -d releases/*classic*/Questie/foo
    ls: releases/*classic*/Questie/foo: No such file or directory

The impetus for this change is the difference between Classic and TBC XP rates, for example:
https://classic.wowhead.com/quest=2904/a-fine-mess
https://tbc.wowhead.com/quest=2904/a-fine-mess
Currently, on Classic-era servers, the TBC XP values are displayed, which are often incorrect.

I've also taken the liberty of adding the appropriate version of `LibQuestXP` to new `Libs-only-classic` and `Libs-only-tbc` directories under the main Questie directory, so when the appropriate release is generated, the correct XP values are placed in `Libs/LibQuestXP`. I'm happy to put those changes in a separate PR if that's preferred.

Tested and working on both TBC & Classic-era servers.


BreakBB:
Hey @ainslievaux

Thanks for this PR, I totally forgot to look into the XP difference between Classic Era and TBCC 👍🏻 

This is really a great fix and also a well written PR 👌🏻

PR #3061, opened 2021-07-25
Fix quest turn in for instant quests which don't fire QUEST_REMOVED

Laumesis:
"Instant" quests (I don't know correct term) won't fire QUEST_REMOVED and so they pollute latestTurnedInQuestIds[]. This breaks next quest turn ins / completions. Visible to users mostly so that quest markers won't disappear from the map / the minimap.

There was also missing a call to add these "instant" quest completions to Questie Journey. I choosed to go via ready helper function for hopefully easy future  code maintenance. Downside for this is use of timer in helper function until server returns the quest to be completed status.

### DIFF
--- a/Modules/QuestieEventHandler.lua
+++ b/Modules/QuestieEventHandler.lua
@@ -306,9 +306,10 @@ function _EventHandler:QuestFinished()
         Questie:Debug(DEBUG_DEVELOP, "shouldRunQLU still active")
         if next(latestTurnedInQuestIds) then
             Questie:Debug(DEBUG_DEVELOP, "finishedEventReceived is questId")
-            local quest = QuestieDB:GetQuest(latestTurnedInQuestIds[1])
+            -- these quests won't fire QUEST_REMOVED event
+            local questId = table.remove(latestTurnedInQuestIds, 1)
             Questie:Debug(DEBUG_DEVELOP, "Completing automatic completion quest")
-            QuestieQuest:CompleteQuest(quest)
+            _EventHandler:CompleteQuest(questId)
         else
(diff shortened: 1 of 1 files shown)


Laumesis:
I committed stuff to a wrong branch (master) in my fork as I was too sleepy. I had to hard reset it back to the previous state and force push to GitHub.

Sorry @AeroScripts 😟

BreakBB:
Hey @Laumesis can you explain what you mean with "Downside for this is use of timer in helper function until server returns the quest to be completed status."?

Your changes doesn't include a timer, but maybe one is already in use? 🤔

Laumesis:
`function _EventHandler:CompleteQuest(questId, count)` has recursion until the client gets info from a server that the quest is really flagged completed. recursion has with max 5s "timeout" after the function does its thing anyway.

image

Alternative fix would be
```
        if next(latestTurnedInQuestIds) then
            Questie:Debug(DEBUG_DEVELOP, "finishedEventReceived is questId")
            local questId = table.remove(latestTurnedInQuestIds, 1)
            local quest = QuestieDB:GetQuest(questId)
            Questie:Debug(DEBUG_DEVELOP, "Completing automatic completion quest")
            QuestieQuest:CompleteQuest(quest)
            QuestieJourney:CompleteQuest(questId)
        else
```

Laumesis:
Came to my mind:
Was it intentional to not save completion of these quests to the journey? For example repeatable mat turn ins for reputation would count as these, I think. On the other hand for example completing each Midsummer Bonefire quest would be nice to get saved into journal.

BreakBB:
@Laumesis The Journey should track all quests handed in, repeatable and non repeatable imo. Not sure how it was initially designed, since the guy who wrote the Journey is long gone 😅

BreakBB (approved):

PR #3090, opened 2021-07-30
Remove 1st of double entries in localization tables

Laumesis:
The first entry would be overwritten by 2nd entry anyway.

This pull request **MUST be reviewed and tested** before merge.


BreakBB:
Good catch. Normally the CI checks for double entries in a table, but we are skipping the generated files as they _should_ have the correct format.

I'll review this in detail the next days if nobody else chims in before me

BreakBB:
Still didn't find the time to test this @Laumesis but everything looks fine from what I see.

BreakBB (approved):

PR #3093, opened 2021-07-30
Unify some formating in DB corrections

Laumesis:
Formating of questKeys.startedBy and .finishedBy values were varying a lot. I choose to unify all those for hopefully better readability.

Example:
```
old: {{1939,},nil,{3317,},},
new: {{1939},nil,{3317}},
```

questkeys.finishedBy should have only 2 fields, so extra nil at end is always waste. I removed those.

vscode find & replace pairs:
```
(\[questKeys\.finishedBy\] *= *\{(\{.*?\}|nil),(\{.*?\}|nil)),nil,?\}
$1}
```
```
(\[questKeys\.finishedBy\] *= *\{.+?),\}
$1}
```
```
(\[questKeys\.startedBy\] *= *\{.+?),\}
$1}
```


BreakBB:
I guess you should be able to remove all trailing `,` in the corrections. So everything that is `,}` ca be replaced by just `}`

Laumesis:
@BreakBB Yes that is doable just fine as noted on discord for automatically generated data.

But is it good for those who update the files? @drejjmit

BreakBB:
@Laumesis It is also fine for the people doing data corrections. I think Drejjmit already omits the trailing `,` and I am responsible for most of them as I thought it would increase readability. But I am not sure about it anymore and would be totally fine without any of the trailing ones.

So feel free and simply remove them 👍🏻

Laumesis:
I replaced all `,}` -> ´}`  in `Database\Corrections\Classic\*` and `Database\Corrections\TBC\*`
I split some lines having multiple zoneIDs locations so that each zone has own line.
I recompiled DB on BCC & Era without errors.

BreakBB (approved):
Looks good, thanks 👍🏻

PR #3107, opened 2021-08-04
Update General.lua

anon1231823:


Localization/Translations/Options/General.lua
```diff
@@ -53,25 +53,25 @@ local generalOptionsLocales = {
     ["Config Tracking Icons"] = {
         ["ptBR"] = "Configurar ícones de rastreamento",
         ["ruRU"] = "Отслеживаемые значки",
-        ["deDE"] = "Tracking Icons konfigurieren","Permet de sélectionner les icônes de suivi (comme la boîte aux lettres, les PNJ de réparation) à afficher sur la carte et la mini-carte.",
```
BreakBB, line 56: O.o no idea how that was added there. Good catch!

BreakBB (approved):

PR #3115, opened 2021-08-10
Add annotations for few classes

Laumesis:
Add annotations for fields defined in other files

Remove one empty line to get automatical annotations from fields defined in code in same file.

### DIFF
--- a/Database/QuestieDB.lua
+++ b/Database/QuestieDB.lua
@@ -24,7 +24,7 @@
 ---@field npcDataTBC nil | table
 ---@field itemDataTBC nil | table
 ---@field sortKeys table
----@field private private table
+---@field private table
 ---@field private _itemAdapterQueryOrder table @temporary, until we remove the old db funcitons
 ---@field private _objectAdapterQueryOrder table @temporary, until we remove the old db funcitons
 ---@field private _questAdapterQueryOrder table @temporary, until we remove the old db funcitons
--- a/Database/QuestieDB.lua
+++ b/Database/QuestieDB.lua
@@ -24,10 +24,11 @@
 ---@field npcDataTBC nil | table
 ---@field itemDataTBC nil | table
 ---@field sortKeys table
----@field _itemAdapterQueryOrder table @temporary, until we remove the old db funcitons
----@field _objectAdapterQueryOrder table @temporary, until we remove the old db funcitons
----@field _questAdapterQueryOrder table @temporary, until we remove the old db funcitons
----@field _npcAdapterQueryOrder table @temporary, until we remove the old db funcitons
+---@field private private table
+---@field private _itemAdapterQueryOrder table @temporary, until we remove the old db funcitons
+---@field private _objectAdapterQueryOrder table @temporary, until we remove the old db funcitons
+---@field private _questAdapterQueryOrder table @temporary, until we remove the old db funcitons
+---@field private _npcAdapterQueryOrder table @temporary, until we remove the old db funcitons
 local QuestieDB = QuestieLoader:CreateModule("QuestieDB")
 local _QuestieDB = QuestieDB.private
 
--- a/Database/QuestieDB.lua
+++ b/Database/QuestieDB.lua
@@ -1,4 +1,33 @@
 ---@class QuestieDB
+---@field npcCompilerTypes table
+---@field npcCompilerOrder table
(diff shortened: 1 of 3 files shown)


Database/QuestieDB.lua
```diff
@@ -1,4 +1,33 @@
...
+---@field itemData any
+---@field questDataTBC nil | table
+---@field objectDataTBC nil | table
+---@field npcDataTBC nil | table
+---@field itemDataTBC nil | table
+---@field sortKeys table
+---@field _itemAdapterQueryOrder table @temporary, until we remove the old db funcitons
+---@field _objectAdapterQueryOrder table @temporary, until we remove the old db funcitons
+---@field _questAdapterQueryOrder table @temporary, until we remove the old db funcitons
+---@field _npcAdapterQueryOrder table @temporary, until we remove the old db funcitons
```
BreakBB, line 30: These should never be really accessed, so we better don't add them to the "official" class definition

BreakBB:
I was also thinking about adding some dedicated `types.lua` file which contains the definition of all important types and don't bloat the already full module files.

Database/QuestieDB.lua
Laumesis: Well, those still generate code diagnostics warnings.

Laumesis:
> I was also thinking about adding some dedicated `types.lua` file which contains the definition of all important types and don't bloat the already full module files.

Is someone really going to keep that file up to date? 🙄

Laumesis:
> I was also thinking about adding some dedicated `types.lua` file which contains the definition of all important types and don't bloat the already full module files.

Also would it make git merges harder? All code file changes could possibly change tha same file

BreakBB:
There could be multiple type files and I agree that it might be outdated quite quickly.

Personally I am just not a fan of 20 lines of `---@field foo bar` inside some components and adding a dedicated file next to it is way cleaner imo.

drejjmit:
funcitons 😅

Laumesis:
> funcitons 😅

I copy/pasted 😁

btw, no idea how to do `@field` for function and tell its parameters and return type. Fun sides of defining things all over place in zillion files :(

Laumesis:
Opinions now?

In my opinion we shouldn't define stuff to a module/class outside its file.
As a wild idea: otherwise maybe ".h" file for each module/class that tells all fields. `QuestieFrame_h.lua`? And tell things so well that no need to hunt 1h around code what possible stuff the table may contain 😅

Database/QuestieDB.lua
Laumesis: I added private to those

Database/QuestieDB.lua
```diff
@@ -1,4 +1,34 @@
...
+---@field npcData any
+---@field questData any
+---@field objectData any
+---@field itemData any
+---@field questDataTBC nil | table
+---@field objectDataTBC nil | table
+---@field npcDataTBC nil | table
+---@field itemDataTBC nil | table
+---@field sortKeys table
+---@field private private table
```
BreakBB, line 27:
```suggestion
---@field private table
```

Database/QuestieDB.lua
```diff
@@ -1,4 +1,34 @@
...
+---@field objectKeys table
+---@field objectKeysReversed table
+---@field itemCompilerTypes table
+---@field itemCompilerOrder table
+---@field itemKeys table
+---@field itemKeysReversed table
+---@field npcData any
+---@field questData any
+---@field objectData any
+---@field itemData any
```
BreakBB, line 21: Those should also be `table` I think

Database/QuestieDB.lua
```diff
@@ -1,4 +1,34 @@
...
+---@field questDataTBC nil | table
+---@field objectDataTBC nil | table
+---@field npcDataTBC nil | table
+---@field itemDataTBC nil | table
+---@field sortKeys table
+---@field private table
```
BreakBB, line 27:
```suggestion
---@field private private table
```

Database/QuestieDB.lua
Laumesis: Done in https://github.com/Questie/Questie/pull/3115/commits/d7d473bcf3757d69408e589e2bbe706d07fb4517

BreakBB (approved):

PR #3116, opened 2021-08-10
Update Townsfolk.lua

anon1231823:


Localization/Translations/MinimapIcon/Townsfolk.lua
```diff
@@ -252,19 +252,19 @@ local townsfolkLocales = {
...
         ["esES"] = "Objetos comerciables",
-        ["frFR"] = "Fournitures",
+        ["frFR"] = "Fournitures d'artisanat",
     },
     ["Vendor"] = {
-        ["ptBR"] = "Vendedor",
+        ["ptBR"] = "Comerciante",
         ["ruRU"] = "Торговец",
-        ["deDE"] = "Händler",
+        ["deDE"] = "Verkäufer",
```
BreakBB, line 260: "Händler" fits better for this translation.

```suggestion
        ["deDE"] = "Händler",
```

anon1231823:
@BreakBB Actually if you look at the words used in the german localization, it's Verkäufer

See
https://de.classic.wowhead.com/npc=3323/horthus
https://de.classic.wowhead.com/npc=1257/keldric-boucher
https://de.classic.wowhead.com/npc=1285/thurman-mullby

Localization/Translations/MinimapIcon/Townsfolk.lua
anon1231823: @BreakBB Actually if you look at the words used in the german localization, it's Verkäufer

See
https://de.classic.wowhead.com/npc=3323/horthus
https://de.classic.wowhead.com/npc=1257/keldric-boucher
https://de.classic.wowhead.com/npc=1285/thurman-mullby
https://de.classic.wowhead.com/npc=4585/ezekiel-graves
https://de.classic.wowhead.com/npc=3369/gotri

BreakBB:
Hey @anon1231823 wow, that is really a thing 😅 As a German persona, playing on an English client I would totally expect it to be translated into "Händler" instead of "Verkäufer 🙈 

Will change it again, thanks for the hint and all the great PRs! 🙏🏻

PR #3126, opened 2021-08-11
[fix] Show quest complete icons and objectives icons if not specially disabled

Laumesis:
Show quest complete icons on maps for following quest type icons even other icons for those types are disabled
- Repeatable
- Event
- Dungeon
- Raid
- PvP
- AQWarEffort

I'm not sure if showing some of those types are OK as npc won't be there anymore necessary. **Opinions?**

Also using more local variables instead of searching from tables


Laumesis:
Atleast Dungon & Raid & PvP quest turn-in icons must be shown.

Laumesis:
Please review to avoid brain farts.

Laumesis:
I did double check: It contains all old checks (AQWarEff is commented out).

Only real change is to move part of check also require `iconType == "available"`. If that subset of checks is okey, feel free to merge.

PR #3127, opened 2021-08-12
refactor iconFrame pools in QuestieFramePool

Laumesis:
Refactored usedFrames, unusedFrames, allFrames
Moved iconFrame (re)init code to outside of GetFrame()
The (re)init code needs to be moved to QuestieFrame.lua

### DIFF
--- a/Modules/FramePool/QuestieFramePool.lua
+++ b/Modules/FramePool/QuestieFramePool.lua
@@ -16,17 +16,21 @@ local l10n = QuestieLoader:ImportModule("l10n")
 
 local HBDPins = LibStub("HereBeDragonsQuestie-Pins-2.0")
 
+local tinsert = tinsert
+local tremove = tremove
+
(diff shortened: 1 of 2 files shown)


Modules/FramePool/QuestieFramePool.lua
```diff
@@ -16,17 +16,21 @@ local l10n = QuestieLoader:ImportModule("l10n")
 
 local HBDPins = LibStub("HereBeDragonsQuestie-Pins-2.0")
 
+local tinsert = tinsert
+local tremove = tremove
```
BreakBB, line 20:
```suggestion
local tinsert = table.insert
local tremove = table.remove
```

Laumesis:
Added [fix] increment lineFrameCount only for new frames

Still needs The (re)init code needs to be moved to QuestieFrame.lua

PR #3133, opened 2021-08-19
for PR #3127: loop less in quest objects. more objective tries

Laumesis:
As little changes as possible. A bit speed up before beta testing. Looks almost bugless while playing. Clicking quest log entries to enable / disalbe questie tracker tracking bugs a bit by affecting wrong quest sometimes.

I have not yet reviewed QuestEventHandler.lua latest version for bug & stuff.


Laumesis:
Added `QuestieDB:GetQuest(questId):Objectives[i].isUpdated = true/false/nil` boolean to skip unnecessary objective updates (`_ObjectiveUpdate`). There is slight chance for race conditions as I didn't check for those. Depending if some part of Questie processes stuff before QuestEventHandler has processed related events fully to recheck hash.

BreakBB (approved):

PR #3135, opened 2021-08-20
fixes to quest-event-refactor

Laumesis:
I adapted fixes from my other branches to fit into the real `quest-event-refactor` branch. I have not yet gone thru all of my fixes, so more may come. I haven't yet tested code in this PR-branch. I will test later once I'm less zombie and comment into this PR. I'm making this PR now so I remember later what I was doing.

Read all commit messages first. Then check each commits code.

There is also one bug fix to master: c38a914b6fee303b1fbe3b21161b9b147cbde63f


Laumesis:
@BreakBB I think this could be merged before things diverge.

Read first all commit messages and then check each commits code.

PR #3148, opened 2021-09-06
[fix] backport fixes from event refactoring development

Laumesis:
IMHO these are worth to backport to master for a p2 release.
1st commit fixes a bug when player has collapsed entries in quest log.
2nd commit removes unnecessary looping on heavy stuff. It could be improved a lot as even Levenshtein calculation can be left out. No such changes now as we want avoid new bugs.


BreakBB:
I absolute agree @Laumesis 👍🏻 Thanks for looking into it.

The #3111 branch still needs some time and work 🙈

PR #3152, opened 2021-09-10
[fix] Remove addon version check. Add addon name check. Enable retail game check

Laumesis:
Review of VersionCheck.lua
 -  Remove addon version check as it didn't do its job and should be soon unnecessary once WoW supports full addon reloading at `/reload`. Hopefully CurseForge gets their shit together too and release addon manager supporting multiple tocs okey.
 - Added extra check that Questie really is run from `Questie` folder and with that name. So users won't try to run multiple at once. Even double Questie global is already checked.
 - Enable retail game check which was most likely unintentionally left disabled by old development testing while getting retail/tbc zoneIDs.

### DIFF
--- a/Modules/VersionCheck.lua
+++ b/Modules/VersionCheck.lua
@@ -1,11 +1,13 @@
+local addonName, _ = ...
+
 if GetBuildInfo() <= '1.13' then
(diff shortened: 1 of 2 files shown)


Laumesis:
Changed `cli.lua` to support addon namespace

PR #3153, opened 2021-09-10
Change DEBUG_levels to non-global

Laumesis:
Basicly: `DEBUG_*` -> `Questie.DEBUG_*` so we won't pollute globals. Specially with names that someone else might use.
Example: `DEBUG_CRITICAL` -> `Questie.DEBUG_CRITICAL`

And some additional changes to `Questie.lua`:
- Modify `Questie:Debug( )` function
- Don't load modules not used in the .lua file
- `local band = bit.band`

And deleting old `DEBUG_CRITICAL`, etc. from globals list in `.luacheckrc`


Questie.lua
```diff
@@ -1,62 +1,31 @@
...
 ---@type QuestieTooltips
 local QuestieTooltips = QuestieLoader:ImportModule("QuestieTooltips");
----@type QuestieTracker
-local QuestieTracker = QuestieLoader:ImportModule("QuestieTracker");
 ---@type QuestieDBMIntegration
 local QuestieDBMIntegration = QuestieLoader:ImportModule("QuestieDBMIntegration");
----@type QuestieLib
-local QuestieLib = QuestieLoader:ImportModule("QuestieLib");
----@type QuestiePlayer
-local QuestiePlayer = QuestieLoader:ImportModule("QuestiePlayer");
```
BreakBB, line 59: Uff, how the hell was this unnoticed for so long? 🙈 Good catch Laume! 👍🏻

BreakBB (approved):

Questie.lua
Laumesis: Because I removed need for those only a month ago <https://github.com/Questie/Questie/commit/d8b545bd7cbcb5887b2c5066d62fef1d485ecbb5> 🤫

PR #3155, opened 2021-09-12
[fix] Don't open Quest Log in combat from tracker.

### LINKED ISSUE #2854: incombat can't click Tarcked Quest button
click tracked quest list  can show QuestLogFrame
but in combat is bad

because HideUIPanel/ShowUIPanel Protected during combat(by [Patch 8.2.0 / API changes](https://wowpedia.fandom.com/wiki/API_ShowUIPanel) )

so,can fix it?
like :
function S_ShowUIPanel(frame, force)
	if ( frame and (InCombatLockdown() or issecure() )) then
		frame:Show();
	else
		ShowUIPanel(frame, force);
	end
end

Laumesis:
ShowUIPanel(  ) is protected, so it can't be used while in combat.
Calling QuestLogFrame:Show() directly would mess up UI.
Select the quest in questlog anyway and try scroll to it.
Scrolling has old bug if the quest is inside collapsed, but I didn't find good way to fix it now.
If we couldn't open quest log, tell user open it themself.

Fixes #2854

Changes aren't tested  with other quest log addons mentioned in code

### DIFF
--- a/Modules/Tracker/QuestieTrackerUtils.lua
+++ b/Modules/Tracker/QuestieTrackerUtils.lua
@@ -17,7 +17,7 @@ local tinsert = table.insert
 function QuestieTracker.utils:ShowQuestLog(quest)
     -- Priority order first check if addon exist otherwise default to original
     local questFrame = QuestLogExFrame or ClassicQuestLog or QuestLogFrame;
-    HideUIPanel(questFrame);
+    --HideUIPanel(questFrame); -- don't use as I don't see why to use and protected function taints in combat
     local questLogIndex = GetQuestLogIndexByID(quest.Id);
     SelectQuestLogEntry(questLogIndex)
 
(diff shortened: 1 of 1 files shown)


BreakBB (approved):

Modules/Tracker/QuestieTrackerUtils.lua
```diff
@@ -17,19 +17,25 @@ local tinsert = table.insert
...
 
-    --Addon specific behaviors
-    if(QuestLogEx) then
-        QuestLogEx:Maximize();
+            --Addon specific behaviors
+            if(QuestLogEx) then
+                QuestLogEx:Maximize();
+            end
+        else
+            Questie:Print("Can't open Quest Log while in combat. Open it manually.")
```
Laumesis, line 37: I forgot to addon translations for this. Also I'm not yet sure how translations work.

BreakBB: Oh right. You basically just add a table entry into [the best fitting file](https://github.com/Questie/Questie/blob/master/Localization/Translations/DebugMessages.lua) and add all the translations you can. For unknown translations add `false`.

PR #3161, opened 2021-09-16
[fix] Fix daily quests showing up again immediately after turn in

BreakBB:


Laumesis:
Oh and, are you sure there isn't adverse effects from returning all real dailies to be daily quests? I am mainly thinking about seasonal quests.

PR #3168, opened 2021-09-20
Refactor DailyQuest.lua

Laumesis:
- Refactor whole `DailyQuest.lua`
 - Fix non-random everyday dailies showing always available - even right after turn in
 - "Old" random dailies should work as before.
 - Reset non-radom everyday dailies to be available again at server's daily reset

### DIFF
--- a/Modules/Quest/DailyQuests.lua
+++ b/Modules/Quest/DailyQuests.lua
@@ -1,43 +1,39 @@
 ---@class DailyQuests
-local DailyQuests = QuestieLoader:CreateModule("DailyQuests");
+local DailyQuests = QuestieLoader:CreateModule("DailyQuests")
 local Questie = _G.Questie
 
 local IsQuestFlaggedCompleted = IsQuestFlaggedCompleted or C_QuestLog.IsQuestFlaggedCompleted
 
 ---@type QuestieMap
-local QuestieMap = QuestieLoader:ImportModule("QuestieMap");
+local QuestieMap = QuestieLoader:ImportModule("QuestieMap")
 ---@type QuestieQuest
-local QuestieQuest = QuestieLoader:ImportModule("QuestieQuest");
+local QuestieQuest = QuestieLoader:ImportModule("QuestieQuest")
 ---@type QuestieTooltips
-local QuestieTooltips = QuestieLoader:ImportModule("QuestieTooltips");
+local QuestieTooltips = QuestieLoader:ImportModule("QuestieTooltips")
 ---@type QuestiePlayer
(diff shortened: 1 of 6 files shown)


Modules/Quest/DailyQuests.lua
```diff
@@ -205,4 +249,85 @@ pvpDailyIds = {
     [11340] = true,
     [11341] = true,
     [11342] = true,
-}
\ No newline at end of file
+}
+
+everydayDailyIds = {
```
BreakBB, line 254: I don't think it is a good idea to have all these IDs in here. They should come from the database and most likely be checked by the `questFlags` field as suggested in #3161

Modules/Quest/DailyQuests.lua
```diff
@@ -1,157 +1,200 @@
...
+---@return nil
+local function _ResetEverydayDailyQuests()
+    Questie:Debug(Questie.DEBUG_INFO, "[DailyQuests]: ResetEverydayDailyQuests() - Okey to get run also at non-reset times")
+
+    local dbCharCompletedQuests = Questie.db.char.complete
+
+    for questId, _ in pairs(everydayDailyIds) do
+        if not IsQuestFlaggedCompleted(questId) then
+            -- Mark the quest uncompleted so it can be available again
+            dbCharCompletedQuests[questId] = nil
```
BreakBB, line 54: Why do we need to do this manually? `Questie.db.char.complete` is filled on QuestieInit by calling the API for all completed quests.

Or does this strange timer code also handles the daily quest reset without a relog? 🤔

Laumesis: Yes it handles daily quest reset on the fly without relog or reload or smoothreset. So far only for general (non random) daily quests. (Not yet for random dailies which only one is available per type at once)

Modules/Quest/DailyQuests.lua
Laumesis: I hope they will come in future from DB, so that `everydayDailyIds` (or renamed `generalDailyIds`) is auto generated at database recompile. As well as other dailyIds.

Modules/Quest/DailyQuests.lua
BreakBB: Guess I have to dig deeper then, to fully understand the timer. Sounds great 👍🏻

Laumesis: It is still WIP now as I started to implement more stuff

BreakBB:
Closing this as the module needs to be rewritten either way.

PR #3174, opened 2021-09-24
Support different seasonal event dates per expansion (#3174)

Laumesis:
Support different seasonal event dates or fully missing event per expansion.
Fix Harverst Festival dates for year 2021.
Disable Brewfest on Classic Era.

Squash merge with this commit as message.

### DIFF
--- a/Database/Corrections/QuestieEvent.lua
+++ b/Database/Corrections/QuestieEvent.lua
@@ -30,6 +30,21 @@ Day of the Dead    1st Nov - 2nd Nov    Day of the Dead
 WoW's Anniversary    16th Nov - 30th Nov
 Pilgrim's Bounty    22nd Nov - 28th Nov    Thanksgiving
 Feast of Winter Veil    15th Dec - 2nd Jan    Christmas
+
+Harvest Festival history:         lunar calendar 15/8 in gregorian calendar:
+2009: Su-Sa 27/9 - 3/10 (wowpedia)    3/10
+2010: Th-We 16/9 - 22/9 (wowpedia)   22/9
+2011: Tu-Mo  6/9 - 12/9 (wowpedia)   12/9
+2012: Mo-Mo 24/9 - 1/10 (wowpedia)   30/9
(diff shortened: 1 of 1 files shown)


Database/Corrections/QuestieEvent.lua
```diff
@@ -58,6 +73,13 @@ function QuestieEvent:Load()
     QuestieEvent.eventDates["Lunar Festival"] = QuestieEvent.lunarFestival[year]
     local activeEvents = {}
 
+    local eventCorrections = Questie.IsTBC and QuestieEvent.eventDateCorrections["TBC"] or QuestieEvent.eventDateCorrections["CLASSIC"]
+    for eventName,dates in pairs(eventCorrections) do
+        if dates then
+            QuestieEvent.eventDates[eventName] = dates
```
BreakBB, line 79: You set `["Brewfest"] = false,` below so `dates = false`.

The next loop should break because of this, since it iterates over `QuestieEvent.eventDates` and wants to extract two values. Won't `eventData.startDate` will be `nil` and then errors on `eventData.startDate`?

Laumesis: `if dates then` -> it never sets `QuestieEvent.eventDates[eventName] = false`

BreakBB: Oh right, but how is that then removing Brewfest for Classic Era?

BreakBB: Ah nevermind. Found the `and (eventCorrections[eventName] ~= false)` 🤦🏻

Laumesis: This first block was just to change dates to correct ones.

Non-existing events are removed in next changed block:
`if _WithinDates(startDay, startMonth, endDay, endMonth) and (eventCorrections[eventName] ~= false) then`

There was some reason to not skip it fully at first changed block, but I can't remember anymore why. It would have created possible bug

BreakBB (approved):

PR #3194, opened 2021-10-06
[quest] Fix Darkrune turn in quests in Ogrila

Laumesis:
11027 doesn't seem to be in game. Only 11060 versio of it - not both
A Crystalforged Darkrune (11060) fixed to repeatable


Laumesis:
wowhead links:
https://tbc.wowhead.com/quest=11027/yous-have-da-darkrune
https://tbc.wowhead.com/quest=11060/a-crystalforged-darkrune

PR #3205, opened 2021-10-22
Only add map icons for tracked quests

svalentin:
For my personal needs I don't like Questie showing map icons for untracked quests. I made this change for myself but I figured others might find it useful too.

I've tested it with wow classic TBC. Haven't tested all types of quests though. For example, I haven't tested dungeon quests or daily quests.

In terms of performance, it avoids full resets so it shouldn't cause any problems.


svalentin:
Friendly ping!

BreakBB (approved):
This looks great, thanks @svalentin 👍🏻

svalentin:
Awesome!
Though remember to merge the PR! I don't have write access. :)

PR #3232, opened 2021-11-08
[feature] Add SoM support

BreakBB:
This PR is supposed to add support for SoM.

Currently there is not API which helps doing so on addon load, but there is a Buff which can be used to determine if the player is on an SoM realm:

```lua
local IsClassicSeason = select(10, UnitAura("player", 1)) == 362859;
```

For now the PR uses a hard coded phase which is planned to be replaced by an API call if one will become available.

We also know that in Phase 1 of SoM the following quests will be available on top of all the ones which were available on Classic release:

* Dire Maul quests
* T0,5 quests
* Level 50 class quests (which might all be related to T0,5)

TODOs:
- [x] Add basic blacklisting of quests based on a new table
- [x] Use correct Blizzard API
- [x] Correct quest XP
- [x] Add all quests to the new table, which will become available at a later phase (see GIT history)
- [x] ~~Add check to determine if the current realm is a SoM realm (and store it in `Questie.db.char.isSoMCharacter`)~~ (not required anymore since we have an official API


Laumesis:
Current latest Era PTR v1.4.1.41009 has:
`C_Seasons.HasActiveSeason()` returns boolean
`C_Seasons.GetActiveSeason()` returns integer (I am not sure if nil is possible)
`Enum.SeasonID = { NoSeason=0, SeasonOfMastery=1 }` integer matches GetActiveSeason() returns.

```lua
local isSoM = C_Seasons and C_Seasons.GetActiveSeason and C_Seasons.GetActiveSeason() == Enum.SeasonID.SeasonOfMastery
```
Works okey on both SoM and Era realms on PTR.

Laumesis:
Uh. Brain stopped working. Need sleep. This is what I wrote before :|

I have not yet looked your code at all.
- Not sure if we need to store SoM into character
- Maybe have something like:  (don't use dates and keys in example)
```lua
--- use human readable key values
expansionsSeasonsPhases = {
  WOW_PROJECT_CLASSIC = {
    0 = { -- Era, 0 == Enum.SeasonID.NoSeason  on Classic_Vanilla client.
      { start = "2020-12-01", key = "EraP6" }, -- date doesn't matter anymore, as long as it is in past
    },
    1 = { -- SoM, 1 == Enum.SeasonID.SeasonOfMastery  on Classic_Vanilla client.
      -- MUST be sorted ascending by start field 
      { start = "2021-12-31", key = "SoMP1" },
      { start = "2022-02-01", key = "SoMP2" },
      { start = "2022-02-15", key = "SoMP2NewInstanceWithQuestsOpens" },
      { start = "2022-05-01", key = "SoMP3" },
      { start = "2022-08-01", key = "SoMP4" },
      { start = "2022-12-01", key = "SoMP5" },
    },
  },
  WOW_PROJECT_BURNING_CRUSADE_CLASSIC = {
    0 =  { -- Current TBC, possibly future TBC Era? No Enum available atm.
      { start = "2021-06-01", key = "TBCP1" },
      { start = "2021-07-01", key = "TBCnewquestsopened" },
      { start = "2021-08-01", key = "TBCnewraidsopened" },
      { start = "2021-08-24", key = "TBCnextstuffopens" },
    }
  },
}

local realmTime = << get current realm time >>
local season = C_Seasons and C_Seasons.GetActiveSeason() or 0 -- TBC client doesn't have this yet
local phases = expansionsSeasonsPhases[WOW_PROJECT_ID][season]
for i=1, #phases do
  local start = phases[i].start
  if << compare realmTime and start -> if start is earlier then >> then

  end
end
```

BreakBB:
Since we now have an API call we don't need to store information about SoM in the SVs 👍🏻 

I am not sure about the dates. We have the API to get the active season, so no need to care for them, right? It could also potentially raise issues, when different servers will have different seasons active and we only support one fixed date. The API should also help to make everything way more dynamic :)

PR #3237, opened 2021-11-09
Fix quest 123 "The Collector" starters

gouletr:
NPC `100` "Gruff Swiftbite" does not directly starts quest `123` "The Collector", it is obtained from item `1307` "Gold Pickup Schedule" drop solely


BreakBB (approved):
Wow, no idea how that stayed in there for so long. You are absolutely right about it and I think it already is like this in our TBC data.

Thanks @gouletr

PR #3248, opened 2021-11-13
Improvements to Quest event refactor PR

Laumesis:
I have not tested these enough for merge, but they most likely are still okey for that.

### DIFF
--- a/Modules/Quest/QuestEventHandler.lua
+++ b/Modules/Quest/QuestEventHandler.lua
@@ -350,6 +350,13 @@ function _QuestEventHandler:BankFrameClosed()
     end
 end
 
+function _QuestEventHandler:ChatMsgCombatFactionChange()
+    print("[Event] CHAT_MSG_COMBAT_FACTION_CHANGE")
+
+    --- Reputation level changes changing quest progression don't fire UNIT_QUEST_LOG_CHANGED event, only QUEST_LOG_UPDATE event.
+    doFullQuestLogScan = true
+end
+
 --- Helper function to insert a callback to the questLogUpdateQueue and increase the index
 function _QuestLogUpdateQueue:Insert(func)
     questLogUpdateQueue[questLogUpdateQueueSize] = func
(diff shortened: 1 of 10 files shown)


Laumesis:
Note to Self: Last commits should be checked as I was half sleep.

`UNIT_FACTION` event is not good either as it fires too often. Has to look for better solution.

Modules/Quest/QuestEventHandler.lua
```diff
@@ -70,6 +73,28 @@ function _QuestEventHandler:InitQuestLog()
         return false
     end
 
+    for i = 1, MAX_QUEST_LOG_INDEX do
```
BreakBB, line 76: This loop should be merged with the the loop below. Both are partly doing the same

Modules/Quest/QuestieQuest.lua
```diff
@@ -444,7 +421,7 @@ function QuestieQuest:AbandonedQuest(questId)
 
         local quest = QuestieDB:GetQuest(questId);
         if quest then
-            quest.Objectives = {};
+            quest.Objectives = nil;
```
BreakBB, line 424: I am not sure I like this change. I think it was just recently changed so a table, right? But an empty table is just as good as `nil` from the code logic perspective, but is also saver to not result in an error.

Modules/Quest/QuestEventHandler.lua
BreakBB: We will keep these separate to make the code more readable. They are also run only on addon load so performance does not matter.

Laumesis:
`UNIT_FACTION` seems okey to use if all reputation quest objectives require only standing change. can't be used for windsaber quests availability. I don't know where I saw it fire too much for `player` if I ever saw and not mixed it to something else.

TODO: init at login / reload should be tested better - specially with empty quest log.  And it could be implemented better.

BreakBB:
Most of these changes are now already included in #3111

Laumesis:
Closing this as all important changes are in #3111 now. I left few review comments.

Added fake debug level (to filter most spammy spam messages away) is not "merged" into #3111 and should be handled elsewhere in future.

PR #3272, opened 2021-11-29
Update tbcQuestFixes.lua

Duugu:
Removed preQuestSingle for 1204. There's no pre-quest required in tbc.

### DIFF
--- a/Database/Corrections/TBC/tbcQuestFixes.lua
+++ b/Database/Corrections/TBC/tbcQuestFixes.lua
@@ -91,6 +91,9 @@ function QuestieTBCQuestFixes:Load()
         [1135] = {
             [questKeys.startedBy] = {{4456},nil,nil},
         },
+        [1204] = {--no required pre-quest anymore with tbc
+            [questKeys.preQuestSingle] = {},
+        },        
         [1437] = {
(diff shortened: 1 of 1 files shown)


Laumesis:
Thank you for PR, there are two problems:

1. You should also mark the removed preQuest quest as a breadcrump quest so that it won't show available anymore after 1240.
2. Seems to work correctly already. Screenshot crops from TBC, lvl 40 character:
image
Because this is already fixed in Classic and TBC loads also Classic fixes (I didn't know myself it loads also Classic):
https://github.com/Questie/Questie/blob/aa9a6f7e78fbe124201da0d83bbb2c585d705cd1/Database/Corrections/Classic/classicQuestFixes.lua#L578-L580
https://github.com/Questie/Questie/blob/aa9a6f7e78fbe124201da0d83bbb2c585d705cd1/Database/Corrections/Classic/classicQuestFixes.lua#L596-L598
Those are from commits: https://github.com/Questie/Questie/commit/cb15374ae6fd76432857dab4e601aa13766d0d36 https://github.com/Questie/Questie/commit/771dca166aeefb86a2e436a2e1ce5e46f73b6a90

So this change is not necessary at all. I wonder if there are some other requirements broken if you experienced this to be available while Questie didn't show so? Or if this was some kind of issue after gaining an level :(

PR #3273, opened 2021-11-29
[feature] added option to announce quest acceptance, abortion or completion to party

lbuchholz79:
Usually my wife and I play together in a group, but we get torn apart sometimes (by those nasty herb and mining nodes :) ), so one of us might miss if the other one took a new quest or simply overheard that a tiresome quest was abandoned. So I thought it would be nice to get informed every time, a party member accepts, abandons or hands-in a quest. 

Initially I was hoping to find a existing addon like QuestAnnounce3, but it does not work with the recent client version. Also it seems to be too simple does not actually query the Quest API. I knew Questie announces progress steps already, so I thought it would be simple just to add a couple of options to notify the group on these events too.

Here is what I came up with, so you can review it, please. I hope I found the right spots and did not spread the code too much. Also only german translation is provided atm (sry, not a multilingual genius). I'm looking forward to when this becomes an official feature of Questie.


Modules/QuestieAnnounce.lua
```diff
@@ -32,20 +28,24 @@ function QuestieAnnounce:AnnounceParty(questId, progressType, itemId, objectiveT
...
         end
 
-        alreadySentBandaid[message] = true
+        _QuestieAnnounce:AnnounceParty(message, false)
+    end
+end
 
-        SendChatMessage(message, "PARTY")
+function _QuestieAnnounce:AnnounceParty(message, allowRepeat)
+    if (not message) or ((allowRepeat == nil or not allowRepeat) and alreadySentBandaid[message]) then
```
BreakBB, line 42:
```suggestion
    if (not message) or (not allowRepeat) and alreadySentBandaid[message]) then
```

The `not` operator returns `true` for everything that is falsy. So the condition before checked the same twice.

Modules/Options/GeneralTab/QuestieOptionsGeneral.lua
```diff
@@ -339,10 +339,54 @@ function QuestieOptions.tabs.general:Initialize()
...
+                        width = 1.5,
+                        disabled = function() return (not Questie.db.char.questAnnounce); end,
+                        get = function () return Questie.db.char.questAnnounceAccepted; end,
+                        set = function (info, value)
+                            Questie.db.char.questAnnounceAccepted = value
+                        end,
+                    },
+                    questAnnounceAbandoned = {
+                        type = "toggle",
+                        order = 1,
```
BreakBB, line 362:
```suggestion
                        order = 2,
```

Modules/Options/GeneralTab/QuestieOptionsGeneral.lua
```diff
@@ -339,10 +339,54 @@ function QuestieOptions.tabs.general:Initialize()
...
+                        width = 1.5,
+                        disabled = function() return (not Questie.db.char.questAnnounce); end,
+                        get = function () return Questie.db.char.questAnnounceAbandoned; end,
+                        set = function (info, value)
+                            Questie.db.char.questAnnounceAbandoned = value
+                        end,
+                    },
+                    questAnnounceCompleted = {
+                        type = "toggle",
+                        order = 1,
```
BreakBB, line 374:
```suggestion
                        order = 3,
```

BreakBB (changes requested):
Hey @lbuchholz79 thank you very much for taking your time and looking into Questie <3

This feature sounds indeed quite nice and the code looks mostly alright. I left you a few really minor notes 😁 

Currently we are not merging new features, since we first want to finally finish #3111 and release it. Since that is quite a critical rebuild we want to keep the release "as clean as possible" to reduce the chance of side effects.

But you can bet, we will merge this 👍🏻

Modules/QuestieAnnounce.lua
```diff
@@ -86,3 +86,43 @@ function QuestieAnnounce:ItemLooted(text, notPlayerName, _, _, playerName)
     end
 end
 
+---@return table<string, number, string>
+local function _GetQuestInfos(questId)
```
BreakBB, line 90: You should define these functions at the top of the file first so they are defined before they are used.

At the top you can simply do:

```lua
local _GetQuestInfos
```

and then here it will become:

```suggestion
_GetQuestInfos = function(questId)
```

Modules/QuestieAnnounce.lua
lbuchholz79: True indeed, in my first draft it the flag was initially negated, so the condition made more sense.

lbuchholz79:
Thanks for the feedback and I already looking forward for the final release :D
I'll incorporate your fixes, run some testing and push changes tomorrow.

lbuchholz79:
I have some more improvements in mind for follow-up tickets:
1. rename original announce option to something like "announce progress steps", so it better fits in line with the events.
2. adding a master announcement toggle (currently events use "announce progress steps" for this 
3. option to announce in group, raid or both (or even include "none" to act as master toggle)
   this requires some work on the options panel and settings storage (incl. migration)

Do you think it's worth I start working on that?

BreakBB:
Hey @lbuchholz79 I am sorry for letting you down on this month ago. If you are still willing to put time into this, it would be great if you could review this PR again, resolve the merge conflicts and then I will be very happy to merge this along with your other suggestions!

lbuchholz79:
Sure, I can do that, shouldn't be a big deal.

lbuchholz79:
I got merge conflicts too, so I rebased manually. Also included announcement to different channels as proposed in #3286 and run a few tests. Seems to work fine so far, However GitHub still thinks a change request is pending.

lbuchholz79:
@BreakBB would you mind reviewing again?

Modules/QuestieAnnounce.lua
```diff
@@ -77,13 +99,42 @@ function QuestieAnnounce:ItemLooted(text, notPlayerName, _, _, playerName)
...
 
-            QuestieAnnounce:AnnounceParty(startQuestId, "item", itemId)
+            _QuestieAnnounce:AnnounceSelf(startQuestId, itemId)
         end
     end
 end
 
+function QuestieAnnounce:AcceptedQuest(questId)
+    if (_QuestieAnnounce:AnnounceEnabledAndPlayerInChannel()) and Questie.db.char.questAnnounceAccepted then
+        -- TODO test if this works if quest hasn't yet been accepted by other party members
```
BreakBB, line 114: Did you have the time to complete this TODO?

BreakBB (approved):
Hey @lbuchholz79 the changes look really, really good! Thanks for this PR and sorry again that it was on hold for so long 👍🏻

I just need an answer about that TODO and then we are good to go.

Modules/QuestieAnnounce.lua
lbuchholz79: Ah, I remember. I wanted to use `QuestieLink:GetQuestHyperLink(questId)` in the following line, but I got the error message ```
SendChatMessage(): Invalid escape code in chat message
```
so I reverted back to `QuestieLink:GetQuestLinkString`.
I also observed I cannot click the accepted quest link in my own messages, while party members can. Intriguingly the same method produces a clickable link for abandoned quests. I strongly believe there is something odd in the link syntax.
For know I'd consider this a minor issue.

lbuchholz79:
Had another round of tests and found an error when group messages got sent in raids, which isn't intended since there is a dedicated raid setting for this. 
Also tested announcements with unaccepted quests for other members, didn't pose a problem.
@BreakBB ready for merge

BreakBB (approved):

PR #3275, opened 2021-12-04
[quest] Change "The Broken Sigil" quest link text to match description

pgdagenais:
I am not sure this is the correct place to do the modification but the description in the quest link is incorrect.

Looks like it is using the tbc/old quest description: 

- Retail: https://www.wowhead.com/quest=640/the-broken-sigil
- TBC: https://tbc.wowhead.com/quest=640/the-broken-sigil

Compared to the new one: https://classic.wowhead.com/quest=640/the-broken-sigil

If there is anything else I should modify please tell me and I'll update the MR accordingly.

Thanks!


BreakBB:
Hey @pgdagenais Thanks for taking the time to do a database correction for Questie!

Actually we have a [Corrections folder](https://github.com/Questie/Questie/tree/master/Database/Corrections) which contains different files for the corrections. The database files themself are generated so changes might get lost in the future.

For your correction you want to checkout the [classicQuestFixes.lua](https://github.com/Questie/Questie/blob/master/Database/Corrections/Classic/classicQuestFixes.lua) file.

pgdagenais:
@BreakBB Thanks for the update!

I updated the MR accordingly.

BreakBB (approved):
Hey @pgdagenais that was really fast! 💪🏻 Thank you very much, we appreciate every help and PR we get to improve Questie 😁

PR #3279, opened 2021-12-07
Updated the NPC pathing coordinates for the XT:4, part of the Shredding Machines quest in Stonetalon Mountains

### LINKED ISSUE #3278: Shredding Machines XT:4 path on map is incorrect
## Description
For this quest:

https://tbc.wowhead.com/quest=1068/shredding-machines

You have to kill 2 NPCs. Both XT:9 and XT:4. Both of them Questie draws a path on the world map to show where they walk around. This one: https://tbc.wowhead.com/npc=4073/xt-4 has an incorrect path drawn:

## Wowhead link
https://tbc.wowhead.com/npc=4073/xt-4

You can see on Wowhead the area is much larger as well.

## Screenshots
image
You can see in this screenshot the path on the map that Questie draws and you can see where I see the NPC walking. You can tell that he walks in a much larger area than what Questie draws.

## Questie version
Latest

Jakobud:
This fixes #3278 

Here is the new, more accurate path for the NPC:

image


BreakBB:
Hey @Jakobud 

The updated path looks indeed way more accurate! But instead of changing the generated database files it would be great if you could add the updated path to the [classicNPCFixes.lua](https://github.com/Questie/Questie/blob/master/Database/Corrections/Classic/classicNPCFixes.lua). The corrections done to the Classic database are also loaded for the TBC database, so this one place is enough to fix both versions 💪🏻

Jakobud:
Oh okay, so can you clarify this for me a bit? The files I edited are automatically generated? How are they generated? Ideally I should leave those untouched and just put it in the fixes?

BreakBB:
Sure, I am happy to explain it to you @Jakobud 👍🏻 

The database files were generated on base of [cmangos](https://github.com/cmangos), which is the reverse engineered version of WoW (database + server). We have a script to grab selected database entries/fields and extract them into a format we can use in Questie. So in the future we might want to grab another extra field or simply run a more complicated update, which will then replace the database files.

To make sure fixes we do don't get lost, we added the correction related files, which will always override what is in the base database.

So yes, it would be great if you could revert the changes you did in the database files and move them into the corrections file. And as I wrote before: The [classicNPCFixes.lua](https://github.com/Questie/Questie/blob/master/Database/Corrections/Classic/classicNPCFixes.lua) is what you want to look into. You want to add an entry in a format like this (but you need to update to your NPC ID and insert your zone + waypoints:

```lua
        [2449] = {
            [npcKeys.waypoints] = {[zoneIDs.HILLSBRAD_FOOTHILLS]={{{32.5,35.8},{32.5,37.6},{32.5,38.9},{32.4,40.7},{32.1,41.8},{32.1,43.4},{32.5,44.4},{32.6,45.3},{32,46.5}},{{36.7,38},{35.5,38.9},{34.2,39.3},{32.8,39.9},{32.1,41.8},{32.0,43.1},{31.6,44.6},{30.3,44.4},{29.9,42.6}}}},
        },
```

Jakobud:
Closing in favor of #3287

PR #3287, opened 2021-12-11
Updated XT:4 NPC Waypoints and spawn point in fix file

Jakobud:


BreakBB:
This looks great @Jakobud thanks 👍🏻

PR #3307, opened 2022-01-13
Refactor Game's cached quest log validation

Laumesis:
- Validate game's cached quest log asyncronously
- Make Questie's init flow to pause and wait valid quest log for parts needing it.
- Has QuestieInit.lua modified to support easier refactor of whole init flow, which is a bit mess now (even without PR-3111).

Note: Required test for future in PTR when blizzard is going to release a new phase - as those usually may have bigger changes to how things work.

1. Delete game cache. For example the directory: `...\_classic_era_\Cache\`.
2. Login a character having many quests in quest log.
3. /reload
4. Logout and login the character again.
5. Login another character having 0 quests in quest log. (Or if you use same character, then delete game cache again after emptying the quest log).
6. /reload
7. Logout and login the character again.

### DIFF
--- a/Modules/QuestieInit.lua
+++ b/Modules/QuestieInit.lua
@@ -133,15 +133,13 @@ QuestieInit.Stages[1] = function() -- run as a coroutine
 
     QuestieCleanup:Run()
 
-    coroutine.yield()
-
     -- continue to next Init Stage
     return QuestieInit.Stages[2]
 end
 
 QuestieInit.Stages[2] = function() -- not a coroutine
     _QuestieInit.WaitForGameCache()
-    -- continues to next Init Stage from above function
+    -- continues to next Init Stage in above function
 end
 
 QuestieInit.Stages[3] = function() -- run as a coroutine
@@ -357,7 +355,7 @@ function _QuestieInit:StartStageCoroutine(stage)
                 initFrame:SetScript("OnUpdate", nil)
                 initFrame:SetParent(nil)
                 initFrame = nil
-                if type(ret) == "function" then -- continue to next stage, which was returned by coroutine
+                if type(ret) == "function" then -- continue to next stage, if it was returned by coroutine
                     ret()
                 end
             end
--- a/Modules/QuestieInit.lua
+++ b/Modules/QuestieInit.lua
@@ -47,10 +47,22 @@ local ChatFilter = QuestieLoader:ImportModule("ChatFilter")
 ---@type Hooks
 local Hooks = QuestieLoader:ImportModule("Hooks")
 
--- initialize all questie modules
--- this function runs inside a coroutine
-function QuestieInit:InitAllModules()
+-- 3 * (Max possible number of quests in game quest log)
+-- This is a safe value, even smaller would be enough. Too large won't effect performance
+local MAX_QUEST_LOG_INDEX = 75
 
+local MAX_INIT_QUEST_LOG_TRIES = 5
+local initQuestLogTries = 0
+local eventFrame
+
+
+-- ********************************************************************************
+-- Start of QuestieInit.Stages ******************************************************
+
+-- stage worker functions. Most are coroutines.
+QuestieInit.Stages = {}
+
+QuestieInit.Stages[1] = function() -- run as a coroutine
     HBDHooks:Init()
 
     QuestieFramePool:SetIcons()
@@ -118,6 +130,15 @@ function QuestieInit:InitAllModules()
 
     QuestieCleanup:Run()
 
+    -- continue to next Init Stage
+    coroutine.yield(QuestieInit.Stages[2])
+end
+
+QuestieInit.Stages[2] = function() -- not a coroutine
+    _QuestieInit.WaitForGameCache()
+end
+
+QuestieInit.Stages[3] = function() -- run as a coroutine
     -- register events that rely on questie being initialized
     QuestieEventHandler:RegisterLateEvents()
     QuestEventHandler:RegisterEvents()
@@ -185,6 +206,11 @@ function QuestieInit:InitAllModules()
     end
 end
 
+-- End of QuestieInit.Stages ******************************************************
+-- ********************************************************************************
+
+
+
 function QuestieInit:LoadDatabase(key)
     if QuestieDB[key] then
         coroutine.yield()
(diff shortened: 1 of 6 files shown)


Modules/QuestieInit.lua
```diff
@@ -226,20 +255,37 @@ function _QuestieInit:OverrideDBWithTBCData()
...
+    local function InitOnUpdate()
+        local success, ret = coroutine.resume(routine)
         if success then
             if coroutine.status(routine) == "dead" then
                 initFrame:SetScript("OnUpdate", nil)
+                initFrame:SetParent(nil)
+                initFrame = nil
+                if type(ret) == "function" then -- continue to next stage, if it was returned by coroutine
+                    ret()
+                end
```
BreakBB, line 274: This is optional, because some stages want to return a function which can/has to run right after the previous stage completed?
Can't the init stages just be looped or all functions be started as coroutine? 🤔

Modules/QuestieInit.lua
```diff
@@ -118,6 +125,23 @@ function QuestieInit:InitAllModules()
...
+end
+
+QuestieInit.Stages[2] = function() -- not a coroutine
+    -- Wait for Game Cache's Questlog be cached okey by the game.
+
+    if QuestieValidateGameCache.IsCacheGood() then
+        -- continue to next Init Stage
+        _QuestieInit:StartStageCoroutine(3)
+    else
+        -- tell to game cache checker how to continue to next Init Stage
```
BreakBB, line 139:
```suggestion
        -- QuestieValidateGameCache will take care of starting the next init stage
```

Modules/QuestieValidateGameCache.lua
```diff
@@ -0,0 +1,145 @@
...
+local tunpack = table.unpack or unpack
+
+-- 3 * (Max possible number of quests in game quest log)
+-- This is a safe value, even smaller would be enough. Too large won't effect performance
+local MAX_QUEST_LOG_INDEX = 75
+
+local eventFrame
+local numberOfQuestLogUpdatesToSkip
+
+local cacheGood = false
```
BreakBB, line 27:
```suggestion
local isCacheGood = false
```

Modules/QuestieValidateGameCache.lua
```diff
@@ -0,0 +1,145 @@
...
+        if initQuestLogTries < MAX_INIT_QUEST_LOG_TRIES then
+            print("--> QuestieValidateGameCache.OnQuestLogUpdate()", GetTime(), "Game's quest log is empty. Tries:", initQuestLogTries.."/"..MAX_INIT_QUEST_LOG_TRIES)
+            _WaitForGameCache_TryAgainAtEvent()
+            return
+        else
+            print("--> QuestieValidateGameCache.OnQuestLogUpdate()", GetTime(), "Decide player has no quests for real.")
+        end
+    end
+    ]]--
+
```
BreakBB, line 77: Can this be removed or is it a WIP comment?

Modules/QuestieValidateGameCache.lua
```diff
@@ -0,0 +1,145 @@
...
+    local isGameCacheGood = true
```
BreakBB, line 78:
```suggestion
    local isQuestLogFullyCached = true
```
I think this name is better because it is not as easy to confuse with the global `cacheGood`

Modules/QuestieValidateGameCache.lua
```diff
@@ -0,0 +1,145 @@
...
+
+    print("--> QuestieValidateGameCache.OnQuestLogUpdate()", GetTime(), "Game's quest log is |cff00bc32".."OK".."|r. Good quest: "..goodQuestsCount.."/"..numQuests)
+
+    DestroyEventFrame()
+
+    cacheGood = true
+
+    -- Call all callbacks
+    while (#callbacks > 0) do
+        local callback = table.remove(callbacks, 1)
```
BreakBB, line 123: Should `table.remove` also get a local variable?

Modules/QuestieValidateGameCache.lua
```diff
@@ -0,0 +1,145 @@
...
+        local func, args = tunpack(callback)
+        print("--> QuestieValidateGameCache.OnQuestLogUpdate()", "Calling a callback.")
+        func(tunpack(args))
+    end
+end
+
+local function OnPlayerEnteringWorld(_, _, isInitialLogin, isReloadingUi)
+    assert(isInitialLogin or isReloadingUi)
+
+    numberOfQuestLogUpdatesToSkip = isReloadingUi and 0 or 1
```
BreakBB, line 133:
```suggestion
    numberOfQuestLogUpdatesToSkip = isReloadingUi and 0 or 1 -- We want to skip the first QUEST_LOG_UPDATE at login (fires twice at login and only once when reloading)
```

BreakBB (changes requested):

Modules/QuestieInit.lua
Laumesis: Some stages might not want to be coroutines. For example while cache checking was still its own stage and not done async and earlier, it was not a coroutine

Modules/QuestieValidateGameCache.lua
Laumesis: it is saved as history in case Blizzard decides to change things.  It may get useful then.

Currently it is totally useless. I guess it should have a comment why it is commented out.

Modules/QuestieValidateGameCache.lua
Laumesis: Yes, even doesn't really matter as the line is run on average 0 times currently.

Modules/QuestieValidateGameCache.lua
Laumesis: It may fire many more times. But on login the first QUEST_LOG_UPDATE after PLAYER_ENTERING_WORLD has still invalid cache.

Modules/QuestieValidateGameCache.lua
Laumesis: I am still thinking is starting variable names with `is` a good thing or should that be just for functions/"methods" ?

Modules/QuestieInit.lua
Laumesis: Another option would assume just iterating all stages in order and add some other table to tell about each if it is a coroutine or not.
Current one allows to move execution to where ever, not necessary into next stage. I didn't know what would be best option to do

Modules/QuestieValidateGameCache.lua
Laumesis: Done in 78b16dab24437ff7c33ec1111cd51a01e38c7ea2

Modules/QuestieValidateGameCache.lua
Laumesis: Removed  in 78b16dab24437ff7c33ec1111cd51a01e38c7ea2

Modules/QuestieValidateGameCache.lua
Laumesis: Renamed differently in 78b16dab24437ff7c33ec1111cd51a01e38c7ea2

Modules/QuestieValidateGameCache.lua
Laumesis: Done in 78b16dab24437ff7c33ec1111cd51a01e38c7ea2

Modules/QuestieValidateGameCache.lua
Laumesis: Done in 78b16dab24437ff7c33ec1111cd51a01e38c7ea2 :
Logic changed to show better what we really want to do (functionality is still the same on current game behavior).
Comment added.

BreakBB (dismissed):
Looks good 👍🏻

BreakBB (approved):

PR #3324, opened 2022-01-22
[fix] Show Tracker even without any quest tracked

### LINKED ISSUE #3315: Show tracker header even 0 quests are tracked.
Show tracker header even 0 quests are tracked.

BreakBB:
Fixes #3315

### DIFF
--- a/Modules/Tracker/QuestieTracker.lua
+++ b/Modules/Tracker/QuestieTracker.lua
@@ -1822,9 +1822,6 @@ function QuestieTracker:Update()
         end)
     end
 
-    if (not hasQuest) then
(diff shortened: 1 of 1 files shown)


Modules/Tracker/QuestieTracker.lua
```diff
@@ -1826,7 +1822,11 @@ function QuestieTracker:Update()
         end)
     end
 
-    if hasQuest and _QuestieTracker.IsFirstRun == nil then
+    if (not hasQuest) then
+        QuestieTracker:Collapse()
+    end
```
BreakBB, line 1827:
```suggestion
```

PR #3329, opened 2022-01-24
Update LibQuestXP to 1.0.8 and 2.0.3

Laumesis:
The lib has build in support for SoM quest xp bonus,
so Questie's code for +40% is removed now.
Extra bonus xp for elite and dungeon quests is still missing.

Questie modifications of library's TBC version removed.
It was modified to use `QuestiePlayer:GetPlayerLevel()` while Questie was in debug mode.


Laumesis:
quest xp gain correctness tested in Era, SoM, TBC low level quest.

BreakBB (approved):
I don't think we need the Questie specific modifications, so this PR looks good 👍🏻

PR #3359, opened 2022-02-20
Suggestions to Remove Common Auction House Items to Reduce Map and Tooltip Clutter

Gogo1951:
Adding a few more TBC items to reduce clutter around "commodity" items that can be readily obtained from the Auction House. Also cleaned up TBC list by sorting by Item ID.


BreakBB:
Good suggestions @Gogo1951 thanks 👍🏻

BreakBB (approved):

PR #3378, opened 2022-03-05
Fix tooltips to show objectives and quest accept on the same mob

BreakBB:
The quests to test this are https://tbc.wowhead.com/quest=443/rot-hide-ichor and https://tbc.wowhead.com/quest=460/resting-in-pieces

Currently only the quest accept line is shown on the mob:

grafik

but correct is:
grafik

Things left todo:

- [ ] Add back comms information to also show the progress of other players


BreakBB:
This got handled in b2c228551c25c30a970f72342320a946792ec989

PR #3393, opened 2022-03-28
NPC Blackmoss the Fetid (3535) does not start quest The Moss-twined Heart (927) directly

gouletr:
The NPC [Blackmoss the Fetid (3535)](https://classic.wowhead.com/npc=3535/blackmoss-the-fetid) does not start the quest [The Moss-twined Heart (927)](https://classic.wowhead.com/quest=927/the-moss-twined-heart) directly, it is only obtainable from the item [Moss-twined Heart (5179)](https://classic.wowhead.com/item=5179/moss-twined-heart) dropped by that same NPC.

I fixed something similar some time ago in Elwynn Forest for the quest [The Collector (123)](https://classic.wowhead.com/quest=123/the-collector) here: https://github.com/Questie/Questie/pull/3237. I realize now that this is a hack that was done in order to get quests obtained from item drops to show on the map.

There are probably a bunch more like this, and honestly those hacks should all be removed from the database, because when we write an addon that uses Questie's database, we get these invalid results. What needs to be done in Questie, if that ain't the case already, is to add or fix some code that will find those quests from item drops, and somehow show them on the map without relying on hacking the database. It would actually be quite useful if it was possible to somehow ask the database if a quest is from a drop rather than an NPC, I could use that information in my addon to determine if a quest is somewhat optional.

### DIFF
--- a/Database/Corrections/Classic/classicQuestFixes.lua
+++ b/Database/Corrections/Classic/classicQuestFixes.lua
@@ -448,7 +448,7 @@ function QuestieQuestFixes:Load()
             [questKeys.exclusiveTo] = {924}, -- #2195
         },
         [927] = {
-            [questKeys.startedBy] = {{3535},nil,{5179}},
+            [questKeys.startedBy] = {nil,nil,{5179}},
         },
         [930] = {
             [questKeys.preQuestSingle] = {918}, -- #971


BreakBB:
Hey @gouletr you are right that this change should be reverted along with other cases and is not really correct. It will require some time though to dig into the code and find a solution for it that is not the simple work around we currently use.

gouletr:
Would it be possible to bake-in whether or not a quest is from an item dropped by an npc when compiling the quest database and store the result in that database? This way when you list all quests that should be shown on the map, you could quickly list those and show them at the locations where these npcs have spawns.

BreakBB:
This information is still in the database. Look at the line changed in this PR:

```
            [questKeys.startedBy] = {{3535},nil,{5179}},
```

The `startedBy` field has three tables which are: NPC starters, object startes and item startes. Even though we added the NPC started for a quest, we didn't remove the item starter. And every quest that has a item starter always starts from this item.

gouletr:
> This information is still in the database. Look at the line changed in this PR:
> 
> ```
>             [questKeys.startedBy] = {{3535},nil,{5179}},
> ```
> 
> The `startedBy` field has three tables which are: NPC starters, object startes and item startes. Even though we added the NPC started for a quest, we didn't remove the item starter. And every quest that has a item starter always starts from this item.
> 
> `/dump QuestieDB.QueryQuestSingle(927, "startedBy")[3] ~= nil`

Yes I am aware of that. I was just theory-crafting about the solution that can be implemented in Questie for these quests to show up on the map. Having to go through the database for this information that is somewhat static is not super efficient though, because you have to loop through item starters, then through the npcDrops of these items, and finally get the spawns of these npcs.

gouletr:
Anyway, is my change going in or you want to make a proper fix in Questie first?

BreakBB:
We will keep this PR open as a reminder for us to implement a proper fix.

gouletr:
> Ironically enough, proper tracking/display of dropped item quest starters is something that we're working on internally and is nearing completion.

Oops I wasn't intending to close this PR, I was just cleaning up my branches.

PR #3414, opened 2022-04-15
Update EasternKingdoms.lua

anon1231823:


Localization/Translations/Zones/EasternKingdoms.lua
```diff
@@ -327,14 +327,14 @@ local easternKingdomsLocales = {
         ["zhCN"] = "奥特兰克山脉",
     },
     ["Dalaran"] = {
-        ["ptBR"] = true,
+        ["ptBR"] = "Dalaran",
```
BreakBB, line 330: `true` always means "take the key as translation". So if the key and the translation are the same it is better to leave it as `true`.

anon1231823: That shipped sailed a while ago, every other translation put it in

BreakBB: No reason to make it worse though.

BreakBB:
Thanks for coming back to this @anon1231823 💪🏻 Great work, thank you! :)

PR #3434, opened 2022-05-01
[quest] Enable TBC Children's Week follow up quests

Laumesis:
This commit is not tested in game.
I am not sure if all of follow up quests are included into commit.
The quests data in tbcQuestDB is not checked,
except zoneOrSort is corrected.


Laumesis:
Stuff like in attached images needs fixing. I'm not going to do that.

image
image

Cabro:
It's different quests on horde side, also :)

PR #3440, opened 2022-05-01
[fix] Fix 'GetAllLeaderBoardDetails' error

### LINKED ISSUE #3412: ERROR! Something went wrong in GetAllLeaderBoardDetails with new questie version
## Bug description
After updating to latest version I get these errors on login and sometimes seemingly at random later on.
[22:59:08] Questie: [ERROR] ERROR! Something went wrong in GetAllLeaderBoardDetails1109 -  : 0/1 
[22:59:08] Questie: [ERROR] ERROR! Something went wrong in GetAllLeaderBoardDetails8585 -  : 0/1 
[22:59:08] Questie: [ERROR] ERROR! Something went wrong in GetAllLeaderBoardDetails8585 -  : 0/20 
[22:59:08] Questie: [ERROR] ERROR! Something went wrong in GetAllLeaderBoardDetails8578 -  : 0/1 
[22:59:08] Questie: [ERROR] ERROR! Something went wrong in GetAllLeaderBoardDetails10601 -  : 0/1 
[22:59:08] Questie: [ERROR] ERROR! Something went wrong in GetAllLeaderBoardDetails8620 -  : 0/1 
[22:59:08] Questie: [ERROR] ERROR! Something went wrong in GetAllLeaderBoardDetails1446 -  : 0/1 

## Screenshots
Have these quests in my log.
image

## Questie & WoW version
Questie 6.9.0
TBC Classic

Muehe:
This error happened due to the API returning uninitialised data.

Now, if we detect uninitialised data, we act like nothing happened.

Fixed #3412

### DIFF
--- a/Modules/Libs/QuestieHash.lua
+++ b/Modules/Libs/QuestieHash.lua
@@ -35,7 +35,8 @@ function QuestieHash:InitQuestLogHashes()
                 Questie._sessionWarnings[questId] = true
             end
         elseif (not isHeader) then
-            questLogHashes[questId] = QuestieHash:GetQuestHash(questId)
+            local oldHash = questLogHashes[questId]
+            questLogHashes[questId] = QuestieHash:GetQuestHash(questId) or oldHash
         end
     end
 end
@@ -49,6 +50,12 @@ function QuestieHash:GetQuestHash(questId)
         isComplete = QuestieDB:IsComplete(questId),
         questObjectives = C_QuestLog.GetQuestObjectives(questId),
     }
+    -- Catch uninitialised objectives
+    for _, objective in pairs(data.questObjectives) do
+        if (objective.text and (string.sub(objective.text,1,1) == " ")) then
+            return nil
+        end
+    end
 
     hash = libC:fcs32update(hash, libS:Serialize(data))
     hash = libC:fcs32final(hash)
@@ -57,7 +64,8 @@ end
 
 function QuestieHash:AddNewQuestHash(questId)
     Questie:Debug(Questie.DEBUG_DEVELOP, "[QuestieHash] AddNewQuestHash:", questId)
-    questLogHashes[questId] = QuestieHash:GetQuestHash(questId)
+    local oldHash = questLogHashes[questId]
+    questLogHashes[questId] = QuestieHash:GetQuestHash(questId) or oldHash
     QuestieQuest:SetObjectivesDirty(questId)
 end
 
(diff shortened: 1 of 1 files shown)


Modules/Libs/QuestieHash.lua
```diff
@@ -35,7 +35,8 @@ function QuestieHash:InitQuestLogHashes()
                 Questie._sessionWarnings[questId] = true
             end
         elseif (not isHeader) then
-            questLogHashes[questId] = QuestieHash:GetQuestHash(questId)
+            local oldHash = questLogHashes[questId]
+            questLogHashes[questId] = QuestieHash:GetQuestHash(questId) or oldHash
```
Laumesis, line 39: oldHash is **always** nil --> questLog looks like there is no quest at all if there is cache miss

Modules/Libs/QuestieHash.lua
```diff
@@ -57,7 +64,8 @@ end
 
 function QuestieHash:AddNewQuestHash(questId)
     Questie:Debug(Questie.DEBUG_DEVELOP, "[QuestieHash] AddNewQuestHash:", questId)
-    questLogHashes[questId] = QuestieHash:GetQuestHash(questId)
+    local oldHash = questLogHashes[questId]
+    questLogHashes[questId] = QuestieHash:GetQuestHash(questId) or oldHash
```
Laumesis, line 68: oldHash is most likely nil --> if there is cachemiss, then the whole quest is missing from questLogHashes again

Modules/Libs/QuestieHash.lua
```diff
@@ -83,7 +91,7 @@ function QuestieHash:CompareHashesOfQuestIdList(questIdList)
         else
             local oldHash = questLogHashes[questId]
             if oldHash then
-                local newHash = QuestieHash:GetQuestHash(questId)
+                local newHash = QuestieHash:GetQuestHash(questId) or oldHash
```
Laumesis, line 94: This leads to wrong return values. Return may be wrong to both ways: Telling there is change when not, and no changes when there was a change.

Modules/Libs/QuestieHash.lua
```diff
@@ -117,7 +125,7 @@ function QuestieHash:CompareQuestHash(questId)
     else
         local oldHash = questLogHashes[questId]
         if oldHash then
-            local newHash = QuestieHash:GetQuestHash(questId)
+            local newHash = QuestieHash:GetQuestHash(questId) or oldHash
```
Laumesis, line 128: This leads to wrong return values. Return may be wrong to both ways: Telling there is change when not, and no changes when there was a change.

Muehe:
Obsolete

PR #3455, opened 2022-05-07
Fix Gubber Blump can train fishing

gouletr:
Fixes [Gubber Blump (10216)](https://classic.wowhead.com/npc=10216/gubber-blump) can train fishing. He's not a conventional trainer, but one of its gossip is to train fishing when you don't have the profession. He then becomes a vendor after you trained fishing from him.

I am not sure if this is the proper way to fix this, please let me know if you want a different fix.

### DIFF
--- a/Modules/QuestieMenu.lua
+++ b/Modules/QuestieMenu.lua
@@ -421,6 +421,9 @@ function QuestieMenu:PopulateTownsfolk()
         end
     end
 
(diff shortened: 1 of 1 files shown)


BreakBB:
Hey @gouletr yeah the profession trainers are not as easy to fix.

And your fix seems sufficent enough. Your NPC is an edge case as it does not have a `<Fishing Trainer>` subname.

PR #3336, opened 2022-01-30
Improve login, database and questlog performance

Laumesis:
Reduces lag spikes at login and at accepting, turning in and progressing a quest.
Improves overall performance everywhere database is used.

~~Backports a fix from quest-event-refactor to SmoothReset(). Really an edge case, I doubt it has been visible to player often.~~

~~This PR includes PR #3337 and PR #3338~~

PR #3473, opened 2022-05-11
Performance Improvements

Laumesis:
Enable skippers on varied size data in DB queries

I need to verify still. Now I just copied my old changes from old PR #3336. Other changes are alredy merged to master.


BreakBB:
We are working on replacing the compiler entirely, so no effort will be spent here.

PR #3477, opened 2022-05-12
Fix item Stormwind Seasoning Herbs (2665)

gouletr:
Fixes item [Stormwind Seasoning Herbs (2665)](https://classic.wowhead.com/item=2665/stormwind-seasoning-herbs)
- npcDrops [Felicia Gump (1303)](https://classic.wowhead.com/npc=1303/felicia-gump)
- relatedQuests [Seasoned Wolf Kabobs (90)](https://classic.wowhead.com/quest=90/seasoned-wolf-kabobs)

### DIFF
--- a/Database/Corrections/Classic/classicItemFixes.lua
+++ b/Database/Corrections/Classic/classicItemFixes.lua
@@ -76,6 +76,9 @@ function QuestieItemFixes:Load()
         [2633] = {
             [itemKeys.npcDrops] = {940,941,942}, -- #2433
         },
+        [2665] = {
+            [itemKeys.relatedQuests] = {90},
+        },
         [2686] = {
             [itemKeys.relatedQuests] = {308},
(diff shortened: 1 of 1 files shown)


BreakBB:
Hey @gouletr with one of the last updates we changed it so that the `vendors` field is also used for quest objectives. So it is no longer required for the work around to add them as `npcDrops`. I might misunderstand you though and the `relatedQuests` is indeed missing!

gouletr:
Ok, I guess the `npcDrops` should be empty then. Do I just set the field to `{}` to make it empty when the correction is applied?

BreakBB:
Since the field has no value in the original `classicItemDB.lua` you can simply remove that line from the corrections. It should stay `nil`

gouletr:
done

BreakBB (approved):

PR #3481, opened 2022-05-14
Use string.byte(text,1) instead of string.sub(text,1,1) for objectives

Laumesis:


Laumesis:
Rebased on newer master

Modules/Quest/QuestEventHandler.lua
```diff
@@ -100,8 +100,8 @@ function _QuestEventHandler:HandleQuestAccepted(questId)
     -- We first check the quest objectives and retry in the next QLU event if they are not correct yet
     local questObjectives = C_QuestLog.GetQuestObjectives(questId)
     for _, objective in pairs(questObjectives) do
-        -- When the objective text is not cached yet it looks similar to " slain 0/1"
-        if (not objective.text) or stringSub(objective.text, 1, 1) == " " then
+        -- When the objective text is not cached yet it looks similar to " slain: 0/1"
+        if (not objective.text) or (stringByte(objective.text, 1) == 32) then
```
BreakBB, line 104: Could you move the `32` into a constant with some `WHITE_SPACE_BYTE` naming?

Laumesis: Rather not because of speed. I could add a comment to those where it is used.
After whole `QuestLogCache` is finished there should be 2 lines left where the check is done - one in `QuestLogCache.lua` and one in `QuestieValidateGameCache.lua`

BreakBB (approved):

Issue #2514, opened 2021-01-25
Show Related Quests in Mob Tooltips


Gogo1951:

Gogo1951:
Closing as stale.

Issue #2354, opened 2020-09-13
Mage quest "Items of Power"


Axx0r:
## Description
During this quest you need to gather 10 "Witherbark totem sitck" and then charge them to get a "bolt charged bramble". After getting the "bolt charged bramble" the spawn points for the sticks are still there.

## Wowhead link
https://classic.wowhead.com/quest=1948/items-of-power

## Questie version
6.0.5

Gogo1951:
Hey @Faeryqueenaeval ,

Check out how these are done to make the fix here.

https://classic.wowhead.com/quest=6022/to-kill-with-purpose

https://classic.wowhead.com/quest=1136/frostmaw

Both of those have fake objectives that you need to gather / combine / use to get to the real objective.

Nerillida:
Still happening with Totem sticks.

anon1231823:
@Cabro is this a case where the extraobjectives can help? I feel like you got something that could fix this

Cabro:
I feel like this isn't even a bug, this is exactly what Blizzard is doing.

PR #3548, opened 2022-06-03
[fix] Fix tooltips not registering for completed quests on mobs and items

BreakBB:
## Issue references

After a reload/login no tooltips on items/mobs are added for complete quests (still in quest log but ready to turn in).
Not sure if there is an open issue for this.

## Proposed changes

- Complete quests will now also register tooltips on mobs and items
- This is the same behaviour for when you complete a quest and don't do a reload

## Screenshots

Taken after a /reload

grafik

### DIFF
--- a/Modules/Quest/QuestieQuest.lua
+++ b/Modules/Quest/QuestieQuest.lua
@@ -656,7 +656,7 @@ function QuestieQuest:PopulateObjective(quest, objectiveIndex, objective, blockI
 
     local objectiveData = quest.ObjectiveData[objective.Index] or objective -- the reason for "or objective" is to handle "SpecialObjectives" aka non-listed objectives (demonic runestones for closing the portal)
 
-    if (not completed) and (not next(objective.spawnList)) and _QuestieQuest.objectiveSpawnListCallTable[objectiveData.Type] then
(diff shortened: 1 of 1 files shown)


Muehe:
Not quite sure I understand correctly. Right now the line gets removed once ALL objectives of a quest are finished but not when ONE objective is finished? And the aim is to not remove them at all?

Did some testing and it seems that the line stays at least for provided items (requiredSourceItems), but didn't have a finished loot quest at hand to test.

Also what is the purpose? If it's preventing players from selling the items then it would make more sense to show something regardless of the player being on the quest or not, wouldn't it? Like "This item is an objective for quest X (ID)".

Muehe:
After some more testing it seems the tooltip for everything seems to be removed on reload. Not sure why it stayed before, might have missed something, have to re-check on that char.

This PR has some relation to #2354 and maybe #2514

BreakBB:
So what is your state on this @Muehe ?

The current `master` state does not remove any tooltip on objective completion. But after a `/reload` the tooltips are not registered again for completed quests. So if you don't do a `/reload` or relog into the game everything works as expected, but not if you do so.

From by point of view (hence this PR) the tooltips should never be removed and always show for items and mobs. Only on turn though ofc.

PR #3562, opened 2022-06-14
Fix npc Private Thorsen (738)

gouletr:
Fix npc [Private Thorsen (738)](https://classic.wowhead.com/npc=738/private-thorsen) spawns.

[Some time ago](https://github.com/Questie/Questie/pull/2095), a correction was made to move the spawn location of that npc to the area where he gets ambushed. But this is incorrect for two reasons:
1. The "ambush" location proposed in the correction is invalid. The correct location is `40.34,8.44` precisely.
2. Most of the time, Private Thorsen spends his time at the original spawn location that was already in the database, at `37.98,3.41`. Making his spawn location to only be the "ambush" location is wrong 99% of the time.

So here I propose we set his spawns to be both at the camp and the ambush location, so that at least if the npc can't be found in one area, it is found at the other.


drejjmit:
I remember setting pickup location to where you're being offered the quest after helping him.
He's not offering the quest where he's at 99% of the time so showing a pickup spot there would be confusing?

gouletr:
> I remember setting pickup location to where you're being offered the quest after helping him. He's not offering the quest where he's at 99% of the time so showing a pickup spot there would be confusing?

But you also need to turn-in [Supplies to Private Thorsen (198)](https://classic.wowhead.com/quest=198/supplies-to-private-thorsen) to him, and that can only be done at the camp where he spends most of his time.

So this is why I think we need to show both locations.

PR #3564, opened 2022-06-17
Fix quest Beginnings (1599)

gouletr:
Fix quest [Beginnings (1599)](https://classic.wowhead.com/quest=1599/beginnings) missing on the map. It was mistaken for quest 1559.


BreakBB:
Wow good catch! :O

PR #3575, opened 2022-06-26
Fix item Frost Oil (3829)

gouletr:
Fix item [Frost Oil (3829)](https://classic.wowhead.com/item=3829/frost-oil#contained-in-item) missing npc drop [Cenarion Hold Infantry (15184)](https://classic.wowhead.com/npc=15184/cenarion-hold-infantry) and item drop [Cenarion Circle Cache (11887)](https://classic.wowhead.com/item=11887/cenarion-circle-cache).


BreakBB:
Hey @gouletr I am quite sure that this NPC does not drop Frost Oil (wowhead is wrong there). And even if that wouldn't be useful for Questie to show, because the quest is in Badlands for many levels below 😅
Would be fine for me to add the Cache relation though

gouletr:
> Hey @gouletr I am quite sure that this NPC does not drop Frost Oil (wowhead is wrong there). And even if that wouldn't be useful for Questie to show, because the quest is in Badlands for many levels below 😅 Would be fine for me to add the Cache relation though

Ok let's drop the npc.

PR #3577, opened 2022-06-26
Fix item Gyrochronatom (4389)

gouletr:
Fix item [Gyrochronatom (4389)](https://classic.wowhead.com/item=4389/gyrochronatom) missing just about all information about it.

### DIFF
--- a/Database/Corrections/Classic/classicItemFixes.lua
+++ b/Database/Corrections/Classic/classicItemFixes.lua
@@ -220,6 +220,12 @@ function QuestieItemFixes:Load()
             [itemKeys.vendors] = {1448,6777,2685,11185,14637,5519,5175,3413,3133,3495,8679,1694,2687,4587,8678,6730,9544,2682,2683,2684,2688,9676},
             [itemKeys.relatedQuests] = {174,2609},
         },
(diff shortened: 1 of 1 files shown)


BreakBB:
I wouldn't want to add the NPCs for this as well. Otherwise users on the related quest will be confused and might head to Stranglethorn for the NPCs.

gouletr:
Those NPCs do drop that item thought, isn't better if the database is 100% accurate? However I can understand your reluctance, since the drop rate is so low anyway. I'll just remove the NPCs for now.

drejjmit:
I thought the vendors were limited supply 🤔

gouletr:
> I thought the vendors were limited supply 🤔

Sure, but why not show them on the map nonetheless? It's the same for `Bronze Tube` and I've used multiple vendors exactly for that purpose.

What is Questie trying to achieve here? Isn't accuracy important just as much?

BreakBB:
I am with your on the vendors @gouletr 👍🏻  But yeah if the drop chance on NPCs is too low we don't show them as that won't be very helpful in most cases. Limited supply vendors are a good chance imo.

PR #3578, opened 2022-06-26
Fix item Battered Junkbox (16882) causing stack overflow

gouletr:
Fix item [Battered Junkbox (16882)](https://classic.wowhead.com/item=16882/battered-junkbox) causing stack overflow, because it contained itself.


BreakBB:
Hahaha, great one! 😂 

Thanks!

PR #3609, opened 2022-07-14
Fix spawn position and add waypoints for Cursed Sycamore (5881)

gouletr:
Fix [Cursed Sycamore (5881)](https://classic.wowhead.com/npc=5881/cursed-sycamore) spawn location and add waypoints.

@BreakBB is there a quick hack to display waypoints for testing, other than having the quest active? Thanks


BreakBB:
I am not sure @gouletr if the "Advanced Search" function shows the waypoints as well, if you add an NPC to the map 🤔

gouletr:
Ah cool, didn't know that feature. But it only shows a small pin. No waypoints. Anyway, these changes, I verified them on live server, I actually followed the npc around to get accurate positions.

BreakBB:
I fear not (at least not a simple UI option). It might be doable with combined function calls though 🙈

PR #3785, opened 2022-07-22
Test datatype overflows at DB compile while debug on.

Laumesis:
Add "raw_assert" mode to test overflows while writing into a raw stream.
This mode is active for DB compile while Questie debug mode is on.
Also:
* Kind of require defining decoding mode to QuestieStream.
* Check mode is correct when using recycled stream.

### DIFF
--- a/Database/compiler.lua
+++ b/Database/compiler.lua
@@ -827,7 +827,7 @@ function QuestieDBCompiler:CompileTableCoroutine(tbl, types, order, lookup, data
     QuestieDBCompiler.index = 0
 
     QuestieDBCompiler.pointerMap = {}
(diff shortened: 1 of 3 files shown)


Laumesis:
Rebased. Git was annoying at rebasing thanks to same changes in master and commit, but some indents not fixed.

Laumesis:
I think this might be okey to merge. I'll fix cli.lua output in another PR.
@BreakBB Please review and **test**.
Did I remember to handle raw vs raw_assert mode selections correctly?
Is anywhere used `QuestieStreamLib:GetStream(mode)`  so that mode is not defined?
I hope I didn't break anything in QuestieComms.

BreakBB (approved):

PR #3799, opened 2022-07-24
Internal QuestXP

Logonz:
Using internal QuestXP logic and databases rather than an outside library.


Laumesis:
Is there licensing issues in this kind of use?

BreakBB (changes requested):
I think the database files could be renamed to be simply `classic.lua`, `tbc.lua` and `wotlk.lua`.

Other than that it looks good 👍🏻

PR #3800, opened 2022-07-24
Add extra objectives to quest Defenders of Darrowshire (5211)

gouletr:
Added extra objectives to quest [Defenders of Darrowshire (5211)](https://classic.wowhead.com/quest=5211/defenders-of-darrowshire) so that the ghouls that need to be killed in order to free Darrowshire spirits do appear on the map.


drejjmit:
Do you mind adding the l10n post to
https://github.com/Questie/Questie/blob/master/Localization/Translations/ExtraObjectives/ClassicObjectives.lua

gouletr:
> Do you mind adding the l10n post to https://github.com/Questie/Questie/blob/master/Localization/Translations/ExtraObjectives/ClassicObjectives.lua

done

PR #3854, opened 2022-07-31
Remove raceKeys.ALL, use raceKeys.NONE

Laumesis:
Please change fix DB import scripts and do import again.

The PR should work as is, but DB would be more efficient with fixed DB import scripts using NONE.
Search results are valid, but messy until DB import is fixed.
PS. Still missing Vanilla reputations from DB


BreakBB (approved):
Looks fine to me 👍🏻 Not much that could go wrong with this change.

Laumesis:
@BreakBB Who knows DB import stuff and does the scripts? Could you poke them to do the change? Those should be changed to use "NONE" instead of old "ALL".

BreakBB:
Already poked Muehe about it 👍🏻

PR #3855, opened 2022-07-31
[fix] DB Query return value handling

Laumesis:
Can't use wow's unpack() as it stops at first nil value.
This is a bit band aid fix, but as bonus those querying only two fields
are faster when using QuerySingle twice.

**TEST BEFORE MERGE**

### DIFF
--- a/Modules/Journey/QuestieSearchResults.lua
+++ b/Modules/Journey/QuestieSearchResults.lua
@@ -90,7 +90,8 @@ local function CreateShowHideButton(id)
 end
 
 function QuestieSearchResults:QuestDetailsFrame(details, id)
-    local name, requiredLevel, requiredRaces, objectivesText, startedBy, finishedBy = unpack(QuestieDB.QueryQuest(id, "name", "requiredLevel", "requiredRaces", "objectivesText", "startedBy", "finishedBy"))
+    local ret = QuestieDB.QueryQuest(id, "name", "requiredLevel", "requiredRaces", "objectivesText", "startedBy", "finishedBy") or {}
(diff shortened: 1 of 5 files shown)


Laumesis:
I reviewed code, but not test it so much that I could say it is tested.

Modules/Journey/QuestieSearchResults.lua
```diff
@@ -90,7 +90,8 @@ local function CreateShowHideButton(id)
 end
 
 function QuestieSearchResults:QuestDetailsFrame(details, id)
-    local name, requiredLevel, requiredRaces, objectivesText, startedBy, finishedBy = unpack(QuestieDB.QueryQuest(id, "name", "requiredLevel", "requiredRaces", "objectivesText", "startedBy", "finishedBy"))
+    local ret = QuestieDB.QueryQuest(id, "name", "requiredLevel", "requiredRaces", "objectivesText", "startedBy", "finishedBy") or {}
+    local name, requiredLevel, requiredRaces, objectivesText, startedBy, finishedBy = ret[1], ret[2], ret[3], ret[4], ret[5], ret[6]
```
BreakBB, line 94: Is this pattern faster than the multiple single queries? If yes why not using it at the other places as well?

Laumesis: For querying two fields QueryQuestSingle() is faster. I haven't tested 3+ fields.

Generally this PR intent is not speed up, but getting rid of unpack, which leads to bugs if there is `nil` within return values.

BreakBB (approved):

PR #3856, opened 2022-07-31
implement own tunpack into QuestieValidateGameCache

Laumesis:
Current use of the "module" works fine, but future use cases may break without fix.


Laumesis:
Tested:
```
/run function lautunpack(tbl) if tbl.n == 0 then return nil end local function recursion(i) if i == tbl.n then return tbl[i] end return tbl[i], recursion(i+1) end return recursion(1) end
/dump lautunpack({nil,2,3,nil,nil,6, nil, n=7})
```

BreakBB (approved):

PR #3864, opened 2022-08-02
[fix] Improve performance. Refactoring == nil

Laumesis:
Convert `x == nil` comparisons to `not x`, if `x` doesn't have different meanings for `nil` and `false`.

`(not x)` is more readable than `x == nil` most of time. As a bonus it is slightly faster (around 5%).

NOT REVIEWED. NOT TESTED.


Laumesis:
Rebased on master instead of another PR branch.

BreakBB (approved):
Looks good 👍🏻 Feeling better now that all the `== nil`s are gone? 😬

PR #3886, opened 2022-08-02
[fix] Improve DB performance

Laumesis:


Database/QuestieDB.lua
```diff
@@ -832,39 +811,26 @@ function QuestieDB:GetQuest(questId) -- /dump QuestieDB:GetQuest(867)
...
-            end
-        end
-    end
-
-    local zoneOrSort = rawdata[questKeys.zoneOrSort]
-    if zoneOrSort and zoneOrSort ~= 0 then
-        if zoneOrSort > 0 then
-            QO.Zone = zoneOrSort
-        else
-            QO.Sort = -zoneOrSort
```
BreakBB, line 853: I guess QO.Sort was never used so you removed it?

Database/QuestieDB.lua
```diff
@@ -935,34 +902,37 @@ function QuestieDB:GetCreatureLevels(quest)
     return creatureLevels
 end
 
-local playerFaction = UnitFactionGroup("player")
+local playerFaction = UnitFactionGroup("player")  -- TODO: FIXME: Api call won't return correct value yet. Only after player entering world event
```
BreakBB, line 905: Don't forget this TODO 😁 Can't you simply use QuestiePlayer for this?

BreakBB (commented):
I know this is still a draft, but I find these changes quite readable and couldn't find much to argue 😁 👍🏻

Database/QuestieDB.lua
Laumesis: Both `.Zone` and `.Sort` are never used

Database/QuestieDB.lua
Laumesis: It is a bug in old old old master, I just mentioned it to be a bug 😛

Laumesis:
```
git rebase master
git push --force-with-lease
```

PR #3931, opened 2022-08-07
Fix quest The Infernal Orb (1954)

gouletr:
Fix quest [The Infernal Orb (1954)](https://classic.wowhead.com/quest=1954/the-infernal-orb), the quest [Return to the Marsh (1953)](https://classic.wowhead.com/quest=1953/return-to-the-marsh) is not required to pick up this quest, its only a breadcrumb.

### DIFF
--- a/Database/Corrections/Classic/classicQuestFixes.lua
+++ b/Database/Corrections/Classic/classicQuestFixes.lua
@@ -947,6 +947,9 @@ function QuestieQuestFixes:Load()
         [1950] = {
             [questKeys.triggerEnd] = {"Secret phrase found", {[zoneIDs.THOUSAND_NEEDLES]={{79.56,75.65}}}},
         },
+        [1954] = {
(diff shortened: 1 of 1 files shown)


BreakBB:
Can you still do "Return to the Marsh" when accepting "The Infernal Orb"? If no then the quests should be marked exclusive to each other.

gouletr:
Correct, you cannot do "Return to the Marsh" if you pick up "The Infernal Orb", however looking at the map "Return to the Marsh" is no longer visible after I picked up "The Infernal Orb" with these changes, so I am unsure what making them exclusive would do. They are already chained together with the `nextQuestInChain` field.

BreakBB (approved):
Yeah they are already connected through the "nextQuestInChain" which markes them as exclusive to each other. So your changes are everything needed to mark the quest as breadcrumb quest 👍🏻

PR #3937, opened 2022-08-08
Better TomTom integration

anzz1:
## Proposed changes

- in worldmap ctrl + left-click can now remove old waypoint when clicked second time

### DIFF
--- a/Modules/FramePool/QuestieFrame.lua
+++ b/Modules/FramePool/QuestieFrame.lua
@@ -174,10 +174,17 @@ function _Qframe:OnClick(button)
     end
     if self and self.UiMapID and IsControlKeyDown() and TomTom and TomTom.AddWaypoint then
         -- tomtom integration (needs more work, will come with tracker
+        local m = self.UiMapID
+        local x = self.x/100
+        local y = self.y/100
+        local title = self.data.Name
+        local add = true
         if Questie.db.char._tom_waypoint and TomTom.RemoveWaypoint then -- remove old waypoint
-            TomTom:RemoveWaypoint(Questie.db.char._tom_waypoint)
+            local waypoint = Questie.db.char._tom_waypoint
+            TomTom:RemoveWaypoint(waypoint)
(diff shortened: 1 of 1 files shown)


Modules/FramePool/QuestieFrame.lua
```diff
@@ -174,10 +174,17 @@ function _Qframe:OnClick(button)
...
+        local m = self.UiMapID
+        local x = self.x/100
+        local y = self.y/100
+        local title = self.data.Name
+        local add = true
         if Questie.db.char._tom_waypoint and TomTom.RemoveWaypoint then -- remove old waypoint
-            TomTom:RemoveWaypoint(Questie.db.char._tom_waypoint)
+            local waypoint = Questie.db.char._tom_waypoint
+            TomTom:RemoveWaypoint(waypoint)
+            add = (waypoint[1] ~= m or waypoint[2] ~= x or waypoint[3] ~= y or waypoint.title ~= title or waypoint.from ~= "Questie")
```
BreakBB, line 185: The last check is not required, is it?

` or waypoint.from ~= "Questie"` can this even be true? 🤔

MyTechnoHunter: not having looked at the other parts of the code.. if you have tomtom installed i would presume IT can add a waypoint that wouldnt be marked a"Questie"

anzz1: on the line directly below that the waypoint is added with a from="Questie" field like addons are supposed to (check api for TomTom:AddWaypoint). 
yes the check isn't strictly required but that will make sure it'll only remove waypoints which were added by questie, like it should.

anzz1: test in-game it works. it is much more intuitive that way, that clicking it again removes it. that's something you intuitively try when wanting to remove a waypoint only to discover that it doesn't work and is "stuck".

BreakBB (approved):

PR #3940, opened 2022-08-11
DB update

Muehe:
General DB update

Major changes:

* Added reputation rewards
* Added WotLK waypoints

For more details see commit message.


Muehe:
Ok, so Classic works totally fine (no killCredit objectives), but TBC/WotLK don't compile, so here's what's TODO:

- [ ] Update `["objectives"]` code in writers table of compiler.lua
- [ ] Update `["objectives"]` code in readers table of compiler.lua
- [ ] Update `GetQuest` function code in QuestieDB.lua

Maybe more...

Muehe:
Ok, so with #3947 merged and the additional commit to QuestieDB.lua the multiple killCredit change seems to now be working as intended.

TODO:

- [ ] Fix killCreditObjectives where the base ID has spawns of it's own (included base ID in creature list)
- [ ] Check DB for regressions (e.g. Ghost Howl waypoints seem to be missing).

Database/QuestieDB.lua
```diff
@@ -790,14 +790,16 @@ function QuestieDB:GetQuest(questId) -- /dump QuestieDB:GetQuest(867)
...
         end
-        if objectives[5] then
-            local killCreditObjective = {
-                Type = "killcredit",
-                IdList = objectives[5][1],
-                RootId = objectives[5][2],
-                Text = objectives[5][3]
-            }
-            tinsert(QO.ObjectiveData, killCreditObjective);
+        if type(objectives[5]) == "table" and #objectives[5] > 0 then
```
BreakBB, line 793: We can assume this is always `nil` or a table, Can't we?

```suggestion
        if objectives[5] and #objectives[5] > 0 then
```

Database/compiler.lua
```diff
@@ -540,13 +545,22 @@ QuestieDBCompiler.writers = {
             QuestieDBCompiler.writers["objective"](stream, value[2])
             QuestieDBCompiler.writers["objective"](stream, value[3])
             QuestieDBCompiler.writers["u24pair"](stream, value[4])
-            if value[5] and value[5][1] and value[5][1][1] then
-                stream:WriteByte(#value[5][1])
-                for _, v in pairs(value[5][1]) do
-                    stream:WriteInt24(v)
+
+            local killobjectives = value[5]
+            if type(killobjectives) == "table" and #killobjectives > 0 then
```
BreakBB, line 550: Same as before, right?

```suggestion
            if killobjective and #killobjectives > 0 then
```

BreakBB (commented):
Just a quick look over. Didn't check the DB diff

Database/QuestieDB.lua
Muehe: To be honest this was kinda cargo-culted over from the compiler and is probably not needed *here*.

Database/compiler.lua
Muehe: Not so sure about this one, when I tried to code this on my simpler checks didn't seem to work here, but maybe I did something wrong. @Laumesis might be able to say more on this?

Database/QuestieDB.lua
BreakBB: I totally agree, this (and also the other point) is nothing worth investing much time at this point. Especially if the code is working as it is.

PR #3943, opened 2022-08-11
Added german translation to daylie questes(Classic).

hugelol:
## Proposed changes

Added german translation to daylie quests.


BreakBB:
Hey @hugelol thanks for starting with these translations! ❤️ It is a lot of work and we appreciate every help we get 👍🏻

BreakBB (approved):

PR #3953, opened 2022-08-18
Correct readme.md

hugelol:
Modified the readme.md since i got trouble for my first contribution finding a "nil" value in the translation-files, ther were only "false" when no translation exists.

if this is wrong, could someone explain if i am missing a point here. thanks :)


BreakBB (approved):
Great catch. Thanks for this!

PR #3957, opened 2022-08-23
Adding zones

hugelol:
added a zone i found which was not in the translation and also corrected some names to the standard german in the gameclient.


hugelol:
forgott the other continent, lol

PR #3959, opened 2022-08-23
Added all the subzones and their parents

Logonz:
[3917] = 3688, -- diff both these are "valid" one is a subzone (3688) and one is a zone (3917)
    [3696] = 3522, -- Wrong, points to blades edge but it should be terokkar
    [3845] = 3842, -- wrong? This is not even on the outlands map but is the instance which is pointed to the subzone tempest keep

These three we need to take a look at otherwise i think i've got them dumped correctly and matching the ones we had previously 🙂


BreakBB:
`3917` = Auchindoun and so is `3688` which seems fine to me. Doesn't matter either, as no object or NPC has one of these IDs listed in our DB.

`3696` = [Barrier Hills](https://wowpedia.fandom.com/wiki/Barrier_Hills) which is correct

> The Barrier Hills are located in the northeastern corner of [Nagrand](https://wowpedia.fandom.com/wiki/Nagrand) and northwestern [Terokkar Forest](https://wowpedia.fandom.com/wiki/Terokkar_Forest), forming the border to [Zangarmarsh](https://wowpedia.fandom.com/wiki/Zangarmarsh) in the north and bordering [Shattrath City](https://wowpedia.fandom.com/wiki/Shattrath_City) to the southeast.


`[3845] = 3842` doesn't matter either as no spawn is in our database 🤷🏻

BreakBB (approved):

Logonz:
3696 = [Barrier Hills](https://wowpedia.fandom.com/wiki/Barrier_Hills) which is correct
But it points to 3522 which is blades edge, my export points to terrokar forest which should be correct?

BreakBB:
Yes @Logonz :grin:

Logonz:
Then i think this branch is good enough to merge :)
It should at least be an improvement from before, even if it is not perfect!

BreakBB:
Feel free to do so :grin:

PR #3960, opened 2022-08-23
Update WotLK lookups

EKE00372:
* Rename folder to Wotlk
* Update **items**, **units** and **quest** data, full language based on beta 3.4.0.45189
*extractor buy @qqytqqyt*


EKE00372:
should switch to WotLK brench

PR #3961, opened 2022-08-23
Add Wotlk lookups (full language)

EKE00372:
Add **items**, **units** and **quest** data, full language based on beta 3.4.0.45189
*extractor buy @qqytqqyt*

#3960


qqytqqyt:
Thank you. To add that as suggested we put the data onto a separate folder to not change the existing modules, but the data has been tested against the TBC folder and we did not encounter any errors.
all
The items and npcs in the lookups have been filtered based on the id list in _Questie\Database\Wotlk_ folder. The quest list is unfiltered.
Due to the limitation of the in-game scan (that is dependent on server status), there might be some elements missing but should be very few.

PR #4003, opened 2022-09-01
Fixed the tooltip to allow us to make the lines with a prefix and size

Logonz:
unknown

First one is the new one second is the old one for comparison


Modules/Libs/QuestieLib.lua
```diff
@@ -535,3 +535,114 @@ function QuestieLib.tunpack(tbl)
...
+local font, size = frameObject:GetFont()
+--Chinese? "Fonts\\ARKai_T.ttf"
+objectiveFontString:SetFont(font, size);
+objectiveFontString:Hide()
+
+---Emulates the wrapping of a quest description
+---@param line string @The line to wrap
+---@param prefix string @The prefix to add to the line
+---@param combineTrailing boolean @If the last line is only one word, combine it with previous? TRUE=COMBINE, FALSE=NOT COMBINE, default: true
+---@param splitOnDot boolean @Should we add a linebreak if a dot appears thats not at the end of a line TRUE=NEW ROW, FALSE=NO NEW ROW, default: true
```
BreakBB, line 564: Do we need both parameters if they are true in both usages?

Modules/Libs/QuestieLib.lua
```diff
@@ -535,3 +535,114 @@ function QuestieLib.tunpack(tbl)
...
+---@param desiredWidth number @Set the desired width to wrap, default: 275
+---@return table[] @A table of wrapped lines
+function QuestieLib:TextWrap(line, prefix, combineTrailing, splitOnDot, desiredWidth, questid)
+    if(objectiveFontString:IsVisible()) then print("TextWrap already running... strange, look into pooling the FontString ugh") end
```
BreakBB, line 568: Shouldn't this be a real debug print or something?

Modules/Libs/QuestieLib.lua
```diff
@@ -535,3 +535,114 @@ function QuestieLib.tunpack(tbl)
...
+    --We have to show it or else the functions won't work... But we set the opacity to 0 on creation
+    objectiveFontString:SetWidth(desiredWidth or frameObject:GetWidth() or 275) --QuestLogObjectivesText default width = 275
+    objectiveFontString:Show()
+
+    --Make a linebreak on each "dot" character if there is a space after (don't want it on end of line)
+    local useLine = string.gsub(line, "%. ", "%.%\n")
+
+    objectiveFontString:SetText(useLine)
+    --Is the line wrapped?
+    if(objectiveFontString:GetUnboundedStringWidth() > objectiveFontString:GetWrappedWidth()) then
```
BreakBB, line 583:
```suggestion
    local isLineWrapped = objectiveFontString:GetUnboundedStringWidth() > objectiveFontString:GetWrappedWidth()
    if(isLineWrapped) then
```

Modules/Libs/QuestieLib.lua
```diff
@@ -535,3 +535,114 @@ function QuestieLib.tunpack(tbl)
...
+        local lines = {}
+        local startIndex = 1
+        local endIndex = 2 --We should be able to start at a later index...
+        --This function returns a list of size information per row, so we use this to calculate number of rows
+        local numberOfRows = #objectiveFontString:CalculateScreenAreaFromCharacterSpan(startIndex, strlen(useLine))
+        --print("lineLen", strlen(line))
```
BreakBB, line 589:
```suggestion
```

Modules/Libs/QuestieLib.lua
```diff
@@ -535,3 +535,114 @@ function QuestieLib.tunpack(tbl)
...
+                end
+            end
+            --Change the startIndex to be the new line, and add the line to the lines list
+            startIndex = endIndex - 2
+            endIndex = endIndex
+
+            table.insert(lines, prefix..newLine)
+        end
+        --Hide the fontstring
+        objectiveFontString:Hide()
```
BreakBB, line 639:
```suggestion
        objectiveFontString:Hide()
```

This is quite good to read, no need for a comment

Modules/Libs/QuestieLib.lua
```diff
@@ -535,3 +535,114 @@ function QuestieLib.tunpack(tbl)
...
+        return lines
+    else
+        --Line was not wrapped, return the string as is.
+        --Hide the fontstring
```
BreakBB, line 643:
```suggestion
```

BreakBB (changes requested):
Great and small change. A few minor things came to my eye, nothing important though 😅

Modules/Libs/QuestieLib.lua
Logonz: Why do it this way? Isn't making a local variable with a descriptive name just a comment with more steps? :D Also worse performance because it saves it in memory!

Modules/Libs/QuestieLib.lua
Logonz: Well, i made it general so that we can control these things in other places if we want. While they default to true maybe one would want to not have these behaviors?

Modules/Libs/QuestieLib.lua
Logonz: There are two ways to think here, this should NEVER happen. But if it does it is actually really bad. So if it happens I kind of want to know about it.
But as long as no one changes the function to add a coroutine yield in the middle of it the code is basically "dead".
If we do a regular debug, we might as well remove the entire line for performance :)

Modules/Libs/QuestieLib.lua
BreakBB: I come from a world where comments usually means a descriptive variable/method name is preffered 🤷🏻 But I am totally fine to keep it the way it is and I understand the performance aspect 😁 👍🏻

Modules/Libs/QuestieLib.lua
BreakBB: Sounds fine to me 👍🏻

Modules/Libs/QuestieLib.lua
BreakBB: Why would `Questie:Error("TextWrap already running... Please report this on GitHub or Discord.")` be worse?

Logonz: Ehh didn't think of error :D Yeah then it should be changed!
I always use print when developing for the speed but your idea is better!

BreakBB: > I always use print when developing for the speed but your idea is better!

So do I! :D

PR #4047, opened 2022-09-05
Performance fixes in IsDoable

Logonz:
Performance improvements in IsDoable
A tip is to add a quick escape with an extra cache-list here


Modules/QuestieReputation.lua
```diff
@@ -1,10 +1,13 @@
 ---@class QuestieReputation
+------@field hasFactionChange boolean @true if some faction leveled up since the last QuestieQuest:CalculateAndDrawAvailableQuestsIterative
```
BreakBB, line 2:
```suggestion
---@field hasFactionChange boolean @true if some faction leveled up since the last QuestieQuest:CalculateAndDrawAvailableQuestsIterative
```

Issue #4080, opened 2022-09-09
Stop showing "In Search of Thaelrid" for Horde


Synopt:
## Description
In Wrath, Questie still displays the quest "In Search of Thaelrid" in Darnassus on the map for Horde, however in patch 3.2 it was fixed to be available for Alliance only. 

I've verified this by going to Darnassus on a character who hasn't completed the quest, it wasn't available.

## Wowhead link
https://www.wowhead.com/wotlk/quest=1198/in-search-of-thaelrid

## Questie & WoW version
Version 7.0.2
Wrath 3.4.0 

Please also state if your quest issue is about Classic Era/SoM or TBC.
It's Wrath Classic

Synopt:
I believe this is the line of corrections that enables this quest to show for horde in the first place:

https://github.com/Questie/Questie/blob/master/Database/Corrections/Classic/classicQuestFixes.lua#L625
Since it's linked with this commit about adding it for horde:
https://github.com/Questie/Questie/commit/ecc498ee176e2abef3c4c7b79b7be8ccafc72ba3

I don't know if you're maintaining separate versions for each game type but I'll make a PR to remove it

Synopt:
A method to add specific wotlk corrections is available on the `wotlk-support` branch so I've proposed the fix to that branch instead

PR #4082, opened 2022-09-09
Hide "In Search for Thaelrid" from horde once again

Synopt:
## Issue references

Fixes [4080](https://github.com/Questie/Questie/issues/4080)

## Proposed changes

- Remove the correction that shows "In Search of Thaelrid" to all races

### DIFF
--- a/Database/Corrections/Classic/classicQuestFixes.lua
+++ b/Database/Corrections/Classic/classicQuestFixes.lua
@@ -622,9 +622,6 @@ function QuestieQuestFixes:Load()
         [1193] = {
             [questKeys.specialFlags] = 1, -- #1348
         },
-        [1198] = {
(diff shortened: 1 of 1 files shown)


BreakBB:
Hey @Synopt instead of removing the correction from Classic you should add a replacement to the Wotlk quest fixes. Otherwise the quest will falsely show for Classic Era/SoM.

Synopt:
Ah thanks very much - I was wondering if you could point me at the wotlk corrections file? I was looking for it but couldn't find it

BreakBB:
Oh right, that file is atm only available on the `wotlk-branch`. The wotlk changes are not yet in master

Synopt:
I can PR to that branch instead and you can merge when it's ready 👌

PR #4115, opened 2022-09-13
Performance fixes

Logonz:
This contains a lot...
Improves performance on the readers (Thanks Laumesis!)
Fixes the validators, changes Query to Keys instead of creating a table each time.
Fixes so the CLI runs correctly, also including dockerfiles to run it locally
Performance fixes in Available quest
Emmy fixes


Database/QuestieDB.lua
```diff
@@ -1,31 +1,6 @@
...
----@field objectData table
----@field itemData table
----@field sortKeys table
----@field private private table
----@field private _itemAdapterQueryOrder table @temporary, until we remove the old db funcitons
----@field private _objectAdapterQueryOrder table @temporary, until we remove the old db funcitons
----@field private _questAdapterQueryOrder table @temporary, until we remove the old db funcitons
----@field private _npcAdapterQueryOrder table @temporary, until we remove the old db funcitons
 local QuestieDB = QuestieLoader:CreateModule("QuestieDB")
+QuestieDB.private = QuestieDB.private or {}
```
BreakBB, line 3: Why would that be necessary?

Modules/Libs/QuestieLib.lua
```diff
@@ -590,6 +573,26 @@ objectiveFontString:Hide()
 ---@param desiredWidth number @Set the desired width to wrap, default: 275
 ---@return table[] @A table of wrapped lines
 function QuestieLib:TextWrap(line, prefix, combineTrailing, splitOnDot, desiredWidth, questid)
+    if not objectiveFontString then        
+        if _G["QuestLogObjectivesText"] then -- classic
+            frameObject = _G["QuestLogObjectivesText"] 
+        elseif _G["QuestInfoObjectivesText"] then -- Wotlk Classic
+            frameObject = _G["QuestInfoObjectivesText"]
+        end
```
BreakBB, line 581: In most other files we do these kind of things in the very first lines of a fine and just concat them with `or` to get the last non nil one.

Like `local frameObject = QuestLogObjectivesText or QuestInfoObjectivesText`. Might be applicable here as well?

Modules/Quest/QuestieQuest.lua
```diff
@@ -3,6 +3,8 @@ local IsQuestFlaggedCompleted = IsQuestFlaggedCompleted or C_QuestLog.IsQuestFla
 
 ---@class QuestieQuest
 local QuestieQuest = QuestieLoader:CreateModule("QuestieQuest")
+---@type QuestieQuestPrivate
+QuestieQuest.private = QuestieQuest.private or {}
```
BreakBB, line 7: Same as before, why is this needed?

Modules/Quest/QuestieQuest.lua
```diff
@@ -1291,7 +1297,10 @@ function QuestieQuest:CalculateAndDrawAvailableQuestsIterative(callback)
     local showPvPQuests = Questie.db.char.showPvPQuests
     local showAQWarEffortQuests = Questie.db.char.showAQWarEffortQuests
 
+    --- Fast Localizations
     local autoBlacklist = QuestieQuest.autoBlacklist
+    local hiddenQuests = QuestieCorrections.hiddenQuests
+    local hidden  = Questie.db.char.hidden
```
BreakBB, line 1303:
```suggestion
    local hidden = Questie.db.char.hidden
```

Modules/Quest/QuestieQuest.lua
```diff
@@ -1332,6 +1341,7 @@ function QuestieQuest:CalculateAndDrawAvailableQuestsIterative(callback)
                                 --Draw a specific quest through the function
                                 _QuestieQuest:DrawAvailableQuest(quest)
                             else
+                                --* TODO: How the frames are handled needs to be reworked, why are we getting them from _G
```
BreakBB, line 1344: I fully agree. We are totally flooding the global namespace with our stupid frames

Modules/QuestieInit.lua
```diff
@@ -311,30 +344,31 @@ end
...
-                if type(ret) == "function" then -- continue to next stage, if it was returned by coroutine
-                    ret()
-                end
-            end
-        else
-            Questie:Error(l10n("Error during initialization!"), ret)
-            initFrame:SetScript("OnUpdate", nil)
-            initFrame:SetParent(nil)
-            initFrame = nil
+    local coStatus, coResume = coroutine.status, coroutine.resume
```
BreakBB, line 347: This could also be done outside of the function for even better performance, right?

Database/Classic/classicItemDB.lua
```diff
@@ -1,6 +1,6 @@
 -- AUTO GENERATED FILE! DO NOT EDIT!
 
----@type QuestieDB
+---@class QuestieDB
```
BreakBB, line 3: This should be reverted and I guess our QuestieDB types need a single file

Database/QuestieDB.lua
```diff
@@ -527,60 +510,86 @@ end
...
-    end
-
-    if Questie.db.char.hidden[questId] then
-        if debugPrint then Questie:Debug(Questie.DEBUG_SPAM, "[QuestieDB.IsDoable] quest is hidden manually!") end
-        return false
-    end
+    -- if QuestieCorrections.hiddenQuests[questId] then
+    --     if debugPrint then Questie:Debug(Questie.DEBUG_SPAM, "[QuestieDB.IsDoable] quest is hidden!") end
+    --     return false
+    -- end
```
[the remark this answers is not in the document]
Logonz: Good catch it is important!
The problem is that we check it twice when we calculate available quests but we also use isdoable in other places so it has to be there. Have to either add a hidden check everywhere or check it twice in available

Database/QuestieDB.lua
Logonz: The weirdness i do with the .private is for the ide to be able to understand that it is defined elsewhere and make it actually link the objects together. Because it knows a private field exists but it doesn't otherwise understand what is in it.
It doesn't actually have any impact on the actual running of the code it's purely for the IDE

Modules/Libs/QuestieLib.lua
Logonz: Yeah you are totally right!

BreakBB: Well maybe we can use a better name than `frameObject` which better reflects that we are referencing a default UI Frame?

Logonz:
What is happening here?

The best way to follow what happened is to look first from QuestieInit, Most of the functions it touches has been threaded better but also improved in performance where possible. Some initialization has been restructured but done in a safe way.

The compiler is mostly performance fixes that also had been identified by Laume, Skippers instead of readers are validated through the validators for Item,Quest,Object,NPCs:

Very little logic has been changed from previous mostly just shuffled around / optimized.

Database/compiler.lua
```diff
@@ -1122,53 +1131,135 @@ function QuestieDBCompiler:ValidateItems()
...
+            if (requiredLevel > level) then
+                questLevel = requiredLevel;
+            else
+                questLevel = level;
+                -- We also set the requiredLevel to the player level so the quest is not hidden without "show low level quests"
+                requiredLevel = level;
+            end
+        end
+        return questLevel, requiredLevel
+    end
```
BreakBB, line 1196: This can also be created outside `QuestieDBCompiler:ValidateQuests`

Database/compiler.lua
```diff
@@ -1186,13 +1277,15 @@ function QuestieDBCompiler:GetDBHandle(data, pointers, skipMap, keyToRootIndex,
     handle.stream = stream
 
     if overrides then
+        ---@param id QuestId|ObjectId|ItemId|NpcId
+        ---@param key string
+        ---@return any
         handle.QuerySingle = function(id, key)
```
BreakBB, line 1283: Inline function creation is bad, especially if the function it is created in is called multiple times. Can we extract that function together with all the other `handle.X = function` things below?

Modules/Quest/QuestieQuest.lua
```diff
@@ -1279,65 +1287,65 @@ function _QuestieQuest:GetQuestIcon(quest)
...
 
----@type Ticker?
-local timer
-function QuestieQuest:CalculateAndDrawAvailableQuestsIterative(callback)
-    Questie:Debug(Questie.DEBUG_INFO, "[QuestieQuest:CalculateAndDrawAvailableQuestsIterative] PlayerLevel =", QuestiePlayer.GetPlayerLevel())
-
-    -- Localize the variable for speeeeed
-    local debugEnabled = Questie.db.global.debugEnabled
+--? Creates a localized space where the local variables and functions are stored
+do
```
BreakBB, line 1291: Why is this needed? 🤔 

Top level logic is scary

Modules/Journey/QuestieJourney.lua
```diff
@@ -113,19 +117,23 @@ function QuestieJourney:IsShown()
 end
 
 function QuestieJourney:ToggleJourneyWindow()
-    if (not isWindowShown) then
-        PlaySound(882)
+    if QuestieJourneyFrame then
```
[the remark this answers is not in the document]
Logonz: With the current yielding there is a case of questie still initializing and someone clicking the minimap button

Modules/Libs/QuestieLib.lua
```diff
@@ -590,34 +573,52 @@ objectiveFontString:Hide()
 ---@param desiredWidth number @Set the desired width to wrap, default: 275
 ---@return table[] @A table of wrapped lines
 function QuestieLib:TextWrap(line, prefix, combineTrailing, splitOnDot, desiredWidth, questid)
-    if(objectiveFontString:IsVisible()) then Questie:Error("TextWrap already running... Please report this on GitHub or Discord.") end
+    if not textWrapObjectiveFontString then
+        --                    Classic                          Wotlk Classic
+        textWrapFrameObject = _G["QuestLogObjectivesText"] or _G["QuestInfoObjectivesText"]
```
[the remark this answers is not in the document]
Logonz: The reason i did i this way is because if we put it in root we also need to support it in the CLI, thought it was prettier to just do it within the function as it is pretty static. Or am i misunderstanding the question? :D

Database/compiler.lua
Logonz: It is only used in here so id rather not have it outside

Database/compiler.lua
Logonz: Well maybe, but i think that it is outside of the scope of this PR, this is the way Aero did it way back and i just added some Emmy for it and some performance changes.

Modules/Quest/QuestieQuest.lua
Logonz: I don't really understand?
The do block is used to not enable anything outside of it to be able to modify the timer variable which was more global before.
The only logic that escapes the do-block scope is the actual function that gets created into QuestieQuest but the local function and variable are only available within scope by design.

Modules/Journey/QuestieJourney.lua
```diff
@@ -41,17 +41,21 @@ QuestieJourney.questCategoryKeys = {
     EVENTS = 7,
 }
 
-function QuestieJourney:Initialize()
+
+---@param yield boolean?
+function QuestieJourney:Initialize(yield)
```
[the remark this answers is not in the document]
Logonz: Fixed

Database/compiler.lua
Logonz: furthermore a much nicer way of fixing this is in the refactor/DB branch, so i made a ticket for it and we'll fix it in there.

PR #4156, opened 2022-09-19
Added Aresella (18991) to first aid profession

stu177:
## Proposed changes

- Adds [Aresella](https://www.wowhead.com/wotlk/npc=18991/aresella) as a First Aid profession trainer

### DIFF
--- a/Modules/QuestieMenu.lua
+++ b/Modules/QuestieMenu.lua
@@ -464,7 +464,9 @@ function QuestieMenu:PopulateTownsfolk()
     tinsert(professionTrainers[QuestieProfessions.professionKeys.FISHING], 10216)
 
     -- Fix NPC Aresella (18991) can train first aid profession
-    tinsert(professionTrainers[QuestieProfessions.professionKeys.FIRST_AID], 18991)
+    if Questie.IsTBC then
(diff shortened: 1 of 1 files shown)


Modules/QuestieMenu.lua
```diff
@@ -463,6 +463,9 @@ function QuestieMenu:PopulateTownsfolk()
     -- Fix NPC Gubber Blump (10216) can train fishing profession
     tinsert(professionTrainers[QuestieProfessions.professionKeys.FISHING], 10216)
 
+    -- Fix NPC Aresella (18991) can train first aid profession
+    tinsert(professionTrainers[QuestieProfessions.professionKeys.FIRST_AID], 18991)
```
BreakBB, line 467:
```suggestion
    if Questie.IsTBC then
        tinsert(professionTrainers[QuestieProfessions.professionKeys.FIRST_AID], 18991)
    end
```

BreakBB (approved):

BreakBB:
Hey @stu177 thanks a lot for this nice small PR 👍🏻 

I updated it to only be added for TBC and Wotlk so there is no conflict for Classic Era/SoM players, but it's good you noticed that NPC needed to be added 🥳

stu177:
Hey, no problem 😄

Ah yeh that makes sense with only showing for TBC/WotLK 👍

PR #4269, opened 2022-09-30
XP at max level french traduction

Tib0b:
## Proposed changes

- French traduction for new option (837d162548800c0625b87e078f497eaee6fe43a3)


BreakBB:
Thanks @Tib0b for these translations 👍🏻

PR #4345, opened 2022-10-06
[quest] fixes for 12979

### LINKED ISSUE #4325: Missing objective data for quest 12979 and objective Dark Armor Sample
Missing objective data for quest 12979 and objective Dark Armor Sample
Quest started by https://www.wowhead.com/wotlk/item=42203/dark-armor-plate
Item dropped by https://www.wowhead.com/wotlk/npc=29380/stormforged-war-golem

Questie version 7.3.0 - no updates pending
Wow version: wotlk classic - 3.4.0.45854

dscreet:
## Issue references

Fixes #4325

## Proposed changes

- Added objective data for quest

### DIFF
--- a/Database/Corrections/Wotlk/wotlkItemFixes.lua
+++ b/Database/Corrections/Wotlk/wotlkItemFixes.lua
@@ -206,6 +206,10 @@ function QuestieWotlkItemFixes:Load()
         [41399] = {
             [itemKeys.npcDrops] = {},
         },
+        [42203] = {
+            [itemKeys.npcDrops] = {29380},
+            [itemKeys.startQuest] = 12979,
+        },
         [42252] = {
             [itemKeys.objectDrops] = {192124,192127},
         },
--- a/Database/Corrections/Wotlk/wotlkQuestFixes.lua
+++ b/Database/Corrections/Wotlk/wotlkQuestFixes.lua
@@ -1036,6 +1036,9 @@ function QuestieWotlkQuestFixes:Load()
             [questKeys.name] = "Hodir's Call",
             [questKeys.objectives] = {{{29974,"Niffelem Forefather freed"},},nil,nil,nil,{{{30144,30135},30144,},},},
         },
+        [12979] = {
+			[questKeys.objectives] = {nil,nil,{{42204},},nil,nil,},
+		},
         [12981] = {
             [questKeys.preQuestSingle] = {12967},
         },


Database/Corrections/Wotlk/wotlkQuestFixes.lua
```diff
@@ -1036,6 +1036,9 @@ function QuestieWotlkQuestFixes:Load()
             [questKeys.name] = "Hodir's Call",
             [questKeys.objectives] = {{{29974,"Niffelem Forefather freed"},},nil,nil,nil,{{{30144,30135},30144,},},},
         },
+        [12979] = {
+			[questKeys.objectives] = {nil,nil,{{42204},},nil,nil,},
```
Yttrium-tYcLief, line 1040: This is fine, but just for future reference you don't need to `nil` the rest of the table *after* your data. The preceding `nil`s are to ensure your data goes into the 3rd parameter (itemObjective), anything after that is needless (but legal) unless you need to add data to a later parameter.

The very act of re-defining the table will wipe out any existing data in `questKeys.objectives`.

Database/Corrections/Wotlk/wotlkItemFixes.lua
```diff
@@ -206,6 +206,10 @@ function QuestieWotlkItemFixes:Load()
         [41399] = {
             [itemKeys.npcDrops] = {},
         },
+        [42203] = {
+            [itemKeys.npcDrops] = {29380},
+            [itemKeys.startQuest] = 12979,
```
Yttrium-tYcLief, line 211: Does defining `itemKeys.startQuest` fix the chat prompt ("Picked up X item which starts Y quest!")? I've noticed that none of the WotLK quest-start items print that text, and as far as I can see they're missing this `itemKey`.

Yttrium-tYcLief:
Added a few comments but overall looks good to me!

Database/Corrections/Wotlk/wotlkItemFixes.lua
BreakBB: Yes, that is absolute true. Good catch @dscreet 💪🏻 I'll see if we can script these missing fields into the other quest starting items.

BreakBB (approved):

BreakBB:
Hey @dscreet this looks correct to me 👍🏻 Thank you very much for taking a look into it and providing this PR 🥳

Database/Corrections/Wotlk/wotlkQuestFixes.lua
dscreet: Gotcha, thanks.

Database/Corrections/Wotlk/wotlkItemFixes.lua
dscreet: Didn't know that was the case haha. I just double-checked all fields and saw this was missing. Gd to know.

PR #4354, opened 2022-10-07
Add Automated Builds, update minimap and tracker icons

### LINKED ISSUE #2645: Setup BigWigsMods packager
Looks like something that will really make releases a lot easier. 

Need to fully understand how it would work for us, and likely make a few changes to our Git branch processes, but seems really promising.

@BreakBB & @AeroScripts feel free to reach out to @QartemisT in Discord - she's currently using this on other projects.

https://github.com/BigWigsMods/packager

miracledinner:
## Issue references

Fixes #2645

## Proposed changes

-Add automated builds with BigWigs Package Manager
In order for this to function, you will need to do a few small things:
  

1. Get an API token from CurseForge
2. Add the API key in a GitHub Secret called "CF_API_KEY"
3. Ensure CurseForge project ID is in all relevant .toc files `## X-Curse-Project-ID: ######` it seems like it is currently there already
4. Update the "tags" variable in `.github/workflows/package.yml` to regex matching how you would like to tag releases
5. Currently * will tag anything in master branch as an alpha release
6. That's it, any time you tag a release it will be sent out to CurseForge, WoWUp, etc

-Update minimap icon and quest tracker header to use the project logo

## Screenshots

Logo Icon:
image
Relevant section of `package.yml`:
image


BreakBB:
Hey @miracledinner uhm... WOW! 🤩 Thanks a lot for looking into this and improve our pipeline 💪🏻 

I moved that topic again and again and never found the time to actually investigate how it's correctly done.

Trhee questions though:

1. In the past we used the `changelog.py` file to generate a proper changelog. It would be nice to keep that, but from my understanding the packager simply pickups "all" commits?

2. We also use `build.py` to only bundle specific files/folders. I guess we also need an exclusion file for the package to pick up on this rule?

3. We still want to ship Questie for Classic Era/SoM, I guess we just extend the script for this, right?

BreakBB:
Moreover me and the rest of the team don't think that icon change is a good idea. I mean we like that icon, but it doesn't really fit in the game.

Could you revert these changes?

Muehe:
@BreakBB Ignoring files is possible with a ".pkgmeta" file: https://github.com/BigWigsMods/packager/wiki/Preparing-the-PackageMeta-File

Also lots of other stuff in there we probably need to take care of, like attributing libraries, but didn't look into it in detail. Not sure about our scripts, but we might be able to include them by modifying the workflow.

As for the additional commits, we could just cherry-pick one of the two commits adding the package workflow and go from there. @TannerShmoog are you fine with that?

TannerShmoog:
> @BreakBB Ignoring files is possible with a ".pkgmeta" file: https://github.com/BigWigsMods/packager/wiki/Preparing-the-PackageMeta-File
> 
> Also lots of other stuff in there we probably need to take care of, like attributing libraries, but didn't look into it in detail. Not sure about our scripts, but we might be able to include them by modifying the workflow.
> 
> As for the additional commits, we could just cherry-pick one of the two commits adding the package workflow and go from there. @TannerShmoog are you fine with that?

Absolutely, I did them as seperate commits so that you could choose what you want. Feel free to whatever you want.

Muehe:
> We still want to ship Questie for Classic Era/SoM, I guess we just extend the script for this, right?

BigWigs [doesn't seem to support](https://github.com/BigWigsMods/packager/blob/master/release.sh#L259) doing multi-version releases. We could just make a second workflow for Era I guess, but that would probably result in separate uploads on CF I think...

Muehe:
@TannerShmoog @miracledinner Just to recap where we are at with this.

There are some road-blocks to merging this PR (or rather the workflow) as is.

- We need a pkgmeta file to exclude unused files from releases (>80MB) and give attribution to libraries so they get credits for the CF rewards program.
- We need a way to add our changelog. Should be a relatively simple wrapper-script though.
- The script forces one upload per expansion apparently, so we can't do a single upload for both versions anymore. Not a really road-block, just a bit annoying.
- Github secrets need to be added (just noting it down here for reference).
- Biggest problem is that the script errors though, I ran it locally [according to their instructions](https://github.com/BigWigsMods/packager#building-locally) which resulted in this error, no matter which options I threw at it:
    `Ambiguous addon name. No fallback TOC file or addon name includes an expansion suffix (-WOTLKC). Set 'package-as' in .pkgmeta`
    Even adding a pkgmeta file didn't help sadly. Not sure what's going on exactly, might be related to our default TOC file just loading an error message but not sure...

In conclusion we either have to change/add a lot of stuff or we could straight-up just take the upload functions from the BigWigs release script and modify them to suit our purposes (that *might* integrate better with our existing scripts).

If anybody has an idea how to get around that error please let me know.

MyTechnoHunter:
https://github.com/BigWigsMods/packager#multiple-toc-files

Basically, multi game support needs proper named toc files OR one toc file with multi interface lines..

 ie. 
Questie.toc
Questie_Wrath.toc
So on and so forth. Sorry if you already tried that

Muehe:
@MyTechnoHunter It definitely doesn't like our base file, but I figured out why pkgmeta file wasn't working at least. Typo in the options. :facepalm: But now I need to add ignores or it copies all the old releases I made, so no further progress to report yet.

Muehe:
Ok, so here is how to get the script running locally:

- Create a `.release` directory and put the `release.sh` file from BigWigs packager in it.
- Create a `.pkgmeta` file into the same directory with this content:
```yaml
package-as: Questie

ignore:
  - .git
  - .github
  - .types
  - .vscode
  - ExternalScripts(DONOTINCLUDEINRELEASE)
  - releases
  - .gitattributes
  - .gitignore
  - .luacheckrc
  - cli.lua
  - "*.py"

```
- Run this command: `./release.sh -d -e -l -m .pkgmeta`

This just creates the package locally without uploading it. The created package is actually a single zip containing all the TOC files like we are used to (note that I did **not** provide the `-g wrath` option as the PR does). Testing how the resulting uploads would look is still TODO (disabled by `-d` option and not providing keys).

Bit concerned with script runtime, it took 4m19s, against 5s for our own build script (which does basically the same thing the release script does with the options provided, move files and zip them).

Muehe:
@TannerShmoog @miracledinner As discussed I have cherry-picked the initial commit and started a new branch/PR for further discussion please follow #4446

Thanks again. :+1:

Issue #4427, opened 2022-10-20
Questie is Killing FPS During Raids


Gogo1951:
Frame rates were going down to 10 FPS. Turning Questie off, it's back up to 120 FPS.

This was 20 seconds of profiler, seems like a tooltip comms thing.

![image](https://user-images.githubusercontent.com/55365231/196836894-0f01b11a-da95-4866-aaef-91f54aef2b8f.png)

stratoru:
![image](https://user-images.githubusercontent.com/43893231/196852142-769ced41-4402-47cc-bb86-683ad91fe003.png)


Here is a screenshot of my profiler just for some additional information. I have also had some severe performance issues with Questie enabled, this is not in a raid or group btw. The "QuestieMap.fadeLogicTimerShown._callback" and "QuestieMap.drawTimer._callback" and "QuestieLib.Euclid" are the three that seem to be exponentially increasing constantly even when afk. The others only have large calls in bursts when activity is occurring.

Gogo1951:
![image](https://user-images.githubusercontent.com/55365231/197094916-8f779621-e001-4637-9f3f-5c7b4ad58548.png)

Hey guys, this issue is all over. When I go to loot. When I'm in town. When I'm solo. When I'm in a dungeon. It's grinding my game to a halt. Then I turn Questie off and it's right as rain. Something is really wrong with the last build. I can't run Questie due to performance issues, and everyone I know in game is messaging me to say the same. 

It seems to get better when I comment out the Journey stuff specifically. Journey needs an off switch. You can turn everything else off. But... I don't know if that's 100% it. Just... this last build has made Questie unusable in-game.

BreakBB:
Hey @Gogo1951 so the profiler record you posted are not that bad (even if they might look like it). The last 5 or more releases almost exclusively fixed our databases, no major refactorings/logic changes were added 🤔 

Would be nice if you could try to play without just Questie enabled to see if there is a conflict with some other addon.

We will have a detailled look into the functionsabove though. Especially the first one you posted shows tooltip code which shouldn't run when you are in a raid (but that wouldn't explain the loss of 130 FPS).

Gogo1951:
Yeah... I'm grasping as straws. I think maybe server throttle?

The report isn't crazy. But like just standing around crafting some fish feasts, I had a 2-3 second lag at the end of the craft.

[...]

PR #4428, opened 2022-10-20
Adding a few items that drop all over to ignore list

Gogo1951:
May help with this, not sure. 

https://github.com/Questie/Questie/issues/4427


BreakBB:
Not related to the issue above though.

Issue #4213, opened 2022-09-28
Lua error QuestieMap.lua:836


Padparadscha:
Every time I get an Item from the NPC who gives me the quest I get following LUA error

Message: Interface\AddOns\Questie\Modules\Map\QuestieMap.lua:836: bad argument #1 to 'next' (table expected, got nil)
Time: Wed Sep 28 17:48:17 2022
Count: 1
Stack: Interface\AddOns\Questie\Modules\Map\QuestieMap.lua:836: bad argument #1 to 'next' (table expected, got nil)

Locals:

Padparadscha:
Nearly every Quest in Kaltarra

BreakBB:
@Padparadscha could you try to recompile your database? `/questie` -> Advanced -> Recompile Database.

And also make sure you are running the latest version v7.2.3 atm

Padparadscha:
version is 7.2.3. recompiling the Database did not work. Tested with the same quest.
I tried a little bit more. The Fault is in combination with Arkinventory (version 3.09.68)

khan8037:
v7.3 didn't work for me either. I don't think it's related to Arkinventory specifically because I also get the error with Bagnon or different map addons like Leatrix and Mapster

red13dotnet:
I just wanted to chime in to confirm I am seeing this as well, Wrath classic, Version 7.3.  Bagnon and leatrix plus in use.
There are several similar errors to this cropping up and the other similarity beyond map/bag mods seems to be addons that modify/ add to tooltips.

matthewmb:
I have also seen this bug regularly in WOTLK Classic. Here's what I tried:

Tested with quest "Cheese for Glowerfield" in Dalaran, but it occurs randomly on other quests too.

1. Disable all addons, abandon quest, reload UI - still get error

2. Disable all addons, abandon quest, recompile Questie database, reload UI - still get error

3. Disable all addons, abandon quest, reset Questie, reload UI - no error

4. All addons enabled, abandon quest, reset Questie, reload UI - no error

So resetting Questie seems to fix the issue, but I have done this previously and had the issue re-occur.

Padparadscha:
> I have also seen this bug regularly in WOTLK Classic. Here's what I tried:
> 
> Tested with quest "Cheese for Glowerfield" in Dalaran, but it occurs randomly on other quests too.
> 
> 1. Disable all addons, abandon quest, reload UI - still get error
[...]

PR #4438, opened 2022-10-24
Add small delay to GetNearestSpawn

matthewmb:
## Issue references

Fixes https://github.com/Questie/Questie/issues/4213

## Proposed changes

- Adds a short delay before calculating the nearest objective, preventing an intermittent LUA error

## Screenshots

N/A

### DIFF
--- a/Modules/Map/QuestieMap.lua
+++ b/Modules/Map/QuestieMap.lua
@@ -841,30 +841,32 @@ function QuestieMap:GetNearestSpawn(objective)
     local playerX, playerY, playerI = HBD:GetPlayerWorldPosition()
     local bestDistance = 999999999
     local bestSpawn, bestSpawnZone, bestSpawnId, bestSpawnType, bestSpawnName
-    if next(objective.spawnList) then
-        for id, spawnData in pairs(objective.spawnList) do
-            for zone, spawns in pairs(spawnData.Spawns) do
-                for _,spawn in pairs(spawns) do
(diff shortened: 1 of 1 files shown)


BreakBB:
Hey @matthewmb thanks for taking the time to look into this issue. When adding a timer fixes this error, then the actual cause of the error is most likely a race condition. We should dig deeper on this to find what really is causing `objective.spawnList` to be `nil` instead of a table 🤔

matthewmb:
I agree that this doesn't fix the cause of the issue, but it does prevent the error with minimal impact for the end user. At least for me, the error was frequent, and completely broke the tracker each time.

I am not familiar enough with the project to investigate a race condition. If there are more knowledgeable maintainers who have capacity to resolve this, that would be ideal. If not, perhaps this can be applied as a temporary fix to be revisited later.

Modules/Map/QuestieMap.lua
```diff
@@ -841,30 +841,32 @@ function QuestieMap:GetNearestSpawn(objective)
...
-                            bestSpawnId = id
-                            bestSpawnType = spawnData.Type
-                            bestSpawnName = spawnData.Name
-                        end
-                    end
-                end
-            end
-        end
-    end
+	C_Timer.After(2, function() -- Fixes LUA error bad argument #1 to 'next' (table expected, got nil)
```
BreakBB, line 844:
```suggestion
    -- TODO: This is just a temporary workaround - We have to find out why "objective.spawnList" can be nil
	C_Timer.After(2, function()
```

Issue #4367, opened 2022-10-08
[[80] Sneak Preview (13351)] - Location have kill markers instead of the  magnifying glass


Cryces:
![image](https://user-images.githubusercontent.com/2063902/194709970-287a19b5-ec69-4015-a9f7-0e4fb81332fe.png)

Yttrium-tYcLief:
This is a multi-layered issue with the way the database works and the way objectives are set up in the game itself.

Internally within Questie, we have two ways of displaying objective markers: Objectives and extraObjectives. Objectives are the primary objectives of the quest. These can be "monster", "object", "item", "reputation", "killCredit", or "player". It's possible there are others but those are the core ones to talk about here. For those objectives, Questie *only* uses certain icons:

- For `monsterObjective`, Questie displays a slay icon (sword) at the known coordinates for the monster
- For `objectObjective`, Questie displays an object icon (gear) at the known coordinates for the object
- For `itemObjective`, Questie checks to see how that item is obtained; it flags monsters that drop the item with a slay icon (sword), and objects that contain the item with an object icon (gear)
- For `reputationObjective`, Questie doesn't display anything
- For `killCreditObjective`, Questie displays a slay icon (sword) at the known coordinates for the eligible monsters
- For `playerObjective`, Questie actually does not understand this objective type (yet)

extraObjectives are basically manual icons that we can place at any point; we can target NPCs, objects, raw coordinates, items, etc. We choose what custom text is used for the label, and we can choose what icon to display. If you ever see an event (magnifying glass) or talk (speech bubble) icon, these are guaranteed to be extraObjectives.

As an example: Let's say we have a quest where you have to go to a certain location, talk to an NPC, and then an enemy spawns, and you have to kill the enemy. The quest objective might simply be a `monsterObjective` to kill the spawned enemy mob. However, the NPC you have to interact with is not marked, so Questie's objective icons for this quest can seem a bit unintuitive to users. In this situation, we typically add an extraObjective targeting the NPC you have to interact with to summon the mob, and we use a text label indicating as such.

*However*, certain quests structured this exact same way may have their underlying Objectives laid out a bit differently; they may have one objective for "0/1 Spoken to NPC", and then a separate objective for "0/1 Monster slain". While these are clearly different types of objectives, under the hood, they're *both* monsterObjectives, so Questie draws a slay icon for both.

[...]

PR #4439, opened 2022-10-24
[fix] Improve icons for objectives on friendly NPCs

### LINKED ISSUE #4355: [[75] In the Name of Loken (12204)] - Kill markers instead of dialog markers
Incorrect map markers. Quest npcs are marked with the sword instead of the speech bubble.

### LINKED ISSUE #4356: [[78] Into The Wild Green Yonder (13045)] - NPCs to rescue marked with kill marker
image

BreakBB:
## Proposed changes

- This replaces the "slay icons" on friendly NPCs with a new "interact icon" (see comment below).

Fixes #4355
Fixes #4356
Partially addresses #4367

TODOs:

- [ ] Extract new information to recognize neutral NPCs. Otherwise the neutral NPCs are also marked with the new icon instead of a slay icon


Yttrium-tYcLief:
In Discord we've discussed potential icon options, since a talk icon doesn't really feel appropriate for most situations. The one we've all seemingly agreed upon is `Pointer_Map_Pin`:

image

In addition - would it be possible to detect completely uncategorized NPCs like [KC bunnies](https://www.wowhead.com/wotlk/npc=32195/south-gate-kc-bunny)? We could use Event icons for these mobs, which might feel a bit more appropriate with how they're typically targeted (by default).

BreakBB:
Hey @Yttrium-tYcLief I am not sure if we can (or should) detect the KC bunnies 🤔 In general we want to replace all of these references as they are some "bad private server data" and not something really useful, right?

Yttrium-tYcLief:
To clarify I didn't mean to suggest we leave them as-is - just was curious if we could change the default behavior for how we treat them, in case it's a quest/situation we haven't manually reviewed yet.

BreakBB:
Currently the [MapPinCursor](https://github.com/Gethe/wow-ui-textures/blob/d23deaf8f44a7d280dc974a5c9d5321c013db59b/CURSOR/MapPinCursor.PNG) does not seem to be available in Wrath Classic. We have to check if the Dragonflight patch changes this.

PR #4443, opened 2022-10-25
Fix quest Catch and Release (9629)

gouletr:
Fix quest [Catch and Release (9629)](https://www.wowhead.com/tbc/quest=9629/catch-and-release) missing extra objectives.

### DIFF
--- a/Database/Corrections/Wotlk/wotlkQuestFixes.lua
+++ b/Database/Corrections/Wotlk/wotlkQuestFixes.lua
@@ -52,6 +52,9 @@ function QuestieWotlkQuestFixes:Load()
         [9247] = {
             [questKeys.finishedBy] = {{16281}},
         },
+        [9629] = {
+            [questKeys.objectives] = {{{17326}}},
+        },
         [9648] = {
             [questKeys.name] = "Maatparm Mushroom Menagerie",
         },


Yttrium-tYcLief:
What's the benefit of this? That NPC ID should already be tagged with "0/6 Blacksilt Scouts Tagged" with the native objectives.

Yttrium-tYcLief:
Oh, I see what's going on here.

Questie loads/patches data in the following order, with each successive entry overwriting the past (so wrath corrections take highest priority):
**Classic Era client**: Classic DB < Classic corrections
**TBC client**: TBC DB < Classic corrections < TBC corrections
**Wrath client**: Wrath DB < Classic corrections < TBC corrections < Wrath corrections

In this situation, this is a regression in our Wrath database, not an issue with the TBC data. The TBC DB shows the following data for the objective:
```lua
{{{17326,"Blacksilt Scouts Tagged"}}}
```

But the Wrath DB shows the following data for the objective:
```lua
{{{17654,"Blacksilt Scouts Tagged"}}}
```

Technically this is somewhat correct, but it doesn't work for end users, since 17654 has no coordinates. But even if we gave it coordinates, it doesn't natively spawn in the world, so there would be no tooltips.

The correct way to handle this situation would be to add an entry in wotlkQuestFixes for the following:
```lua
[9629] = {
    [questKeys.objectives] = {{{17326}}},
},
```

This will fix it in the Wrath DB without unnecessarily patching it in the TBC DB.

gouletr:
>The correct way to handle this situation would be to add an entry in wotlkQuestFixes

Done.

Yttrium-tYcLief:
Looks good to me, but since this targets Master directly I'll wait for one of the main devs to merge it.

PR #4572, opened 2023-01-02
Fix hunter taming quests

gouletr:
Fix hunter taming quests locations; use objectives rather than triggers.

This has the following advantages:
- No need to specify the objective text
- A lot more locations now appear on the map, which is super useful when someone already tamed the npc where you expected to find it
- This in turn revealed many more locations which are much closer to the quest giver, such as the Strigid Screecher for Night Elf hunters
- In the future, if more locations are added/changed for these npcs, then they will automatically apply to the quest objectives

One thing is that the icons on the map will be a sword instead of a magnifying glass, which is no big deal imho.

### DIFF
--- a/Database/Corrections/Classic/classicQuestFixes.lua
+++ b/Database/Corrections/Classic/classicQuestFixes.lua
@@ -1890,16 +1890,24 @@ function QuestieQuestFixes:Load()
             },
         },
         [6061] = {
-            [questKeys.triggerEnd] = {"Tame an Adult Plainstrider", {[zoneIDs.MULGORE]={{43.5,44.99},{43.69,51.84},{52.68,57.44},{41.96,55.26},{48.15,47.89}}}},
+            [questKeys.objectives] = {{{2956, nil}}},
+            [questKeys.questFlags] = 0,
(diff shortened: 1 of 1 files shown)


BreakBB:
Hey @gouletr did you verfiy these changes? Iirc Questie needs `triggerEnd` objectives if the quest objectives are of type `event` or otherwise Questie won't match them (or at least raise an error).

gouletr:
These quests are definitely not event related. There is no area trigger or special external event required to complete these. Using the taming rod on any of the specified npc will complete the quest. I believe the `questFlags` and `specialFlags` fields in the DB are wrong and needs to be reset to 0.

gouletr:
I verified the changes, after setting the quest flags to 0, and everything works as expected:
- npc icons appears on the map with the correct text objective
- the quest tracker shows the taming rod icon, and it is functional
- after taming the npc, questie properly updated and the quest was marked complete

gouletr:
Note that these changes are in Classic, so I tested them on Classic Era server today, both alliance and horde.

BreakBB:
Hey @gouletr thanks for taking the extra time to verfiy your changes 👍🏻

PR #4657, opened 2023-02-11
Add requiredSpell and requiredSpecialization to quest database

Yttrium-tYcLief:
This PR adds two DB parameters, `requiredSpell` and `requiredSpecialization`. It also adds a list of specializationKeys to QuestieProfessions.

`requiredSpell` means the quest will not show as available if the player lacks the spellID specified.

`requiredSpecialization` has some slightly complex logic. If fed a specializationKey (new), the quest will be hidden if the player does not have that specialization, checked with the basic `IsSpellKnown()` API call. If fed a professionKey, the quest will be hidden if the player either lacks that profession entirely, or has the profession *and* any of its specializations learned. In essence, this is a test to see if the player has *just* the base profession learned, and is therefore eligible to learn a specialization.

By default, all quests have these fields set to `nil`, and these logic trees are only entered (during QuestieDB.IsDoable) if data is found in these fields. Because these cases are so rare, these are the absolute last fields checked - if the quest is ruled as "not doable" from any other requirement prior, we don't even reach this logic.

`requiredSpell` is, as far as I know, not currently needed for any quests up to Wrath, but it may be required for something in Cataclysm. `requiredSpecialization` is already needed in Vanilla.

### DIFF
--- a/Database/Classic/classicQuestDB.lua
+++ b/Database/Classic/classicQuestDB.lua
@@ -41,6 +41,8 @@ QuestieDB.questKeys = {
     ['parentQuest'] = 25, -- int, the ID of the parent quest that needs to be active for the current one to be available. See also 'childQuests' (field 14)
     ['rewardReputation'] = 26, --table: {{faction(int), value(int)},...}, a list of reputation rewarded upon quest completion
     ['extraObjectives'] = 27, -- table: {{spawnlist, iconFile, text, objectiveIndex (optional), {{dbReferenceType, id}, ...} (optional)},...}, a list of hidden special objectives for a quest. Similar to requiredSourceItems
+    ['requiredSpell'] = 28, -- int: quest is only available if character has this spellID
+    ['requiredSpecialization'] = 29, -- int: quest is only available if character meets the spec requirements. Use QuestieProfessions.specializationKeys for having a spec, or QuestieProfessions.professionKeys to indicate having the profession with no spec. See QuestieProfessions.lua for more info.
 }
 
 QuestieDB.questData = [[return {
--- a/Database/QuestieDB.lua
+++ b/Database/QuestieDB.lua
@@ -633,7 +633,7 @@ function QuestieDB.IsDoable(questId, debugPrint)
     local requiredSpecialization = QuestieDB.QueryQuestSingle(questId, "requiredSpecialization")
     if (requiredSpecialization) and (requiredSpecialization > 0) then
         local hasSpecialization = QuestieProfessions:HasSpecialization(requiredSpecialization)
-        if (not (hasSpecialization)) then
+        if (not hasSpecialization) then
             if debugPrint then Questie:Debug(Questie.DEBUG_SPAM, "[QuestieDB.IsDoable] Player does not meet spell requirements for", questId) end
             --? We don't blacklist purely based on lacking the spell, since it's rare and can change quickly
             --if(not hasSpell) then
@@ -646,7 +646,7 @@ function QuestieDB.IsDoable(questId, debugPrint)
     local requiredSpell = QuestieDB.QueryQuestSingle(questId, "requiredSpell")
     if (requiredSpell) and (requiredSpell > 0) then
         local hasSpell = IsSpellKnownOrOverridesKnown(requiredSpell)
-        if (not (hasSpell)) then
+        if (not hasSpell) then
             if debugPrint then Questie:Debug(Questie.DEBUG_SPAM, "[QuestieDB.IsDoable] Player does not meet spell requirements for", questId) end
             --? We don't blacklist purely based on lacking the spell, since it's rare and can change quickly
             --if(not hasSpell) then
--- a/Database/QuestieDB.lua
+++ b/Database/QuestieDB.lua
@@ -630,6 +630,32 @@ function QuestieDB.IsDoable(questId, debugPrint)
         return false
     end
 
+    local requiredSpecialization = QuestieDB.QueryQuestSingle(questId, "requiredSpecialization")
+    if (requiredSpecialization) and (requiredSpecialization > 0) then
+        local hasSpecialization = QuestieProfessions:HasSpecialization(requiredSpecialization)
+        if (not (hasSpecialization)) then
+            if debugPrint then Questie:Debug(Questie.DEBUG_SPAM, "[QuestieDB.IsDoable] Player does not meet spell requirements for", questId) end
+            --? We don't blacklist purely based on lacking the spell, since it's rare and can change quickly
+            --if(not hasSpell) then
+            --    QuestieQuest.autoBlacklist[questId] = "spell"
+            --end
+            return false
+        end
+    end
+
+    local requiredSpell = QuestieDB.QueryQuestSingle(questId, "requiredSpell")
+    if (requiredSpell) and (requiredSpell > 0) then
+        local hasSpell = IsSpellKnownOrOverridesKnown(requiredSpell)
+        if (not (hasSpell)) then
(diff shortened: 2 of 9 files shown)


Modules/QuestieProfessions.lua
```diff
@@ -179,11 +226,59 @@ local sortIds = {
...
+}
+
 ---@return string
 function QuestieProfessions:GetProfessionName(professionKey)
     return professionNames[professionKey]
 end
 
+---@return string
+function QuestieProfessions:GetSpecializationName(specializationKey)
+    return specializationNames[professionKey]
```
BreakBB, line 279:
```suggestion
    return specializationNames[specializationKey]
```

Modules/QuestieProfessions.lua
```diff
@@ -9,6 +9,7 @@ local l10n = QuestieLoader:ImportModule("l10n")
 local playerProfessions = {}
 local professionTable = {}
 local professionNames = {}
+local specializationNames = {}
```
BreakBB, line 12:
```suggestion
local specializationNames
```

Modules/QuestieProfessions.lua
```diff
@@ -124,6 +125,52 @@ function QuestieProfessions:HasProfessionAndSkillLevel(requiredSkill)
...
+---@return boolean HasSpecialization
+function QuestieProfessions:HasSpecialization(requiredSpecialization)
+    if not requiredSpecialization then
+        --? We return true here because otherwise we would have to check for nil everywhere
+        return true
+    end
+    for _, value in pairs(QuestieProfessions.professionKeys) do
+        if value == requiredSpecialization then -- if we determine input is a profession
+            if requiredSpecialization == QuestieProfessions.professionKeys.ALCHEMY then
+                return not (IsSpellKnown(QuestieProfessions.specializationKeys.ALCHEMY_ELIXIR)
```
BreakBB, line 138: I think it is a good idea to add a local reference of `IsSpellKnown`

Modules/QuestieProfessions.lua
```diff
@@ -124,6 +125,52 @@ function QuestieProfessions:HasProfessionAndSkillLevel(requiredSkill)
...
 end
 
+---@param requiredSpecialization { [1]: number } [1] = professionId
+---@return boolean HasSpecialization
+function QuestieProfessions:HasSpecialization(requiredSpecialization)
+    if not requiredSpecialization then
+        --? We return true here because otherwise we would have to check for nil everywhere
+        return true
+    end
+    for _, value in pairs(QuestieProfessions.professionKeys) do
```
BreakBB, line 135: You could store `local professionKeys = QuestieProfessions.professionKeys` before the for-loop together with `QuestieProfessions.specializationKeys` to improve readability of the lower block.

Database/QuestieDB.lua
```diff
@@ -630,6 +630,32 @@ function QuestieDB.IsDoable(questId, debugPrint)
         return false
     end
 
+    local requiredSpecialization = QuestieDB.QueryQuestSingle(questId, "requiredSpecialization")
+    if (requiredSpecialization) and (requiredSpecialization > 0) then
+        local hasSpecialization = QuestieProfessions:HasSpecialization(requiredSpecialization)
+        if (not (hasSpecialization)) then
```
BreakBB, line 636:
```suggestion
        if (not hasSpecialization) then
```

Database/QuestieDB.lua
```diff
@@ -630,6 +630,32 @@ function QuestieDB.IsDoable(questId, debugPrint)
...
+            --    QuestieQuest.autoBlacklist[questId] = "spell"
+            --end
+            return false
+        end
+    end
+
+    local requiredSpell = QuestieDB.QueryQuestSingle(questId, "requiredSpell")
+    if (requiredSpell) and (requiredSpell > 0) then
+        local hasSpell = IsSpellKnownOrOverridesKnown(requiredSpell)
+        if (not (hasSpell)) then
```
BreakBB, line 649:
```suggestion
        if (not hasSpell) then
```

BreakBB (changes requested):
Looks quite fine to me! Thanks a lot for these changes 🥳

I left a few minor suggestions (also fixing the pipeline).

Also do you have a quest to test this change with?

Database/Classic/classicQuestDB.lua
```diff
@@ -41,6 +41,8 @@ QuestieDB.questKeys = {
     ['parentQuest'] = 25, -- int, the ID of the parent quest that needs to be active for the current one to be available. See also 'childQuests' (field 14)
     ['rewardReputation'] = 26, --table: {{faction(int), value(int)},...}, a list of reputation rewarded upon quest completion
     ['extraObjectives'] = 27, -- table: {{spawnlist, iconFile, text, objectiveIndex (optional), {{dbReferenceType, id}, ...} (optional)},...}, a list of hidden special objectives for a quest. Similar to requiredSourceItems
+    ['requiredSpell'] = 28, -- int: quest is only available if character has this spellID
+    ['requiredSpecialization'] = 29, -- int: quest is only available if character meets the spec requirements. Use QuestieProfessions.specializationKeys for having a spec, or QuestieProfessions.professionKeys to indicate having the profession with no spec. See QuestieProfessions.lua for more info.
```
Muehe, line 45: It says `-- AUTO GENERATED FILE! DO NOT EDIT!` at the top of the file, please :sob: :pray: 

Same for the other DB files, they are changed by editing https://github.com/Muehe/sqlua and extracting a new DB with it.

In this case we can leave it I guess, will look into getting data for those fields which I can extract, which will then just overwrite these changes. But in general don't add *anything* to these files, as the changes might get lost on DB updates otherwise.

Database/questDB.lua
```diff
@@ -40,6 +40,8 @@ QuestieDB.questKeys = {
     ['parentQuest'] = 25, -- int, the ID of the parent quest that needs to be active for the current one to be available. See also 'childQuests' (field 14)
     ['reputationReward'] = 26, -- table: {{FACTION,VALUE}, ...}, A list of reputation reward for factions
     ['extraObjectives'] = 27, -- table: {{spawnlist, iconFile, text, objectiveIndex (optional), {{dbReferenceType, id}, ...} (optional)},...}, a list of hidden special objectives for a quest. Similar to requiredSourceItems
+    ['requiredSpell'] = 28, -- int: quest is only available if character has this spellID
+    ['requiredSpecialization'] = 29, -- int: quest is only available if character meets the spec requirements. Use QuestieProfessions.specializationKeys for having a spec, or QuestieProfessions.professionKeys to indicate having the profession with no spec. See QuestieProfessions.lua for more info.
```
Muehe, line 44: This is the correct place to add new fields Questie side.

Muehe:
@Yttrium-tYcLief PR looks good from looking over the diff. :+1:

Only thing I'm wondering about is why add the names and the function returning them without it being used? Not saying it needs to be removed, just wondering if you intended it for something. Also might need translations added if we ever want to use it, but adding a TODO and merge might be fine I guess. :shrug:

BreakBB (approved):
Thanks for adding my requested changes @Yttrium-tYcLief ! This is looking pretty slick now 😏 👍🏻

Muehe:
So I have been looking into it a bit more, still unable to find how cmangos handles specialization requirements, but I have found two quests:

https://www.wowhead.com/wotlk/quest=12803/force-of-nature
https://www.wowhead.com/wotlk/quest=12561/an-issue-of-trust

which require a "spell" being learned:

https://www.wowhead.com/wotlk/spell=54197/cold-weather-flying

Seem to be the only quests with such a requirement best I can tell, but there might be more.

Issue #14, opened 2015-03-18
Cartographer notes can "zombie"


AeroScripts:
Currently, it's possible for cartographer notes to "zombie," which basically means they're on the map, but Questie doesn't realize it, so they'll never get removed.

A quick fix is to remove all notes created by Questie on mod init, but there doesn't seem to be a function for this in Cartographer. This is probably going to need a bit of "black magic"

AeroScripts:
This seems to have been fixed at some point (I could be wrong)

Schaka:
It should be fixed with the current system in place actually, yeah. I'll close it.

AeroScripts:
Reopening this because people have reported it again (or at least something extremely similar)

valmarkuristik:
All quests remain on the map when you've completed the objective, showing no completion on map.

For example you may have 12 Raptor Heads in your inventory, but the map notes for getting them are still there and showing 0/12.

The map also offers you to get quests you are currently already on, bug haven't completed yet.

Logonz:
1.1 is very close, closing to clean up the tracker

PR #4664, opened 2023-02-17
Updated Questie Tracker with full Achievement support.

### LINKED ISSUE #2858: Tracker doesn't refresh text cut off
Tracker resizes but doesn't refresh shortened text until you open/close the tracker.
Looks like it's using "Questie Tracker: 1/25" as a cut off.
Easy to reproduce if you have 0 quest and picks up your first quest.

image

### LINKED ISSUE #3294: Questies' Quest Tracker behaves buggy when questlog is almost full
## Preface
Compared to Blizzards' default quest tracker, the one by Questie leaves out a lot of screen real estate on the right side of it (https://i.imgur.com/VhWPfnK.jpeg), and at the same time extends all the way down even "below" the monitor when you have lots of quests which looks off compared to the rest of Blizzards' UI. So it was reasonable for me to try and reposition and resize it.

## Description:
When trying to move the Quest Tracker further right or down via Ctrl-LDrag it will automatically extend above the minimap, looking completely out of place (https://i.imgur.com/88brZlJ.jpeg)
And it's unmovable from then on unless I do a full Questie reset

I tried it on another char with less quests and it behaved as it was supposed to.

I tried just collapsing all zones but the active one of the Questie Tracker to get it to a reasonable size but upon reloading the entire tracker disappeared. My temporary fix is now to actually use Kaliels' tracker in addition to Questie and the issue won't occur. 

## Speccs:
Questie 6.8.2
TBC Classic 2.5.2.41446
24" 1080p monitor, using 1920x1080 in WoW, Full Screen Mode

## Suggested Solution Idea:
I'd like an option for the quest tracker UI elements to be handled via Blizzards' UI rather than Questie. Or for Questie to just add the functionality of showing tracker quests longer than 5 minutes and that's it. Auto-Collapsing  all quests outside of the zone you're currently at seems like a good idea to me too.

I'm not really a coder so I hope I don't embarrass myself with the suggestion I made.

### LINKED ISSUE #3403: The tracker doesn't resize after the update of a quest whose name exceeds the maximum existing width.
# Bug description

For example the quest that will be updated is the 2nd longest as title in the tracker window. When this quest gets completed, and the tracker adds a "(Completed)" at the end of the title, now this quest name is the longest in the tracker window, and exceeds the current width, and the tracker fails to resize itself.
This happens both when you accept or complete/fail a quest whose name will become longer than any of the existing quests tracked currently.
In the screenshot below the quest The Greatest Race of Hunters was the 2nd longest quest, but became the longest after completing it.
Would a call for resize function whenever a quest is acquired/completed/failed fix this bugged behaviour? I assume this would also fix #3404

# Screenshots
imgur

# Questie & WoW version
6.8.7 TBC

# questie settings
settings1
settings2

### LINKED ISSUE #3404: The tracker fails to resize itself when you track the first quest.
# Bug description

If you track your first quest, the tracker doesn't resize itself.

# Screenshots
imgur

# Questie & WoW version
6.8.7 TBC

# questie settings
settings1
settings2

### LINKED ISSUE #3422: attempted to call a protected function (Questie_TrackedQuests:ClearAllPoints()) during combat lockdown.
Hello, I've been getting this error, intermittently. I think it happens when I leave combat. But I don't know how to reproduce it.

Questie version: 6.9.0
WoW version: 20504

```
Date: 2022-04-23 11:29:39
ID: 1
Error occured in: AddOn: Questie
Count: 1
Message: Note: AddOn Questie attempted to call a protected function (Questie_TrackedQuests:ClearAllPoints()) during combat lockdown.
Debug:
   [string "=[C]"]: ClearAllPoints()
   [string "@Questie\Modules\Tracker\QuestieTracker.lua"]:610: Update()
   [string "@Questie\Modules\Tracker\QuestieTracker.lua"]:1232: Update()
   [string "@Questie\Modules\Quest\QuestieQuest.lua"]:384: val()
   [string "@Questie\Modules\TaskQueue.lua"]:8:
      Questie\Modules\TaskQueue.lua:6
Locals:
None
AddOns:
  Swatter, v2.5.6773 (SwimmingSeadragon)
  AucAdvanced, v2.5.6733 (SwimmingSeadragon)
  AucFilterBasic, v2.5.6759 (SwimmingSeadragon)
  AucStatHistogram, v2.5.6761 (SwimmingSeadragon)
  AucStatiLevel, v2.5.6765 (SwimmingSeadragon)
  AucStatPurchased, v2.5.6762 (SwimmingSeadragon)
  AucStatSimple, v2.5.6763 (SwimmingSeadragon)
  AucStatStdDev, v2.5.6764 (SwimmingSeadragon)
  AucUtilFixAH, v2.5.6766 (SwimmingSeadragon)
  autolootdestroy, v
  BagBrother, v
  Bagnon, v9.2.0
  BagnonConfig, v
  Bartender4, v4.11.1
  BeanCounter, v2.5.6767 (SwimmingSeadragon)
  DBMCore, v7661e31
  DBMStatusBarTimers, v7661e31
  DBMVPVEM, v81c3e41
  Details, v
  DetailsStreamer, v
  DetailsTinyThreat, v
  Enchantrix, v2.5.6768 (SwimmingSeadragon)
  EnchantrixBarker, v2.5.6769 (SwimmingSeadragon)
  GearScoreLite, v0.1 (based on 3x04y05)
  Informant, v2.5.6770 (SwimmingSeadragon)
  LuaDKPCore, v
  OmniCC, v9.2.0
  Pawn, v2.5.37
  Plater, vPlater-v347-BCC
  Quartz, v3.6.9
  Questie, v6.9.0
  Scrap, v9.1.2
  ScrapConfig, v
  ShadowedUnitFrames, vv4.3.2-classic
  SlideBar, v2.5.6771 (SwimmingSeadragon)
  Stubby, v2.5.6772 (SwimmingSeadragon)
  ThreatClassic2, v2.43
  BlizRuntimeLib_enUS v2.5.4.20504 <none>
  (ck=4ad)


```
Thanks

### LINKED ISSUE #3462: Quest Tracker Layout Bugs Out in Combat
Minimize the tracker... then mid-combat... it looks like this:

image

After combat looks like this:

image

Quest Objectives being complete shouldn't force the tracker to show if it's minimized. Tracker should never load different from how it's configured in layout options.

### LINKED ISSUE #3819: Questie Tracker can break when updated in-combat
## Bug description
Questie tracker fails to update if the update is attempted while in combat.
For example: Picking up a new quest with auto-track enabled while in-combat, or completing an objective etc.

The buttons on the tracker for quest item use are secure so their whole anchor hierarchy becomes protected.
Protected frames are prohibited from show/hide/move or re-anchor during combat both explicitly and implicitly (if a re-anchor show/hide etc of a parent frame would cause them to change state)

```lua
1x [ADDON_ACTION_BLOCKED] AddOn 'Questie' tried to call the protected function 'Questie_TrackedQuests:ClearAllPoints()'.
[string "@!BugGrabber\BugGrabber.lua"]:519: in function <!BugGrabber\BugGrabber.lua:519>
[string "=[C]"]: in function `ClearAllPoints'
[string "@Questie\Modules\Tracker\QuestieTracker.lua"]:610: in function `Update'
[string "@Questie\Modules\Tracker\QuestieTracker.lua"]:1231: in function `Update'
[string "@Questie\Modules\Quest\QuestieQuest.lua"]:357: in function `val'
[string "@Questie\Modules\TaskQueue.lua"]:8: in function <Questie\Modules\TaskQueue.lua:6>
```
The easiest solution is probably to defer prohibited updates on the tracker to combat exit (alpha + color changes are fine, show/hide/move/reanchor not). So some kind of CombatQueue.

## Questie & WoW version
Questie Version: v6.13.1
WoW Version: BCC 2.5.4.44400 enGB

### LINKED ISSUE #4073: tracker blank
## Bug description

the quest tracker is enabled and i can see it on the screen but it is blank, no quests even though the header shows 13/25 currently working on. i have deleted the questie folder and all of the questie files in wtf and still have the same problem.

### LINKED ISSUE #4076: Tracker SetPoint Broken
Hi.  I am experiencing an issue where no matter what I set the Tracker SetPoint to in the config, the tracker acts like the setpoint is bottom left.  It grows from bottom up, and quests fill in from bottom to top, left to right.  Is this an issue anyone else is having or it is just me?

### LINKED ISSUE #4114: The tracker tracking the tracking of tasks disappeared
It disappears and may appear if it reset its position, it will appear in the center of the screen and can be moved, but it will disappear again after you re -end the game.

### LINKED ISSUE #4199: Bug when Picking up quest while in combat
1x [ADDON_ACTION_BLOCKED] AddOn 'Questie' tried to call the protected function 'Questie_BaseFrame:SetWidth()'.
[string "@!BugGrabber\BugGrabber.lua"]:480: in function <!BugGrabber\BugGrabber.lua:480>
[string "=[C]"]: in function `SetWidth'
[string "@Questie\Modules\Tracker\TrackerBaseFrame.lua"]:313: in function `UpdateWidth'
[string "@Questie\Modules\Tracker\QuestieTracker.lua"]:1027: in function `Update'
[string "@Questie\Modules\Quest\QuestieQuest.lua"]:374: in function `val'
[string "@Questie\Modules\TaskQueue.lua"]:8: in function <Questie\Modules\TaskQueue.lua:6>

### LINKED ISSUE #4217: Tracker Layout Issue
Still an issue with the tracker...

The icon overlaps the text, and there's a bit of red broken character in the tracker now since the latest patch.

image

### LINKED ISSUE #4227: Cant track quests when you must use quest item
Hi questie cant track quests when you must use quest item.. its so frustrating.. can you resolve that? thx

### LINKED ISSUE #4235: Tracker collapse from top to bottom
I don't know if this was something that should be working this way or if it was removed but the tracker used to be able to collapse from top to bottom in older version. Now when the anchor point is set to bottom right, one it does not collapse from top to bottom and 2 the quest growth still sometimes to grown down word when it should grow upwards. Also there is now an achievement header that is in the way and cannot be removed in the settings from what I can see. Is there a way to try to fix this. I keep my tracker on the bottom right of my screen above my dps meter and when I want to collapse the tracker it it collapses upwards which is super annoying.

### LINKED ISSUE #4272: Tracker not Tracking Quests with Items correctly
## Bug description

Whenever  I accept a quest in Wotlk, which involves Items to use, the Tracker misses this Item and does not correctly track this quests progress. UI Reload solves this and the items now sits in the tracker. Also when not behaving as should: when minimizing and maximizing the background stays solid. When working normaly, it displays solid background for a fraction and goes to translucent as should.
## Questie & WoW version

3.4.0 Wotlk

### LINKED ISSUE #4323: Questie_TrackedAchievements:SetHeight()
Error spams constantly in CoS when a zombie is killed for 'Zombiefest!" (kill 100 zombie in 1min). The tracker will also periodicly track everything 2 or 3 times during combat, but only in CoS. Friend with 7.2.4 doesn't have the issue.
## Bug description

Message: ADDON_ACTION_BLOCKED: Questie tried to call the protected function 'Questie_TrackedAchievements:SetHeight()'.
```
Time: Tue Oct  4 02:40:27 2022
Count: 283
Stack: ADDON_ACTION_BLOCKED: Questie tried to call the protected function 'Questie_TrackedAchievements:SetHeight()'.
[string "@Interface\AddOns\ElvUI\Libraries\Core\Ace\CallbackHandler-1.0\CallbackHandler-1.0.lua"]:29: in function <...Core\Ace\CallbackHandler-1.0\CallbackHandler-1.0.lua:25>
[string "@Interface\AddOns\ElvUI\Libraries\Core\Ace\CallbackHandler-1.0\CallbackHandler-1.0.lua"]:64: in function `Fire'
[string "@Interface\AddOns\ElvUI\Libraries\Core\Ace\AceEvent-3.0\AceEvent-3.0.lua"]:120: in function <...vUI\Libraries\Core\Ace\AceEvent-3.0\AceEvent-3.0.lua:119>
[string "=[C]"]: in function `SetHeight'
[string "@Interface\AddOns\Questie\Modules\Tracker\AchievementTracker.lua"]:67: in function `LoadAchievements'
[string "@Interface\AddOns\Questie\Modules\Tracker\AchievementTracker.lua"]:96: in function <...ddOns\Questie\Modules\Tracker\AchievementTracker.lua:89>
[string "=[C]"]: ?
[string "@Interface\AddOns\ElvUI\Libraries\Core\Ace\CallbackHandler-1.0\CallbackHandler-1.0.lua"]:29: in function <...Core\Ace\CallbackHandler-1.0\CallbackHandler-1.0.lua:25>
[string "@Interface\AddOns\ElvUI\Libraries\Core\Ace\CallbackHandler-1.0\CallbackHandler-1.0.lua"]:64: in function `Fire'
[string "@Interface\AddOns\ElvUI\Libraries\Core\Ace\AceEvent-3.0\AceEvent-3.0.lua"]:120: in function <...vUI\Libraries\Core\Ace\AceEvent-3.0\AceEvent-3.0.lua:119>

Locals: <none>
```

## Questie & WoW version
7.3 Questie
3.4.0.45854 WoW

### LINKED ISSUE #4327: Quest Tracker duplicating quests
Questie v7.3.0 Wrath Classic

In Culling of Stratholme dungeon, the questie tracker visually duplicates the tracked quests repeatedly until they fill the column to the edges of the screen. 

This happens throughout the dungeon, especially when killing the weak zombies in the streets, whilst accompanying the Arthas NPC. As you continue to stay in combat, the quests just keep duplicating. When out of combat, you can disable/enable the tracker to reset the visual bug, but the duplication continues the next time you go into combat.

### LINKED ISSUE #4416: Tracker not updating in real time..
Over the last 2 days I've encountered the same issue on nearly every quest, where the tracker-text on screen that shows quest progress won't update along with your actual progress. The maps and everything else work great and update just fine, and when you open your quest log it's of course completed correctly - but when you kill a monster or a mob for a quest it doesn't update the on screen tracker. So if you don't realize you've killed enough mobs already or picked up enough of the required quest items it'll still show "8/10" or "so and so killed: 0/1" and I'll stay running around trying to finish the quest only to realize 5-10 minutes later after pressing "L" that the quest is already complete. I actively noticed this happening over 20 times yesterday on 10/15/2022.

### LINKED ISSUE #4455: Lua error when receiving an achievement
Message: [ADDON_ACTION_BLOCKED] AddOn 'Questie' tried to call the protected function 'Questie_TrackedAchievements:SetHeight()'.
Time: Sat Oct 29 17:32:43 2022
Count: 1
Stack: [ADDON_ACTION_BLOCKED] AddOn 'Questie' tried to call the protected function 'Questie_TrackedAchievements:SetHeight()'.

Locals: <none>
Questie versions 7.4.2, just received an achievement.

### LINKED ISSUE #4473: Quest spam in quest tracker
## Bug description
Quests multiply in quest tracker while in Culling of Strat instance. If you disable/enable quests tracking it fixes for some time but then happens again.

## Questie & WoW version
wotlk classic questie 7.4.3

### LINKED ISSUE #4513: No side quest display
Questie version 7.4.5

## Bug description
 I have no side quest display on the main screen.  It was there when I logged in about an hour ago and now it is gone and I can't get it back.  I have uninstalled and reinstalled it several times to no avail.  No patch has came out from blizzard in the last several days either.  I haven't added any new addon's.

## Questie & WoW version
This is occuring in classic Wotlk

### LINKED ISSUE #4525: All Quests show as Completed
Latest build in Wrath.

I logged in and everything shows as completed in tracker.

image

That's not accurate. (=

image

### LINKED ISSUE #4526: Tracker Bug
## Bug description
The tracker will not load. Quest progression causes this to pop up:
Message: ...ce\AddOns\Questie\Modules\Tracker\QuestieTracker.lua:940: attempt to concatenate a nil value
Time: Sat Nov 19 20:42:56 2022
Count: 2
Stack: ...ce\AddOns\Questie\Modules\Tracker\QuestieTracker.lua:940: attempt to concatenate a nil value

Locals: 

## Questie & WoW version
<!--
Questie: v7.4.5
WoW: 3.4.0.46368 WotLK

### LINKED ISSUE #4635: Lua Error on Quest Pickup when in Combat
Used to get an error for the tracker, this I think is a new issue.

1x [ADDON_ACTION_BLOCKED] AddOn 'Questie' tried to call the protected function 'Button:Show()'.
[string "@!BugGrabber/BugGrabber.lua"]:480: in function <!BugGrabber/BugGrabber.lua:480>
[string "=[C]"]: in function `Show'
[string "@Questie/Modules/Tracker/QuestieTracker.lua"]:831: in function `Update'
[string "@Questie/Modules/Quest/QuestieQuest.lua"]:382: in function `val'
[string "@Questie/Modules/TaskQueue.lua"]:8: in function <Questie/Modules/TaskQueue.lua:6>

### LINKED ISSUE #4637: Bug with So Far, So Bad Objective in Tracker
image

image

https://www.wowhead.com/wotlk/quest=12669/so-far-so-bad

Top objective won't complete in Questie, but it's complete.

### LINKED ISSUE #4696: Tracker Anchor Issue
I think I reported this before... can't remember. @Dyaxler 

When you have a full tracker, the tracker can move from the anchor point you set.

After it moves, it then impacts the anchor point on other toons.

Here's the tracker, with a lot of quests. Blue line is roughly the anchor point.

image

Then if you expand one more quest... it moves the tracker anchor up. 

image

Then if you log out... that new "top" is where the anchor will be set for all your toons.

### LINKED ISSUE #4762: Quest tracker still does not collapse downward when anchor point is set to bottom right
When the anchor point is set to bottom right the quests do not grow from bottom to top and it does not collapse downward like you would expect if something is anchored to the bottom. The paper doll for broken gear also does not show on the left side of the tracker as you would expect if it is anchored to the right. I reported this bug after the launch of Wrath and was told it was being worked. There have been many updates but still this is not fixed. What is the point of setting anchor points if it is not going to effect how different things work for the quest tracker? This functionality was working fine in Classic and TBC, it broke with Wrath and would be greatly appreciated if it was fixed seeing how we are in Phase 2 already.

Dyaxler:
I've revamped the Achievement feature. Achievements themselves aren't a complicated feature that warrants a separate module, so I've removed the separate Achievement module and combined the code into the Questie Tracker. This is far more efficient and easier to control the Trackers behavior when it comes to formatting, dimensions and overall complexity from a code level. It also allowed me to easily implement formatting features, highlighting, right click menu etc. while using the same code base we use for Quests.

I’ve also updated the Questie Tracker and have hopefully resolved some long-standing issues. I’m very happy with it at this point.

Achievement Features:

- Achievements are now tracked like Quests and placed in their own zone called “Achievements”.
- Clicking on an Achievement will open the Achievement log to the clicked Achievement.
- If an Achievement Objective is another Achievement then it too will open that Achievement in the Achievement log.
- The mouse over highlight feature of the Tracker will highlight Sub-Achievements embedded in Parent Achievements individually. If an Objective isn’t another Achievement then the entire entry (Achievement and Objective text) in the Tracker will highlight.
- You can link Achievements like quests directly from the Tracker into Chat. (Chat entry visible - Shift + Left Click)
- You can untrack an Achievement like quests directly from the Tracker. (Shift + Left Click)
- Achievements now have a “Right Click” menu like quests. (Separate menus for Parent Achievement and if applicable menus for Achievements listed within Achievements)
- Clicking the Achievements Zone header will minimize the entire zone. Shift + Left clicking the Achievements Zone header will minimize all Achievements individually within the zone.
- Each Achievement also has its own minimize button, just like quests. (And via the "Right Click" menu)
- The Questie Tracker is limited to tracking only 10 Achievements at a time, just like the default Blizzard Achievement tracking feature. I don’t foresee an overwhelming need to track more than 10 at a time. I can’t imagine why you would. Some Achievements have quite a few Objectives/Achievements embedded in them and quite a few have verbose descriptions. Tracking 10 of these will quickly fill the player’s screen, especially if the player is tracking several quests.
- Since Questie is technically a "Quest" tracker it didn't make sense to add an additional tag to the "Active Quests Header" to indicate how many Achievements one is tracking. So instead, the Achievement "Zone Header" will have this info handy. (Look at the Sorting Feature screen shot below)
- There is a new option in Questie Options --> Tracker, "Hide Completed Achieve Objectives".

Tracker Features:

- Along with all the new Achievement features, the entire Tracker codebase has been updated and refactored. Most of the formatting centered around the overall Width & Height of each line along with the overall Width & Height of the Tracker itself has been reengineered from scratch. All the other little features and functions have also been optimized for better performance.
- Enable/Disable Questie Tracker is now saved byCharacter vs byGlobal
- Shift-Clicking a Zone header will minimize all Quests in that Zone. You can still pick and choose which quests you'd like to maximize. Shift-Clicking a Zone header a second time will maximize any Quests that are minimized for that Zone.
- Added a new Quest "Tag". Instead of seeing the traditional "Quest Complete!" tag, a player will now see "Completion Text" instead. This only applies to Quests/Achievements with SINGLE objectives. The "Completion Text" is pulled directly from the Quest Log and this kind of text isn't always present for multiple Objective Quests/Achievements. So far, an appropriate "do this/go here to finish this quest" description is always present for these types of Quests/Achievements. Beyond that, we are NOT going to change Questie into a "Guide" type addon so don't get any ideas!
- The Quest Objective Sorting feature has been given a facelift. Since Achievements always live in their own "Zone", using the Sorting Feature looked off when all the Quests were stacked in the Tracker. So now, if you're using a Sorting setting OTHER than "By Zone", Quests will be placed in a new custom "Zone" label based on the Sorting setting. (See screen shot below)
- Implemented custom “wrapping” for long Quest & Achievement Objectives (Zones, Quest Titles are NOT wrapped) to keep the tracker slimmer. This is limited to single Objectives. If there is more than one Objective, then wrapping will not occur. Single Objectives with no "Need/Progress" tags or blocks of text that require wrapping are automatically wrapped with a minimum width setting to keep the blocks from looking malformed due to an overly narrow list of Achievements/Quests. These wrapped blocks aren't limited to a set number of lines so they will shrink and grow if the Trackers width changes.
- Quest & Achievement Objectives with no "Need/Progress" counts or Objectives with a single "Need/Progress" tag have been removed. Progress text coloring is still used so if the text is in Red then it's incomplete (depending on which color scheme you're using of course). 
- Fixed minimizing the Questie Tracker. An empty frame no longer lingers.
- You can now hide/show the Questie Tracker “Sizer” in Questie Options --> Tracker.
- Fixed tracking quests/achievements when only one zone is present in the tracker while minimized caused the zone header to overlap the Questie Tracker Header.
- Changed how Questie Options --> Tracker - Tracked Quests Grow to "Tracker Growth Direction". I’ve updated all the available options so Quests can now grow either "Down & Right", Down & Left, Up & Right, and Up & Left. Growing up is good when the Questie tracker is near the lower half of your screen and vis versa. The default is now Down. It was necessary to revamp this option since the feature I created in Classic WoW that automatically calculated where the tracker was placed on the player’s screen and changing the direction quests were inserted (Up or Down) and moving the Tracker Header dynamically is no longer in the code base. I’m actually fine with this since it added a layer of complexity that wasn’t necessary.
- Questie Options -> Tracker: Auto Move Header. This works in conjunction with Tracked Quests Grow and is only available when it’s set to Up. Setting Up or Down auto moves the Tracker Header to either the top or bottom to give the player a visual indication of which direction quests are being populated in the tracker. If you ALWAYS want the Tracker Header up top, then turn this option off.
- The Questie "Sizer" button now lives in two different corners. For Up orientations, the Sizer is in the upper right corner. For Down orientations the Sizer is in the lower right corner unless the aforementioned setting "Auto Move Header" is disabled or the Header itself is disabled. Basically, the Sizer lives on the opposite side of the Trackers Header and is meant to help remind the player which direction Quests/Achievements are being populated and which direction the Tracker is growing.
- Fixed several Questie Options settings to properly update the tracker. There were a few missing forced updates that needed to happen after the setting was changed.
- When the Questie Tracker isn't tracking any Quests and Achievements then the Tracker is fully hidden. This also includes the "Active Quests Header". A new option was implemented called "Always Show Tracker". Even if the player disables "Enable Active Quests Header" then you will at least see the Header in any of the Trackers various states and it will remember the players preference to have the option "Enable Active Quests Header" in a disabled state. In this option state, the Header will hide itself once again when a Quest or an Achievement is tracked. "Always Show Tracker" is enabled by default.
- There is a new option in Questie Options --> Tracker, "Hide Completed Quest Objectives".
- Quest Item Buttons has been given a new feature. Some quests have both a Primary and a Secondary item that is needed to complete a Quest/Objective. When these items are present, they will appear on the Tracker next to the Quests Title as before. NOTE: Only "usable" items appear on the Tracker. If an Un-usable item appears, please let us know.

- Added a Tracker Height Ratio option in the Questie Config --> Tracker options. This slider will allow you to set the Trackers height.
- And finally! The Questie Tracker now has an adjustable Height. No more growing beyond your screens real-estate! The Tracker will only occupy your available vertical screen height based on the Tracker Height Ratio setting. Where does the rest of it go you ask? SCROLLING. A hidden scroll frame will be enabled automatically, and you can use your Mouse Wheel to scroll through your list of Quests and Achievements when it grows beyond the Tracker Height Ratio setting.

Questie Tracker

Questie's new sorting feature:
Questie Sorting

Issue references

Fixes #4762
Fixes #4696
Fixes #4664
Fixes #4637
Fixes #4635
Fixes #4526
Fixes #4525
Fixes #4513
Fixes #4473
Fixes #4455
Fixes #4416
Fixes #4327
Fixes #4323
Fixes #4272
Fixes #4235
Fixes #4227
Fixes #4217
Fixes #4199
Fixes #4114
Fixes #4076
Fixes #4073
Fixes #3819
Fixes #3462
Fixes #3422
Fixes #3404
Fixes #3403
Fixes #3294
Fixes #2858


Modules/Migration.lua
```diff
@@ -138,9 +138,6 @@ local migrationFunctions = {
         Questie.db[Questie.db.global.questieTLoc].TrackerLocation = nil
     end,
     [14] = function()
```
BreakBB, line 140: Lets just remove this step entirely 😁 

```suggestion
```

Modules/Migration.lua
```diff
@@ -138,9 +138,6 @@ local migrationFunctions = {
...
-        if Questie.db.char.isAchievementsExpanded == nil then
-            Questie.db.char.isAchievementsExpanded = true
-        end
     end,
```
BreakBB, line 141:
```suggestion
```

Modules/Options/TrackerTab/QuestieOptionsTracker.lua
```diff
@@ -53,6 +53,9 @@ function QuestieOptions.tabs.tracker:Initialize()
                         Questie.db.char.AutoUntrackedQuests = {}
                     end
                     QuestieTracker:Update()
```
BreakBB, line 55: Do we need this call if we call with a timer? Also why do we need a timer in the first place?

There are multiple other timers below.

Modules/Quest/QuestieQuest.lua
```diff
@@ -379,7 +379,12 @@ function QuestieQuest:AcceptQuest(questId)
                 Questie:SendMessage("QC_ID_BROADCAST_QUEST_UPDATE", questId)
             end,
             function() QuestieQuest:PopulateObjectiveNotes(quest) end,
-            function() QuestieTracker:Update() end,
+            function() QuestieTracker:Update()
```
BreakBB, line 382:
```suggestion
            function()
                QuestieTracker:Update()
```

BreakBB, line 382:
```suggestion
            function()
                QuestieTracker:Update()
```

Modules/Quest/QuestieQuest.lua
```diff
@@ -379,7 +379,12 @@ function QuestieQuest:AcceptQuest(questId)
...
+                -- This is necessary to call it again to update the trackers formatting
```
BreakBB, line 383: If this is true for all the other double calls below, can't the formatting be split away from `QuestieTracker:Update()`?

Modules/Tracker/LinePool.lua
```diff
@@ -218,128 +237,6 @@ function LinePool.Initialize(trackedQuestsFrame, UntrackQuest, TrackerUpdate)
     end
 end
 
----@param trackedAchievementsFrame Frame
-function LinePool.InitializeAchievementLines(trackedAchievementsFrame, OnAchievementClick, AchievementUpdate)
```
BreakBB, line 222: I know see that it wasn't the best idea to separate the Achievement tracking from the Tracker the way I did it 😅 Lots of duplicated code 🙈

Modules/Tracker/QuestieTracker.lua
```diff
@@ -877,14 +858,47 @@ function QuestieTracker:Update()
...
-                    button.itemID = quest.sourceItemId
+                    if quest.sourceItemId then
+                        button.itemID = quest.sourceItemId
+
+                    elseif type(quest.requiredSourceItems) == "table" and #quest.requiredSourceItems == 1 then
+                        button.itemID = quest.requiredSourceItems[1]
+                    end
                     button.fontSize = fontSizeCompare
                     button.line = line
-                    QuestieCombatQueue:Queue(_UpdateQuestItem, button, quest)
```
BreakBB, line 887: Are you sure it is save to remove the QuestieCombatQueue call? 🤔

Modules/Tracker/QuestieTracker.lua
```diff
@@ -954,13 +970,19 @@ function QuestieTracker:Update()
...
                     -- Tags quest as either complete or failed so as to always have at least one objective.
-                    -- (TODO: change tags to reflect NPC to turn a quest into or in the case of a failure
-                    -- which NPC to obtain the quest from again...)
+
+                    -- TODO: change tags to reflect NPC to turn a quest into or in the case of a failure
+                    -- which NPC to obtain the quest from again...
+
+                    -- TODO: Why? The Quest Tracker only tracks a quests progress, nothing more. Besides,
+                    -- Questie already informs the user where to turn in the quest. In the case of a failure,
+                    -- the quest is marked on the map as available again.
```
BreakBB, line 979: I agree to this. No need to show the turn in/pick up NPC in the Tracker 👍🏻

Modules/Tracker/QuestieTracker.lua
```diff
@@ -989,13 +1011,244 @@ function QuestieTracker:Update()
...
+                if line.label:GetHeight() > Questie.db.global.trackerFontSizeObjective + .01 then
+                    line:SetVerticalPadding(Questie.db.global.trackerFontSizeObjective + Questie.db.global.trackerQuestPadding)
+                else
+                    line:SetVerticalPadding(Questie.db.global.trackerQuestPadding)
+                end
+            end
+        end
+    end
+
+    -- TODO: Is this needed?
```
BreakBB, line 1248:
```suggestion
```

Modules/Tracker/QuestieTracker.lua
```diff
@@ -989,13 +1011,244 @@ function QuestieTracker:Update()
...
+
+    -- TODO: Yes. This is a safety check to make sure we have the last line selected so the tracker can be properly trimmed.
```
BreakBB, line 1250:
```suggestion
    -- This is a safety check to make sure we have the last line selected so the tracker can be properly trimmed.
```

Modules/Tracker/QuestieTracker.lua
```diff
@@ -989,13 +1011,244 @@ function QuestieTracker:Update()
...
+                end
+
+                if line.label:GetHeight() > Questie.db.global.trackerFontSizeObjective + .01 then
+                    line:SetVerticalPadding(Questie.db.global.trackerFontSizeObjective + Questie.db.global.trackerQuestPadding)
+                else
+                    line:SetVerticalPadding(Questie.db.global.trackerQuestPadding)
+                end
+            end
+        end
+    end
```
BreakBB, line 1246: Could this be moved to a new function? The `:Update` function is already something like 300 lines long 🙈

Modules/Options/TrackerTab/QuestieOptionsTracker.lua
```diff
@@ -73,11 +76,14 @@ function QuestieOptions.tabs.tracker:Initialize()
                 order = 1.3,
                 width = 1.0,
                 name = function() return l10n('Min Complete Quests'); end,
-                desc = function() return l10n('When this is checked, completed quests will automatically minimize.'); end,
+                desc = function() return l10n('When this is checked, completed quests will automatically minimize.\n\nNOTE: If this option is currently enabled, disabling this option will maximize ALL minimized quests currently in your Questie Tracker.'); end,
```
BreakBB, line 79: When chaning the content of a `l10n` call you also need to change [the translation](https://github.com/Questie/Questie/tree/master/Localization/Translations)

Modules/Options/TrackerTab/QuestieOptionsTracker.lua
```diff
@@ -133,28 +142,34 @@ function QuestieOptions.tabs.tracker:Initialize()
                 order = 1.7,
                 width = 1.0,
                 name = function() return l10n('Enable Header'); end,
-                desc = function() return l10n('When this is checked, the Questie Tracker Header will become visible and the total number of quests you have in your log will be shown.'); end,
+                desc = function() return l10n('When this is checked, the Active Quests Header will become visible and the total number of quests you have in your log will be shown.'); end,
```
BreakBB, line 145: The [translation](https://github.com/Questie/Questie/tree/master/Localization/Translations) needs to be fixed

Modules/Options/TrackerTab/QuestieOptionsTracker.lua
```diff
@@ -133,28 +142,34 @@ function QuestieOptions.tabs.tracker:Initialize()
...
+                    end)
                 end
             },
             autoMoveHeader = {
                 type = "toggle",
                 order = 1.8,
                 width = 1.0,
                 name = function() return l10n('Auto Move Header'); end,
-                desc = function() return l10n("When this is checked, the Questie Tracker Header will automatically move to the bottom of the Questie Tracker while the Tracker is placed in the bottom area of your screen. Otherwise, it will remain at the top always.\n\nNOTE: This setting only works while the 'Tracker SetPoint' is using the default 'Auto' mode."); end,
+                desc = function() return l10n("When this is checked, the Active Quests Header will automatically move to the top or bottom of the Questie Tracker depending on which 'Tracked Quests Grow' setting is used.\n\nNOTE: This setting only works while the 'Tracked Quests Grow' setting is set to 'Up'."); end,
```
BreakBB, line 161: The [translation](https://github.com/Questie/Questie/tree/master/Localization/Translations) needs to be fixed

Modules/Options/TrackerTab/QuestieOptionsTracker.lua
```diff
@@ -335,6 +349,22 @@ function QuestieOptions.tabs.tracker:Initialize()
                     QuestieTracker:Update()
                 end
             },
+            hideSizer = {
+                type = "toggle",
+                order = 2.504,
+                width = 1.0,
+                name = function() return l10n("Hide Tracker Sizer"); end,
+                desc = function() return l10n("When this is the checked, the Questie Tracker Sizer that appears in the bottom right hand corner will be hidden."); end,
```
BreakBB, line 357: We need empty entries for these translations.

Modules/Options/TrackerTab/QuestieOptionsTracker.lua
```diff
@@ -514,22 +545,22 @@ function QuestieOptions.tabs.tracker:Initialize()
...
+                name = function() return l10n('Tracked Quests Grow'); end,
+                desc = function() return l10n("This determines the direction in which the tracker grows when you add or remove quests. This will also move the Active Quests Header to either the top of the Questie Tracker (when using the 'Down' setting) or the bottom of the Questie Tracker (when using the 'Up' setting). \n\nNOTE: You can override the Active Quests Header movement behavior by disabling the 'Auto Move Header' option in Questie Tracker Options to force the Active Quests Header to remain at the top of the Questie Tracker."); end,
```
BreakBB, line 554: The [translation](https://github.com/Questie/Questie/tree/master/Localization/Translations) needs to be fixed

BreakBB (changes requested):
It's good to have you back @Dyaxler ❤️ The Tracker really needs these kind of updates!

Let's get this PR in asap and out in a release 💪🏻

BreakBB:
Oh and about the pipeline errors: New Blizzard API calls need to be added to [.luacheckrc](https://github.com/Questie/Questie/blob/master/.luacheckrc) in order for LuaCheck to kno about these global functions/variables.

Modules/Migration.lua
Dyaxler: Oh right... I left that in there to avoid any errors. :D I can update that and increment the table keys. It'll make it into my next check-in on my PR.

Modules/Options/TrackerTab/QuestieOptionsTracker.lua
Dyaxler: `C_Timer.After(0.1, function() QuestieTracker:Update() end)`

For now, yes. Sadly. If ANY line in the tracker changes, then it's necessary. To put it simply, the first update nukes the Quest/Achievement & Quest Item button pools and clears all the 'set tags' then rebuilds the pools. Once that is complete the second update changes the formatting (width, height, etc.) and switches on or off any buttons that needs to be visible to the player based on each Quests/Achievements status. In the past I've tried separating this process with very little success. It creates FAR too many bugs and edge cases. In essence, nuking the entire tracker and rebuilding it from scratch seems to be the most efficient method. The QuestieTracker:Update() function is long and seems overly complicated but from an execution standpoint it's slim and fast. I've only been inserting `C_Timer.After(0.1, function() QuestieTracker:Update() end)` in places where its absolution needed. Even though it's called often, if the tracker is minimized and/or hidden. then most of it isn't even executed. The more a player tracks Quests and Achievements the more is executed. Make sense?

Modules/Options/TrackerTab/QuestieOptionsTracker.lua
Dyaxler: Copy that... I'll do my best. :) I forgot about updating all the translations.

Modules/Migration.lua
BreakBB: Not sure if we should increment the table keys, cause that might cause settings changes for users. Just removing it should be fine 👍🏻

Modules/Options/TrackerTab/QuestieOptionsTracker.lua
BreakBB: Yes, this absolutely makes sense. I personally just get "a feeling" when I see timers in our code 😅 I am absolutely on your side that the code is really complex and it's dangerous to do major refactorings/code separations on it (... I failed in the past 🙈 )

Modules/Migration.lua
Dyaxler: I removed #14 and renumbered them. I looked over that stack and it's not going to break anything. Matter of fact, some of the Questie.db keys have changed so I think I should add the new ones in? I honestly don't think it's needed since there isn't anything saved to the old keys that needs to be migrated.

It didn't break anything for me when I changed the DB keys and reloaded my UI. It just simply worked.

Modules/Options/QuestieOptionsDefaults.lua
```diff
@@ -68,25 +68,27 @@ function QuestieOptionsDefaults:Load()
             autoTrackQuests = true,
             trackerShowCompleteQuests = true,
             collapseCompletedQuests = false,
+            hideCompletedObjectives = false,
```
BreakBB, line 71: This should be added to a new entry in the `Migration.lua` so existing configurations pick up this new value.

Modules/Options/QuestieOptionsDefaults.lua
```diff
@@ -68,25 +68,27 @@ function QuestieOptionsDefaults:Load()
...
             trackerShowQuestLevel = true,
             showBlizzardQuestTimer = false,
             hookTracking = true,
             trackerHeaderEnabled = true,
-            trackerHeaderAutoMove = false,
+            autoMoveHeader = true,
```
BreakBB, line 76: This should be added to a new entry in the `Migration.lua` so existing configurations pick up this new name.

Modules/Options/QuestieOptionsDefaults.lua
```diff
@@ -68,25 +68,27 @@ function QuestieOptionsDefaults:Load()
...
             stickyDurabilityFrame = false,
             hideTrackerInCombat = false,
             hideTrackerInDungeons = true,
-            trackerFadeMinMaxButtons = true,
+            trackerFadeMinMaxButtons = false,
             trackerFadeQuestItemButtons = false,
             trackerBackdropEnabled = false,
             trackerBorderEnabled = false,
             trackerBackgroundFader = false,
+            sizerHidden = false,
```
BreakBB, line 85: This should be added to a new entry in the `Migration.lua` so existing configurations pick up this new value.

Modules/Options/QuestieOptionsDefaults.lua
```diff
@@ -68,25 +68,27 @@ function QuestieOptionsDefaults:Load()
...
             trackerColorObjectives = 'white',
             trackerSortObjectives = 'byZone',
             trackerbindSetTomTom = 'ctrlleft',
             trackerbindOpenQuestLog = 'left',
             trackerbindUntrack = "shiftleft",
-            trackerSetpoint = "AUTO",
+            trackerSetpoint = "TOPLEFT",
```
BreakBB, line 91: This should be added to a new entry in the `Migration.lua` so existing configurations pick up this new value.

Rock48:
Just started going for Loremaster and stumbled across this - looking forward to having it as a feature

Modules/Quest/QuestieQuest.lua
```diff
@@ -1354,7 +1455,7 @@ do
                                 ---@type Quest
                                 local quest = QuestieDB:GetQuest(questId)
                                 if (not quest.tagInfoWasCached) then
-                                    Questie:Debug(Questie.DEBUG_SPAM, "Caching tag info for quest", questId)
+                                    --Questie:Debug(Questie.DEBUG_SPAM, "Caching tag info for quest", questId)
```
BreakBB, line 1458:
```suggestion
                                    Questie:Debug(Questie.DEBUG_SPAM, "Caching tag info for quest", questId)
```

BreakBB (commented):
From my side the PR is in really really good shape. Code wise and also from the testing I did in the last days.

One question I still have is: Would it be possible/feasible to split `QuestieTracker:Update` into smaller functions? I mean that function is HUGE and from my side an absolute horror to understand 😄

Rock48:
Any update on when we can expect this feature?

Modules/Quest/QuestEventHandler.lua
```diff
@@ -1,9 +1,13 @@
 ---@class QuestEventHandler
 local QuestEventHandler = QuestieLoader:CreateModule("QuestEventHandler")
+---@class QuestEventHandlerPrivate
 local _QuestEventHandler = QuestEventHandler.private
+
 local _QuestLogUpdateQueue = {} -- Helper module
 local questLogUpdateQueue = {} -- The actual queue
 
+---@type QuestEventHandlerPrivate
+QuestEventHandler.private = QuestEventHandler.private or {}
```
BreakBB, line 10: This should happen before `local _QuestEventHandler = QuestEventHandler.private` in line 4, shouldn't it?

Modules/Tracker/LinePool.lua
```diff
@@ -68,35 +79,64 @@ function LinePool.Initialize(trackedQuestsFrame, UntrackQuest, TrackerUpdate)
...
         end
 
         function line:SetObjective(Objective)
             self.Objective = Objective
         end
 
+        line.OnUpdate = function(self, elapsed)
+            timeElapsed = timeElapsed + elapsed
+            if timeElapsed > 1 and self.trackTimedQuest then
+                _QuestieQuestTimers:UpdateTimerFrame()
```
BreakBB, line 113: I think we should move the `UpdateTimerFrame` function to the main module if we need it outside instead of adding `.private` stuff outside of it.

Modules/Tracker/LinePool.lua
```diff
@@ -213,130 +259,242 @@ function LinePool.Initialize(trackedQuestsFrame, UntrackQuest, TrackerUpdate)
...
+
+        if Questie.db.global.trackerFadeQuestItemButtons then
+            btn:SetAlpha(0)
+        end
 
-    local lastFrame
-    for i = 1, achievementPoolSize do
-        local line = CreateFrame("Button", nil, trackedAchievementsFrame)
-        line:SetSize(1, 1)
+        btn.SetItem = function(self, quest, buttonType, size)
```
BreakBB, line 285: Do you think it would be clever to move this function somewhere else as it is quite big?

Modules/Tracker/LinePool.lua
```diff
@@ -503,8 +668,63 @@ _OnClick = function(self, button)
...
+		end
+
+        if (not AchievementFrame:IsShown()) then
+            AchievementFrame_ToggleAchievementFrame()
+            AchievementFrame_SelectAchievement(self.Quest.Id)
+        else
+			if (AchievementFrameAchievements.selection ~= self.Quest.Id) then
+				AchievementFrame_SelectAchievement(self.Quest.Id)
+			end
+        end
```
BreakBB, line 719: What happened here? Looks like someone bought some extra indents 😁

Modules/Tracker/TrackerMenu.lua
```diff
@@ -280,3 +325,120 @@ StaticPopupDialogs["QUESTIE_WOWHEAD_URL"] = {
...
+
+_AddUntrackAchieveOption = function (menu, achieve)
+    tinsert(menu, {text= l10n('Untrack Achievement'), func = function()
+        LibDropDown:CloseDropDownMenus()
+        Questie.db.char.trackedAchievementIds[achieve.Id] = nil
+        QuestieTracker:UpdateAchieveTrackerCache(achieve, achieve.Id, true)
+
+        if (not AchievementFrame) then
+			AchievementFrame_LoadUI()
+		end
```
BreakBB, line 391: Indents
```suggestion
        if (not AchievementFrame) then
	    AchievementFrame_LoadUI()
	end
```

Modules/Tracker/TrackerMenu.lua
```diff
@@ -280,3 +325,120 @@ StaticPopupDialogs["QUESTIE_WOWHEAD_URL"] = {
...
+
+        if (not AchievementFrame:IsShown()) then
+            AchievementFrame_ToggleAchievementFrame()
+            AchievementFrame_SelectAchievement(achieve.Id)
+        else
+			if (AchievementFrameAchievements.selection ~= achieve.Id) then
+				AchievementFrame_SelectAchievement(achieve.Id)
+			end
+        end
```
BreakBB, line 379: Some more indent hickups.

Modules/Tracker/QuestieTracker.lua
```diff
@@ -171,45 +166,95 @@ function QuestieTracker.Initialize()
...
+
+                QuestieCombatQueue:Queue(function()
+                    C_Timer.After(0.5, function()
+                        QuestieTracker:Update()
+                    end)
+                end)
+            end
+        end)
+
+        -- Tracked Achievements aren't always available upon login
```
BreakBB, line 199: This comment looks lost

BreakBB (commented):
Found just some minor indentation issues. which seem to be related to using tabs instead of spaces 🤷🏻 

Most things look really good @Dyaxler I am looking forward to release this 💪🏻

Modules/Tracker/TrackerLinePool.lua
```diff
@@ -0,0 +1,899 @@
...
+        end
+
+        btn.SetItem = function(self, quest, buttonType, size)
+            local validTexture
+
+            for bag = 0 , 4 do
+                for slot = 1, QuestieCompat.GetContainerNumSlots(bag) do
+                    local texture, _, _, _, _, _, _, _, _, itemID = QuestieCompat.GetContainerItemInfo(bag, slot)
+                    -- These type of quest items can never be secondary buttons
+                    if quest.sourceItemId == itemID and QuestieDB:GetItem(itemID).class == 12 and buttonType == "primary" then
```
Logonz, line 328: You should probably use querysingle here rather than getting the whole object (same with the ones below)

Modules/Tracker/QuestieTracker.lua
```diff
@@ -1221,103 +1889,100 @@ function QuestieTracker:RemoveQuest(questId)
...
-    local hexRed = hexTable[8 + math.floor((1 - percent) * 6)]
-    local hexBlue = hexTable[4 + math.floor(percent * 6)]
-
-    return "|cFF"..hexRed..hexGreen..hexBlue..text.."|r"
-end
-
-_RemoveQuestWatch = function(index, isQuestie)
-    Questie:Debug(Questie.DEBUG_DEVELOP, "QuestieTracker: RemoveQuestWatch")
-    if QuestieTracker._disableHooks then
+QuestieTracker.RemoveQuestWatch = function(index, isQuestie)
```
BreakBB, line 1892: This should work, right?

```suggestion
function QuestieTracker.RemoveQuestWatch(index, isQuestie)
```

BreakBB (changes requested):

Modules/Migration.lua
```diff
@@ -156,25 +151,46 @@ local migrationFunctions = {
...
+    end,
+    [21] = function()
+        if Questie.db.global.hideCompletedAchieveObjectives == nil then -- new option
+            Questie.db.global.hideCompletedAchieveObjectives = false -- set default
+        end
+    end,
+    [22] = function()
+        if Questie.db.global.currentHeaderEnabledSetting == nil then -- new option
+            Questie.db.global.currentHeaderEnabledSetting = false -- set default
+        end
```
BreakBB, line 193:
```suggestion
    [20] = function()
        if Questie.db.global.hideCompletedQuestObjectives == nil then -- new option
            Questie.db.global.hideCompletedQuestObjectives = false -- set default
        end
        if Questie.db.global.hideCompletedAchieveObjectives == nil then -- new option
            Questie.db.global.hideCompletedAchieveObjectives = false -- set default
        end
        if Questie.db.global.currentHeaderEnabledSetting == nil then -- new option
            Questie.db.global.currentHeaderEnabledSetting = false -- set default
        end
```

Modules/Options/TrackerTab/QuestieOptionsTracker.lua
```diff
@@ -694,8 +779,12 @@ function QuestieOptions.tabs.tracker:Initialize()
...
                 type = "select",
                 dialogControl = 'LSM30_Font',
-                order = 3.85,
-                values = outlineTable,
+                order = 5.4,
+                values = {
+                            ["None"] = "NONE",
+                            ["Outline"] = "OUTLINE",
+                            ["Monochrome"] = "MONOCHROME"
+                        },
```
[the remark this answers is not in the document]
Dyaxler: WOW! I wonder how long that has been there. Fixed.

Modules/Migration.lua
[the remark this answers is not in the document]
Dyaxler: Done.

Modules/QuestieEventHandler.lua
```diff
@@ -342,9 +365,20 @@ function _EventHandler:ZoneChangedNewArea()
 
     Questie:Debug(Questie.DEBUG_DEVELOP, "[EVENT] ZONE_CHANGED_NEW_AREA")
     if IsInInstance() then
-        wasTrackerExpanded = Questie.db.char.isTrackerExpanded
         QuestieTracker:Collapse()
-    elseif wasTrackerExpanded then
+        -- By my tests it takes a full 6-7 seconds for the instance to load. There are a lot of
+        -- backend Questie updates that occur when a player zones into an instance. This is
+        -- necessary to get the tracker back into it's "normal" state after all the updates.
```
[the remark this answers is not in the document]
Dyaxler: I think the *.private stuff is a deprecated method left over from the earlier days of Classic. We're using the Module method now to keep everything local. I removed all that on the "Questie Tracker" side and refactored everything calling those old methods. Once my PR is done, I can help refactor all those. The QuestieEventHandler is technically outside the Tracker module, and I only change files when absolutely needed to keep the Tracker functional.

Modules/Tracker/TrackerMenu.lua
```diff
@@ -280,3 +281,175 @@ StaticPopupDialogs["QUESTIE_WOWHEAD_URL"] = {
...
```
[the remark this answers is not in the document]
Dyaxler: Oh boy... yeah, I didn't have Render Whitespace turned on nor was "Format On Paste" or "Format On Save" enabled. :D In VStudio tabs and spaces are rendered equally so I didn't see them. Got it!

Modules/Tracker/TrackerUtils.lua
```diff
@@ -240,77 +258,96 @@ function TrackerUtils:FlashFinisher(quest) -- really terrible animation copypast
...
-        return "RightButton" == button and IsControlKeyDown()
-    end,
-    ['altleft'] = function(button)
-        return "LeftButton" == button and IsAltKeyDown()
-    end,
-    ['altright'] = function(button)
-        return "RightButton" == button and IsAltKeyDown()
-    end,
-    ['disabled'] = function() return false; end,
-}
```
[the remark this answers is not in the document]
Dyaxler: Good point! I was focused on refracting this code and not thinking about the fact that this table would get rebuilt each time. :D I moved the bindTruthTable up to the top of the TrackerUtils "local section" so the table is built once.

I hate it when local vars and tables are built in the middle of files. :D I like them all in one place.

Modules/Tracker/QuestieTracker.lua
```diff
@@ -171,277 +144,136 @@ function QuestieTracker.Initialize()
...
+                    -- Remove quest from the Blizzard Quest Watch
+                    if questIndex then
+                        RemoveQuestWatch(questIndex, true)
 
-                        valid = IsItemInRange(self.itemName, "target")
+                        if Questie.IsWotlk then
+                            WatchFrame_Update()
+                        else
+                            QuestWatch_Update()
+                        end
```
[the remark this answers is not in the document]
Dyaxler: WatchFrame_Update & QuestWatch_Update are only called once in QuestieTracker. It only makes sense to create a local Var for this kind of method when the locally created Var is needed to be called more than once.

Modules/Tracker/QuestieTracker.lua
```diff
@@ -513,242 +373,73 @@ function QuestieTracker:Toggle()
...
-        end
+    if trackerHeaderFrame and trackerHeaderFrame.trackedQuests and (not Questie.db.char.isTrackerExpanded) then
+        trackerHeaderFrame.trackedQuests:Click()
+        QuestieTracker:Update()
     end
-
-    return sortedQuestIds
 end
 
 function QuestieTracker:Update()
```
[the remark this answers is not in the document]
Dyaxler: I totally agree and do plan to comb through this function looking for ways to create more TrackerUtils functions. But that's later. Most of the formatting code in QuestieTracker:Update() has been totally re-written and refactored.

Modules/Tracker/QuestieTracker.lua
Dyaxler: Opps... missed one! I refactored all the other ones. :D

Modules/Migration.lua
Dyaxler: Fixed!

BreakBB (approved):

BreakBB (approved):

PR #4667, opened 2023-02-21
pfQuest/ClassicCodex icons

### LINKED ISSUE #4659: Size / Color / Shape of Map Icons / Quest objectives etc
## Issues:

part of the problem I have is the sword just takes up too much space, even when you shrink it down, then all the other icons, are all brown-ish blending into the background of the map, so seeing them is always a issue.  classic codex was just small dots u could resize with colors like purple, yellow, pink etc so they stand out / easy to see.... you dont get the ' kill quest ' and 'loot' icons but it was worth it to me.  but in theory you could color code the dots with a (key legend) that explained what color was for what (posted in discord): https://discord.com/channels/263036731165638656/991572107773353994/1074496309278560286 

There is a downside to trying to do a "key legend" and certain colors having designations.  For example:  red dots = kill quests, green dots = $ / loot quest.  The issue of doing that would be that when you pick up lots of quests, there would be too many the same color.  The beauty of classic codex is that they were all colored different according to being on differ quests, so you could quickly see at a glance how many dots represented the quest you are on.  you would even shift left click those color dots to make them disappear if you wanted to focus on other dots / quests.  

In conclusion, If I had my choice, I would only use classic codex, just the way it was, regardless of losing the sword / money back icons etc.  It just looked a lot cleaner, I could see the dots better than questie icons.  But classic codex was not kept updated, so Im hoping that Questie can be the best of both and have maybe both options for those that like the swords / bag icons and those that like the dots.  

Thanks in advance for anything you are willing to do towards this aim.  

p.s. please put a #donations channel in the discord

Muehe:
## Issue references

Fixes #4659 

## Proposed changes

* Add pfQuest icons and a toggle in the general settings tab to turn them on and off.
* Improves icon coloring.

For more info see commit messages.

## TODO

* Fix nameplate icons.
* If anybody knows how to re-color icons once they had a color assigned let me know, right now quests/objectives only get a new color on `/reload`. They should choose a new one on SmoothReset and when clicking on an objective icon as well.
* We should add a "style selection" frame with screenshots to the settings, and replace the current pfQuest toggle with a setting for each individual icon type.
* Somebody please take a look at the color ordering, I'm genetically indisposed, but I think some kinda similar colors still follow each other.

## Screenshots
Screenshot_20230221_060723
Screenshot_20230221_060923
Screenshot_20230221_061203


Modules/FramePool/QuestieFrame.lua
```diff
@@ -254,7 +254,7 @@ function _Qframe:UpdateTexture(texture)
     end
 
     self.texture:SetTexture(texture)
-    self.data.Icon = texture;
+    --self.data.Icon = texture;
```
Muehe, line 257: Not so sure what was going on here, from my reading of the code `texture` should always be `self.data.Icon` before the commit changing this anyway, so not sure why it was overwritten here...

Questie.lua
```diff
@@ -128,5 +128,32 @@ function Questie:Debug(...)
     end
 end
 
+Questie.icons = {
```
Muehe, line 131: Not sure if this is the best place for this stuff, I mainly chose it because I didn't want to be loading modules all over the place. :shrug:

Modules/Options/GeneralTab/QuestieOptionsGeneral.lua
```diff
@@ -245,6 +247,42 @@ function QuestieOptions.tabs.general:Initialize()
                             QuestieQuest:SmoothReset()
                         end,
                     },
+                    usePfQuestIcons = {
```
Muehe, line 250: This toggle should ideally be replaced with a drop-down selection for each icon type, but it is good enough for now and can be used as a basis for a "style" selection frame later.

PR #4718, opened 2023-03-23
Fix quest The Tome of Nobility (1661)

gouletr:
Fix quest [The Tome of Nobility (1661)](https://www.wowhead.com/classic/quest=1661/the-tome-of-nobility) not showing up on the NPC in Classic Era. Quest was marked as hidden by Questie. I investigated and found that it was blacklisted. This is a mistake because this quest exists in Classic Era, and also exists in TBC and Wrath.

### DIFF
--- a/Database/Corrections/QuestieQuestBlacklist.lua
+++ b/Database/Corrections/QuestieQuestBlacklist.lua
@@ -1041,7 +1041,7 @@ function QuestieQuestBlacklist:Load()
         [708] = true,
         [909] = true,
         [1288] = true,
-        [1661] = true,
+        [1661] = QuestieCorrections.WOTLK_ONLY,
         [3366] = true,
         [3381] = true,
         [5627] = true,
--- a/Database/Corrections/QuestieQuestBlacklist.lua
+++ b/Database/Corrections/QuestieQuestBlacklist.lua
@@ -1041,7 +1041,7 @@ function QuestieQuestBlacklist:Load()
         [708] = true,
         [909] = true,
         [1288] = true,
-        [1661] = QuestieCorrections.WOTLK_ONLY,
+        [1661] = QuestieCorrections.TBC_AND_WOTLK,
         [3366] = true,
         [3381] = true,
         [5627] = true,


BreakBB:
Hey @gouletr this questId should stay blacklisted for TBC and Wotlk as this one was replaced by a new quest (same name, different ID).

In order to do this update the blacklisting to be `= QuestieCorrections.WOTLK_ONLY`, instead of `= true`.

BreakBB:
The quest was initially called "OLD The Tome of Nobility" with the start of TBC. There are many other quests (see fcec1cfb34ecec3c5119c6195b8232ccf4d22ffe ) which name also starts with "OLD" in the TBC database and are not available in the game after the TBC release.

BreakBB:
If you want to verify this: Yes.

BUT I guess you verified this quest exists on Era and I suggest we just unhide it on Era 👍🏻

gouletr:
> If you want to verify this: Yes.
> 
> BUT I guess you verified this quest exists on Era and I suggest we just unhide it on Era 👍🏻

You say this was removed in TBC, but you suggest I set the value to `QuestieCorrections.WOTLK_ONLY`. Shouldn't it be `QuestieCorrections.TBC_ONLY` ?

gouletr:
Ok sounds good, I pushed the fix.

BreakBB:
> We're just storing booleans, can't we simply do the following? Would it not resolve as true?
> 
> > [1661] = QuestieCorrections.TBC_ONLY or QuestieCorrections.WOTLK_ONLY,

The flags are numbers:

```lua
QuestieCorrections.TBC_ONLY = 1
QuestieCorrections.CLASSIC_ONLY = 2
QuestieCorrections.WOTLK_ONLY = 3
```

The issue is this function https://github.com/Questie/Questie/blob/4e36bf79d966b614d871f7c3e3f78a8e896b202b/Database/Corrections/QuestieCorrections.lua#L83-L108

We would need to add logic for Wotlk to also keep TBC flagged quests hidden. But then the flags are no longer `TBC_ONLY ` but more like `TBC_AND_ABOVE`

Database/Corrections/QuestieQuestBlacklist.lua
```diff
@@ -1041,7 +1041,7 @@ function QuestieQuestBlacklist:Load()
         [708] = true,
         [909] = true,
         [1288] = true,
-        [1661] = true,
+        [1661] = QuestieCorrections.WOTLK_ONLY,
```
BreakBB, line 1044:
```suggestion
        [1661] = QuestieCorrections.TBC_AND_WOTLK,
```

PR #4763, opened 2023-04-15
Fix quest Elmore's Task (1097)

gouletr:
Fix quest [Elmore's Task (1097)](https://www.wowhead.com/classic/quest=1097/elmores-task) missing [Verner Osgood (415)](https://www.wowhead.com/classic/npc=415/verner-osgood) as starter.


BreakBB:
Thanks @gouletr for another wonderful fix 👍🏻

PR #4764, opened 2023-04-18
Add option to hotfix vanilla UI bugs

Yttrium-tYcLief:
This adds an option to hotfix vanilla UI bugs, enabled by default, in the Advanced options tab.

The first vanilla bug hotfixed is this one: https://github.com/JamminL/wotlk-classic-bugs/issues/658

As of Wrath P2 (the DF API update), some quest dialogs are left visible and non-functional after accepting or declining them. This hotfix hooks the [QUEST_ACCEPTED](https://wowpedia.fandom.com/wiki/QUEST_ACCEPTED) event and checks if the `QuestFrameDetailPanel` dialog is still visible; if so, it closes it by simulating a mouseclick on the dialog's close button.

From preliminary searching there is no way to intercept the Decline button being pressed by the user, which unfortunately will still leave this phantom dialog visible, but it's very rare for a player to press Decline on these types of quests (since they mostly require clicking items to pick them up in the first place).

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -2637,6 +2637,7 @@ globals = {
     "ProductChoiceFrame",
     "PvPCommentator",
     "QuestFrame",
+    "QuestFrameCloseButton",
     "QuestFrameDetailPanel",
     "QuestFrameGreetingPanel",
     "QuestFrameGoodbyeButton",


Modules/Options/QuestieOptionsDefaults.lua
```diff
@@ -58,6 +58,7 @@ function QuestieOptionsDefaults:Load()
             enableMapIcons = true,
             enableMiniMapIcons = true,
             questieShutUp = false,
+            bugWorkarounds = true,
```
BreakBB, line 61: We want a migration step for this.

BreakBB (changes requested):
If this works on Wotlk and Era, we are ready to merge after adding a settings migration step.

PR #4838, opened 2023-04-30
Add channel to PlaySoundFile

MyTechnoHunter:
## Proposed changes
add "Master" channel to PlaySoundFile() to allow playing of sound while using ctrl-s to mute game ambient sounds


MyTechnoHunter:
er.. just an FYI.. noticed this while the build check was running.. 
image

BreakBB (approved):

PR #4884, opened 2023-05-11
[Feature] Questie Trackers "Sizer" needs a tooltip

### LINKED ISSUE #4815: Questie Trackers "Sizer" needs a tooltip
## Description
If the Active Quests Header is disabled by the Player then we lose the "info" tooltip in the ? icon. Additionally, most Players are unaware of the Right-Click feature on the Sizer (resets width back to auto width after manually resizing).

## Add Feature
While the Active Quests Header is disabled, move the tooltip info embedded in the ? icon over to the Sizer and insert additional tooltip info on the Sizers features. If the Active Quests Header is enabled, then the Sizers tooltip will only contain use info pertaining to the Sizers features.

The Sizer tooltip will also tell you if you're in Manual mode or Auto mode.

Another idea I've been playing with is to make a small 5 pixel "bar" at either the top or the bottom of the Tracker (for those that prefer to always hide the Active Quests Header). It'll turn Gold when hovered over (otherwise transparent) and the Tooltip that's embedded in the ? Icon (on the Active Quests Header) will appear there instead of the Sizer. This would give you guys the ability to Min/Max the tracker and have a tool tip and it would be at the top to indicate that Quests are being filled Top --> Down. And vis versa...

Dyaxler:
## Issue references

Fixes #4815 

## Proposed changes

- Translation files update
- While the mouse is over the Sizer and the "little tooltip" is shown, holding down the Shift key will make it expand. Letting go of the Shift key will change it back to the "little" one without having to move the mouse pointer off the Sizer. Just like the WorldMap/MiniMap Quest Icons tooltips.

## Screenshots
Tracker Sizer in Auto Mode:
image

image

Tracker Sizer in Manual Mode:
image

image


BreakBB (approved):
Looks fine to me 👍🏻

Starlynk1:
This is way easier with the tooltip now. One thing though is that in Auto mode you can't set the width, just the maximum height. I use my laptop a lot and this causes the tracker to be a bit too wide. Can there be a width option put in the tracker tab?

PR #4886, opened 2023-05-11
[Fix] Fixes a number of Tracker Formatting Updates

### LINKED ISSUE #4807: questie tracker not working
## Bug description

Recently my questie has had everything working besides the tracker, the tracker works for a little bit and then it doesn't update when I turn in or accept new quests then I reload my ui and then the tracker is completely gone. I've tried to do everything for the recommended fixes, any help would be appreciated.

## Screenshots

servers are down otherwise I would submit one 

## Questie & WoW version

Classic Era

### LINKED ISSUE #4860: add a fake tracker frame to help with using the percentage of screen slider in tracker options
## Description
should we have a "fake" tracker that shows when manipulating the percentage slider while options is open? People are not getting the settings right i think.. confusion and locked heights..
If your quest list isn't full you don't see it grow as it stops at the bottom of your current list

### LINKED ISSUE #4878: further enhancement to Tracker sizing
## Description
 we need better granularity for sizing options.. like i mentioned in channel (discord), the feature is having growing pains.. either the default height ratio should not be off for growth, or there should be a way to drag past the current number of quests to see for CERTAIN how big the tracker will get in manual mode with a confirmation popup when you let the sizer go to accept changes, possibly with a NOTE: this will disable auto mode on the confirm popup..
people are used to the tracker growing down and i personally love the scroll feature but not knowing the height.. or being stuck with the height limited to the number of quests you have, while using the sizer is.. counter productive
conversely, maybe a sizer feature expansion in a different direction? like modal windows in windows, mouse change at the bottom to allow sizing taller and edge to allow sizing wider with a corresponding disable of auto grow in that direction?
something.. as it is right now it feels unfinished..

### LINKED ISSUE #4893: Failed timed quests and tracker.
Failed timed quests exhibit these bugs in the tracker:
 - the timer stops at 1 second and the timer line is removed only if you move the tracker. It stays indefinitely visible otherwise or until another quest has an objective update.
 - It is missing the `Failed` text on objective
 - both the map and NPC icons/tooltips are not removed if you untrack this quest in the Failed state, they are forcibly shown

image
image
image

Wotlk 8.2.1

### LINKED ISSUE #4894: Quest tracker disappears when leaving an instance such as a dungeon.
## Bug description

I'm almost certain the questie tracker disappears when you leave an instance. When I enter a dungeon it's fine, but when I leave the tracker is gone and I have to keep going into questie and resetting the tracker.

## Screenshots

I don't really see the need for the image because it's just a screen with no quest tracker on the side.

## Questie & WoW version
 I'm playing on wotlk classic.

Dyaxler:
## Issue references

Fixes #4807 
Fixes #4860
Fixes #4878
Fixes #4894
Fixes #4893 

## Proposed changes

- Using the Tracker Height Ratio while in Sizer Auto Mode will now show you what the overall size of the Tracker will be set to. Once that area is filled with Quests and the overall height goes beyond the Height Ratio, scrolling will take over automatically.

- Using the Sizer to manually set the Trackers Height is no longer limited WHILE the Tracker is being sized. No more "rubber banding". Once you let go of the mouse, the Tracker will "trim" any grey areas and set the Height to the number of Tracked Quests/Achievements. This mode is still considered "Manual" so if anything else is tracked, scrolling will enable automatically unless the player decides to manually make the Tracker higher.

- Added some NOTE's to the Padding Between Quests and the Tracker Height Ratio settings to help the player understand the features.

- Updated applicable Localization strings

- Further refined the Tracker Update Formatting function so it can be used by itself without updating the Tracker.

- Fixed a bug in the Tracker Quest Frame that was preventing some formatting updates

- Fixed an issue with the Minimize Quest Button. It was showing up for Quests with Quest Item Buttons. Only the Maximize Quest Button should appear while the Quest is minimized.

- It would appear that the AutoUntrackedQuests or the TrackedQuests table can have QuestID's that have been left behind after a Quest has been completed or abandoned. This can throw off certain functions that check these tables and make the Tracker do weird things like, appear blank or having a Quest marked Tracked in the Quest Log and not appear in the Tracker. Questie now combs through these tables upon Login or reloadUI and removes any QuestID's that aren't already in the Players Quest Log. Added an additional check, to help prevent this from occuring in the first place, to the QuestieTracker:RemoveQuest() function that looks in these tables and if present, removes them when they are called from QuestieQuest:CompleteQuest() and QuestieQuest:AbandonedQuest().

- Quest Item Button was needed after Quest was completed. There was an item that was usable after a quest was complete and it disappeared. In order to get all quests to work, the Primary quest item button (not Secondary) should always be shown even after a quest is completed unless, "Blizzard Completion" text returns true. I have yet to encounter a quest where a quest item was needed when the quest reaches this state ("Blizzard Completion" text showing). This is only applicable to "Usable" Quest Items. Generally, when you complete a quest that once HAD a Primary Quest Item Button assigned; if that item is no longer usable, it will not appear anyway, so I don't foresee this causing any issues.


MyTechnoHunter:
> Using the Sizer to manually set the Trackers Height is no longer limited WHILE the Tracker is being sized. No more "rubber banding". Once you let go of the mouse, the Tracker will "trim" any grey areas and set the Height to the number of Tracked Quests/Achievements. This mode is still considered "Manual" so if anything else is tracked, scrolling will enable automatically unless the player decides to

This needs clarification please. 
If you release the mouse it shrinks in height to the amount of tracked, but does it then expand to what you dragged when/if you get more quests? Or stay short since its not auto?

Dyaxler:
> This needs clarification please. If you release the mouse it shrinks in height to the amount of tracked, but does it then expand to what you dragged when/if you get more quests? Or stay short since its not auto?

"This mode is still considered "Manual" mode." So, to answer your question, no because it's in Manual mode. If you want it to auto adjust your Trackers height automatically, then put the Sizer in Auto Mode.

Dyaxler:
General rule of thumb... if you size the tracker with the Sizer tab it's in "Manual" mode. To set it back to "Auto" mode then you right click the Sizer tab.

Auto Mode:
Width: Automatically adjusts the width based on the widest line in the Tracker with a few exceptions around wrapped text or split objectives.
Height: Automatically adjusts the height based on the number of Tracked objects up to the Tracker Height Ratio. Once this threshold is reached, scrolling is automatically enabled.

Manual Mode:
Width: Set by the Player and will never be automatically adjusted. Tracked objects that go beyond the Players set Width will get truncated.
Height: Set by the Player and will never be automatically adjusted with one exception noted below. Tracked objects that go beyond the Players set Height will enable scrolling automatically.

Note: To prevent the Tracker from being too tall potentially covering elements on the Players UI preventing those elements from being used we trim the tracker of any vertical "grey/white spaces" to keep the Trackers vertical boundary tightly around the number of tracked elements. This is especially important to Players who don't use the Background and Border options but it's more aesthetically pleasing. We don't adjust the width simply because some players might prefer to momentarily make the Tracker really wide so they can read the entire line.

BreakBB (approved):
These changes are really easy to read, thanks a lot! 💪🏻

PR #4897, opened 2023-05-14
Update `The Darkreaver Menace`

Ngoguey42:
Typos where made when setting `childQuests`.

`7769` is non existant. `7669` exists and is actually the child quest


BreakBB:
Whopps, thanks @Ngoguey42 👍🏻

Issue #4899, opened 2023-05-14
Would like to have the ability to link quests from map into chat via Shift+Lclick.


Cabro:
Currently you can hide available quests from Map with Shift+Lclick and Shift+Rclick.

I would like to have the functionality to just link into chat (when editbox is active) a quest from map by using Shift+Lclick.

Thank you!

Dyaxler:
Shift + Left Click is already assigned to the function of hiding that map icon via a yes/no popup. I'm sure you could add the functionality yourself and then make a PR. :)

`            if (not ChatFrame1EditBox:IsVisible()) then
                    %hide map icon%
            else
                if Questie.db.global.trackerShowQuestLevel then
                    ChatEdit_InsertLink(QuestieLink:GetQuestLinkString(quest.level, quest.name, quest.Id))
                else
                    ChatEdit_InsertLink("[" .. quest.name .. " (" .. quest.Id .. ")]")
                end
            end`

Cabro:
I'm afraid this is beyond my ability, i need a way to extract quest info from Questie map icons, but i got this going for me, which is nice:

`
        if (not ChatEdit_GetActiveWindow()) then
            if self.data.Type == "available" and IsShiftKeyDown() then
                StaticPopupDialogs["QUESTIE_CONFIRMHIDE"]:SetQuest(self.data.QuestData.Id)
                StaticPopup_Show ("QUESTIE_CONFIRMHIDE")
            elseif self.data.Type == "manual" and IsShiftKeyDown() and not self.data.ManualTooltipData.disableShiftToRemove then
                QuestieMap:UnloadManualFrames(self.data.id)
            end
        else
            if IsShiftKeyDown() and button == "LeftButton" then
                if Questie.db.global.trackerShowQuestLevel then
                    ChatEdit_InsertLink(QuestieLink:GetQuestLinkString(quest.level, quest.name, quest.Id))
                else
                    ChatEdit_InsertLink("[" .. quest.name .. " (" .. quest.Id .. ")]")
                end
            end
        end`

Dyaxler:
This assumes that self.data.Type is always "available"... it might nil out on you if self.data.Type is "manual" ;)

Cabro:
Can you manually make quests show on map?!

Dyaxler:
You'll have to be a bit more specific... can you give me a test/use case summary?

Cabro:
[...]

PR #4912, opened 2023-05-17
Shift+Lclick to link quests

Cabro:
Fixes https://github.com/Questie/Questie/issues/4899

### DIFF
--- a/Modules/FramePool/QuestieFrame.lua
+++ b/Modules/FramePool/QuestieFrame.lua
@@ -10,6 +10,8 @@ local QuestieDB = QuestieLoader:ImportModule("QuestieDB")
 local QuestieQuestBlacklist = QuestieLoader:ImportModule("QuestieQuestBlacklist")
 ---@type DailyQuests
 local DailyQuests = QuestieLoader:ImportModule("DailyQuests")
+---@type QuestieLink
+local QuestieLink = QuestieLoader:ImportModule("QuestieLink")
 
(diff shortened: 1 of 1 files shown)


BreakBB:
Just tested this and you zoom in, when linking a quest from the continent map. Do you think you can find a solution for this?

MyTechnoHunter:
when the chat is closed?

BreakBB:
No, when you link a quest to the chat from the continent map it zooms in and also pastes the quest to the chat. But I think it would be better to not zoom in.

MyTechnoHunter:
game client and version please and thanks.. i cant repro this in wrath.. 
image

BreakBB:
Your screenshot shows a zone map, not a continent map.

BreakBB:
Try linking a quest from here
grafik

MyTechnoHunter:
the zone map is the little one on the right.. the map itself is the main game map (leatrix maps addon makes it not full screen)

MyTechnoHunter:
wup.. i see what you mean.. on the continent level it copies the quest to chat and also clicks into the related area/zone map..  once IN the zone map it then copies fine without changing to a different map..

BreakBB:
Correct.

MyTechnoHunter:
the left click action seems to have always done that though? that part of the code wasnt touched..
we checked for right click, and if not right click we pass down the clicked map uid..
```lua
        else
            if self.UiMapID ~= WorldMapFrame:GetMapID() then
                WorldMapFrame:SetMapID(self.UiMapID);
            end
        end
```

PR #4918, opened 2023-05-18
[Feature] New Tracker Sort Options

### LINKED ISSUE #4787: Limit Proximity Tracking by Zone
This would be perfect it it just limited the listed objectives local to the zone the player was currently in.

image

### LINKED ISSUE #4908: Add the option "By Proximity (Reversed)" to the Tracker
Hi, I'm running the latest Questie version for Classic Era [V8.2.1] but please add this feature in for all game versions too since it should be a lot beneficial to them as well imho.

The suggestion is to add the option **"By Proximity (Reversed)"** to the Questie Tracker, so that I can keep it in the upper right corner (or upper left, for the same concept) and still have the quests closest to me nesting from the bottom to the top (it helps my brain visually understand that these are the closest to me).

Thanks! ^^

WoWScrnShot_051723_000209

Dyaxler:
## Issue references

Fixes #4908
Fixes #4787

## Proposed changes

The following Sort Options have been added to the Questie Tracker!
By % Complete (Reversed)
By Proximity (Reversed)
By Zone Proximity
By Zone Proximity (Reversed)

The first two are self-explanatory. The definition of By Zone Proximity is based on the Players "Current" zone. I.E. Only Quests in the Players current Zone will be sorted by Proximity.

## TODO

Add "Sort Quests" side menu with some addtional Sort Options to a Quests "Right Click" menu.


Gogo1951:
You should make the default setting "By Zone Proximity" - I assume that's any and all objectives in the current zone, sorted by distance? Yeah that should be everyone's default.

Dyaxler:
> You should make the default setting "By Zone Proximity" - I assume that's any and all objectives in the current zone, sorted by distance, as well as quests that get turned in to the current zone? Yeah that should be everyone's default. That's going to be huge for usability. Never having to manually sort and toggle quests again... that's huge.

To be clear... The Proximity Sort function will sort Quests not on the Quest's "over all" proximity, but it returns each Quest's nearest incomplete Objective. For completed Quest's, it returns the Quest Finishers cords. But only if this data is available, otherwise it assumes it's really far away because we preassign an arbitrary large number to each checks distance value to keep it from throwing a nil error. If that happens then it'll obviously put that Quest at the bottom of the list (furthest away).

Technically the "By Proximity" sort option will list ALL your Quests from nearest to furthest. So, if a Quest's Objective is in the Zone you're currently in, then they should all be grouped together "by Zone". You just can't see the Zone Headers while using that sorting type. All I did was added an additional filter for "By Zone Proximity" to list the Quest's by Zone first then by Proximity but only if your current Zone matched the Zone Header name.

I saw that screenshot you posted in the Issues thread. Some of those options are interesting and I had already thought about adding an option to place "Completed" Quest's at the top or bottom of the list. I also thought about adding an option to a Quest's "Right Click" menu. Sort Quest's --> "list of Sort Types" to give the player access to change it vs opening Questie Config and changing it under the Tracker Tab.

Issue #4744, opened 2023-03-31
It seems the extraObjectives of completed quests are repopulated upon a zone change.


Cabro:
Had the respective quest completed, did a small zone change before going to turn it in, only to notice extraObjective of "Mount Up" being active again.
Curiously, a /rl doesn't make them pop again.

![img](https://imgur.com/YqPWjAb.png)
7.5.0

Issue #3466, opened 2022-05-09
Tooltips for secondary objective quests are not updated/removed upon completion.


Cabro:
# Bug description

Tooltip for secondary objective will still show even though you completed the quest.
The map/minimap icons are correctly removed.

# Screenshots
![imgur](https://i.imgur.com/XcXQU3F.png)

# Questie & WoW version
6.10.0 TBC

Laumesis:
So tooltip stuff of both `extraObjectives` and `requiredSourceItems` are not removed once `isComplete ~= nil` (`1`or `-1`). Correct?

I think it will be easy fix after few other fixes / refactors are in place.

Laumesis:
Update:
This is in line with all other tooltip stuff. A quest and its objectives are kept in tooltips until it gets completed (<- blizzard word) (read: turned in).
The tooltip text should get changed to green once quest is complete. That's a bug at least.

Some with more power may comment on if intended behavior should be changed.

@Gogo1951 please do not reply, this is not to you.

BreakBB:
As @Laumesis already said, the current behaviour is streamlined to the regular objectives.

I agree that at least the secondary objectives are not useful to show once a quest is completed (not turned it yet).

Even though this is not a bug as it is the intended behaviour, we should change it 👍🏻

Dyaxler:
Opps... forgot to reference this along with #4948 

I'm almost 100% certain this is the same bug... however, since MY fix in #4948 pertains to Quest Finishers, these tooltips will still be present.

Cabro:
I guess just wait and see what BreakBB says

Dyaxler:
Yeah... because personally I think we should leave them. For "Normal" quests we leave the Objectives intact until AFTER the player turns in the Quest. Which is not broken (removing tooltips from Quest Finishers and Objectives after the quest is turned in).

Dyaxler:
@BreakBB Do you agree with my comment? We should leave these tooltips alone since they are attached to an Objective "creature" and they aren't duplicated.

MyTechnoHunter:
> @BreakBB said :
> As @Laumesis already said, the current behaviour is streamlined to the regular objectives.
> 
> I agree that at least the secondary objectives are not useful to show once a quest is completed (not turned it yet).
> 
> Even though this is not a bug as it is the intended behaviour, we should change it 👍🏻


[...]

PR #4932, opened 2023-05-24
[Fix] Tracker-Global-Settings reported by QA Team

Dyaxler:
## Issue references

Couple of different Discord channels references this issue. No official issue # on GitHub was created.
#4938 
#3466 
#4948 
#4744 

## Proposed changes

- Updated TrackerBaseFrame: Re-wrote some of the code responsible for Resizing the Tracker, Dragging the Tracker, and resetting the Sizer Auto/Manual feature.

- Updated TrackerBaseFrame: Added some new debug strings so we can better detect the various mouse button and click states users might be in.

- Updated QuestieTracker: Added some new debug strings and added checks to move the more spammy updates to the DEBUG_SPAM channel.

- Added/Changed a few more misc. debug strings in various places.

- Updated the Durability Frame handlers.

- Fixed Tooltips, Map Icons not updating extraObjectives AKA SpecialObjectives properly.

- Fixed Tooltips, Map Icons disappearing when zoning into and out of an instance.

- Fixed an issue with escort quests not updating properly after failing.

- Created securehook for StaticPopup_Show("DELETE_ITEM") so we can look for Quest Items that a player is trying to delete. This changes the popup dialog for just detected Quest Items.

- Created securehook for DeleteCursorItem() so we'll know when a player decides to delete a Quest Item. This triggers a Quest Log Update.

- Fixed an issue where a Quest Reward that was a Quest Item wasn't triggering a Tracker update nor an Item Tooltip update nor a Map Icon update.

- Created StaticPopup.lua translation file. We should refactor all StaticPopup text into this file. For now, I just moved the GameVersionError and the StaticPopup_Show hook text here.

- Renamed all the FadeTicker handlers to better reflect what it was doing.

- Refactored Quest Item Events & Added Player Trades and Merchants

- Consolidated events into one stack called: QuestRelatedFrameClosed() and unified all events to fire it.

- Passed event strings so Debug can output the triggered event.

- We can now handle Quest Items obtained via Mailbox (which would include items purchased via the Auction House), Bank, Guild Bank, Merchants and Player Trading.

HUGE THANK YOU to @MyTechnoHunter & @Cabro for their contributions and attention to detail!


Modules/Tracker/TrackerBaseFrame.lua
```diff
@@ -349,39 +356,66 @@ function TrackerBaseFrame.ShrinkToMinSize(minSize)
...
+                        if not IsMouselooking() then
+                            MouselookStart()
+                            mouseLookTicker = C_Timer.NewTicker(0.1, function()
+                                if not IsMouseButtonDown(button) then
+                                    MouselookStop()
+                                    mouseLookTicker:Cancel()
+                                end
+                            end)
+                        end
+                    --]]
```
BreakBB, line 402: Do we need this or not?

BreakBB (approved):

Modules/Tracker/TrackerBaseFrame.lua
Dyaxler: I left it in for a few more days of testing. It's not needed anymore. I think it was a REALLY old blizzard bug back in the early days of classic.

PR #4942, opened 2023-05-30
Fixed level check for daily filtering

BreakBB:
## Issue references

Fixes #

## Proposed changes

-
-

### DIFF
--- a/Modules/QuestieInit.lua
+++ b/Modules/QuestieInit.lua
@@ -308,7 +308,7 @@ QuestieInit.Stages[3] = function() -- run as a coroutine
 
     Questie.started = true
 
-    if Questie.IsWotlk and QuestiePlayer.GetPlayerLevel() == 70 then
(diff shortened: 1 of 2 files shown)


MyTechnoHunter:
is maxlevel required? can you do the dalaran dailies for example at max _profession_ level?

BreakBB:
This is just a minor change, which we don't want to lose.

Currently this will have no effect and for activating the Daily quest filtering we need a lot more testing.

BreakBB:
Also on your actual question: At least wowhead says that for example the Fishing quests require you to be level 70 and no skill level is required.

MyTechnoHunter:
sooo.. not max level in wrath then? being as max = 80?

BreakBB:
Yeah for the fishing stuff it is not. But I am not planning to change the „IsDoable logic“, so that doesn’t matter.

The DailyQuest module is about filtering quests which are not available on _that_ day.

PR #4968, opened 2023-06-13
[fix] Fixed weird text format for map/minimap tooltips

### LINKED ISSUE #4892: Would like to have the formatting from QuestLink into the minimap tooltip.
They are not the same, it seems the minimap tooltip (right) writes each phrase to a new line, and messing with the original quest text format. QuestLink (left) is the one that is showing it as it should be.

Also, minimap tooltip has problems with special characters (notice it's not displaying `)` at the end).

image

wotlk 8.2.1

Cabro:
Attempted fix for text wrap for map/minimap tooltip text when shift is pressed, so it doesn't show all weird and unnecessary split in the text definitions of objectivestext in the DB files. Code is not pretty, but the result is clean.
Fixes #4892

### DIFF
--- a/Modules/Libs/QuestieLib.lua
+++ b/Modules/Libs/QuestieLib.lua
@@ -686,7 +686,7 @@ function QuestieLib:TextWrap(line, prefix, combineTrailing, desiredWidth)
             until (#indexes > 1) --Until more than one row
 
             --Get the line we calculated
-            --First to Dot, then space and lastly endIndex(chinese)
+            --First to space then endIndex(chinese)
             local newLine = string.sub(useLine, startIndex, lastSpaceIndex or endIndex)
 
             --This combines a trailing word to the previous line if it is the only word of the line


Modules/Libs/QuestieLib.lua
Dyaxler, line 1:
```lua
--Part of the GameTooltipWrapDescription function
local textWrapObjectiveFontString
---Emulates the wrapping of a quest description
---@param line string @The line to wrap
---@param prefix string @The prefix to add to the line
---@param combineTrailing boolean @If the last line is only one word, combine it with previous? TRUE=COMBINE, FALSE=NOT COMBINE, default: true
---@param splitOnDot boolean @Should we add a linebreak if a dot appears thats not at the end of a line TRUE=NEW ROW, FALSE=NO NEW ROW, default: true
---@param desiredWidth number @Set the desired width to wrap, default: 275
---@return table[] @A table of wrapped lines
function QuestieLib:TextWrap(line, prefix, combineTrailing, splitOnDot, desiredWidth)
```

If you're removing splitOnDot then you should update the function vars and @param.

The two calls to this function in MapIconTooltip.lua that you updated with `false, false` remove one of those once you update the above function. No reason to have references to vars we're no longer using in the code. Might confuse future contributors.

Cabro: Added a commit with removed vars from call and function

Modules/Libs/QuestieLib.lua
```diff
@@ -652,14 +651,13 @@ function QuestieLib:TextWrap(line, prefix, combineTrailing, splitOnDot, desiredW
 
     --Set Defaults
     combineTrailing = combineTrailing or true
-    splitOnDot = splitOnDot or true
     --We show the fontstring and set the text to start the process
     --We have to show it or else the functions won't work... But we set the opacity to 0 on creation
     textWrapObjectiveFontString:SetWidth(desiredWidth or textWrapFrameObject:GetWidth() or 275) --QuestLogObjectivesText default width = 275
     textWrapObjectiveFontString:Show()
 
     --Make a linebreak on each "dot" character if there is a space after (don't want it on end of line)
```
BreakBB, line 659:
```suggestion
```

Modules/Libs/QuestieLib.lua
```diff
@@ -694,7 +687,7 @@ function QuestieLib:TextWrap(line, prefix, combineTrailing, splitOnDot, desiredW
 
             --Get the line we calculated
             --First to Dot, then space and lastly endIndex(chinese)
```
BreakBB, line 689:
```suggestion
            --First to space then endIndex(chinese)
```

PR #5016, opened 2023-07-04
[quest] The Ulduar quests map markers should be now visible.

Cabro:
Trying a way to make Questie show map markers inside the raid maps. It involves using some fake IDs. I would like for someone else to double check and test this PR. It all works great locally, and I tested with modifying some loot tables to also include the objects and NPCs in these changes, and they were drawn correctly.

Also, if you log in outside of Ulduar, the quest map markers should also be visible in the correct map levels of Ulduar when you enter.

If this fix seems to be good, then I can add map markers for the rest of the dungeons.


Database/Corrections/wotlkNPCFixes.lua
```diff
@@ -984,6 +984,10 @@ function QuestieWotlkNpcFixes:Load()
             [npcKeys.questStarts] = {13500},
             [npcKeys.questEnds] = {13500},
         },
+        [33288] = {
+            [npcKeys.spawns] = {[zoneIDs.ULDUAR]={{-1,-1}},[zoneIDs.THE_PRISON_OF_YOGG_SARON]={{68,40.2}}},
```
BreakBB, line 988: I am not sure if these spawns should be the other way around in order to work 🤔 I have something in mind that a loop stops when encountering a `-1,-1` spawn.

Cabro: They were showing just fine when i was tweaking them, but i'll change the order

BreakBB:
Can this be merged @Cabro ?

Cabro:
They all work locally for me, no issues. Just wanted someone to double check the `requiredSpell` changes in the function for logic, and whether someone had a problem with the Ulduar quests, like nil errors or something.

Muehe:
Merged by manual rebase, see fe6c56a00019153f62a841c393089d93d8e1af54 and its 6 parent commits.

PR #5031, opened 2023-07-12
DB Refactor

Logonz:
This PR contains the code needed to dump the corrected data to file which is then used by the new QuestieDB system


Logonz:
Something is emptying fields of data, probably some kind of blacklist
```
[7909] = {'Aquamarine',{92 <-- dropped by
[7910] = {'Star Ruby', {} <- dropped by
```

The strange thing is that Aquamarine which also has a ton of drops isn't empty.

PR #5043, opened 2023-07-17
Misc fixes July 17

Dyaxler:
Change details are noted in each commit or linked issue.

### DIFF
--- a/Modules/Libs/QuestieLib.lua
+++ b/Modules/Libs/QuestieLib.lua
@@ -129,8 +129,7 @@ end
 ---@param showState boolean @ Whether to show (Complete/Failed)
 ---@param blizzLike boolean @True = [40+], false/nil = [40D/R]
 function QuestieLib:GetColoredQuestName(questId, showLevel, showState, blizzLike)
-    local quest = QuestieDB:GetQuest(questId)
-    local name = quest.name
+    local name = QuestieDB.QueryQuestSingle(questId, "name")
     local level, _ = QuestieLib.GetTbcLevel(questId);
 
     if showLevel then
@@ -142,15 +141,15 @@ function QuestieLib:GetColoredQuestName(questId, showLevel, showState, blizzLike
     end
 
     if showState then
-        local isComplete = quest:IsComplete()
+        local isComplete = QuestieDB.IsComplete(questId)
 
         if isComplete == -1 then
             name = name .. " " .. Questie:Colorize("(" .. l10n("Failed") .. ")", "red")
         elseif isComplete == 1 then
             name = name .. " " .. Questie:Colorize("(" .. l10n("Complete") .. ")", "green")
 
-            -- There are a number of quests that are flagged incomplete but they have a "synthetic" objective that is complete
-        elseif isComplete == 0 and #quest.Objectives == 0 and quest.isComplete == true then
+        -- Quests treated as complete - zero objectives or synthetic objectives
+        elseif isComplete == 0 and QuestieDB:GetQuest(questId).isComplete == true then
             name = name .. " " .. Questie:Colorize("(" .. l10n("Complete") .. ")", "green")
         end
     end
--- a/Modules/Libs/QuestieLib.lua
+++ b/Modules/Libs/QuestieLib.lua
@@ -143,9 +143,9 @@ function QuestieLib:GetColoredQuestName(questId, showLevel, showState, blizzLike
         local isComplete = QuestieDB.IsComplete(questId)
 
         if isComplete == -1 then
-            name = name .. " (" .. Questie:Colorize(l10n("Failed"), "red") .. ")"
+            name = name .. " " .. Questie:Colorize("(" .. l10n("Failed") .. ")", "red")
         elseif isComplete == 1 then
-            name = name .. " (" .. Questie:Colorize(l10n("Complete"), "green") .. ")"
+            name = name .. " " .. Questie:Colorize("(" .. l10n("Complete") .. ")", "green")
         end
     end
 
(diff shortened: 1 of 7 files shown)


BreakBB (approved):

Modules/Libs/QuestieLib.lua
```diff
@@ -129,23 +129,29 @@ end
 ---@param showState boolean @ Whether to show (Complete/Failed)
 ---@param blizzLike boolean @True = [40+], false/nil = [40D/R]
 function QuestieLib:GetColoredQuestName(questId, showLevel, showState, blizzLike)
-    local name = QuestieDB.QueryQuestSingle(questId, "name");
+    local quest = QuestieDB:GetQuest(questId)
```
BreakBB, line 132: What's the reason for this change? `QueryQuestSingle` is many times faster than `QuestieDB:GetQuest` especially if you only need a few values.

Dyaxler: I'll re-think this

Modules/Tracker/QuestieTracker.lua
```diff
@@ -749,11 +749,13 @@ function QuestieTracker:Update()
                     end
 
                     -- Completion Text should never be allowed to Auto Collapse
-                    --local shouldCollapse = (#quest.Objectives ~= 0 and completionText == nil) or (Questie.db.global.hideBlizzardCompletionText or objectiveColor == "minimal")
                     local shouldCollapse = completionText == nil or (Questie.db.global.hideBlizzardCompletionText or objectiveColor == "minimal")
 
+                    -- We need to redifine a completed quest here so we can auto-minimize completed quests properly
+                    local questCompleted = complete == 1 or (#quest.Objectives == 0 and quest.isComplete == true and completionText == nil)
```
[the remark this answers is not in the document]
Dyaxler: I was looking at all that today... it's kind of a spaghetti nest at the moment. There have been a lot of bug fixes and logic improvements done to the underlying functions that make some of the tracker checks redundant.

Modules/Libs/QuestieLib.lua
```diff
@@ -142,15 +141,15 @@ function QuestieLib:GetColoredQuestName(questId, showLevel, showState, blizzLike
     end
 
     if showState then
-        local isComplete = quest:IsComplete()
+        local isComplete = QuestieDB.IsComplete(questId)
```
BreakBB, line 144: What's the difference between `QuestieDB.IsComplete(questId)` and `QuestieDB:GetQuest(questId).isComplete` a few lines below?

Dyaxler: The QuestieDB.IsComplete() function is embedded into the `quest` table that is returned from QuestieDB:GetQuest()... I.E. `QuestieDB:GetQuest(questId):IsComplete()`. The difference? Code wise, nothing. But since I'm not storing the entire `quest` table into a local var I can't use `quest:IsComplete()`. Make sense?

Dyaxler: Sorry... you asked about `QuestieDB:GetQuest(questId).isComplete`. That is a key that gets flagged true when all objectives are complete. It's designed to identify quests that should be complete but `:IsComplete()` returns zero or -1.

BreakBB: My point was: Why calling the heavy `QuestieDB:GetQuest(questId).IsComplete` when you already called `QuestieDB.IsComplete(questId)` a few lines before?

Is there a difference of outcome? Can the one be „more complete“ than the other? Or is it just `boolean` vs. `number`?

Dyaxler: `QuestieDB.IsComplete(questId)` and `QuestieDB:GetQuest(questId):IsComplete()` are the same function. It only returns -1, 1 or 0 and it's based on what Blizzard returns. There are quite a few quests that return zero or -1 that should be returning 1... this is why we set the `QuestieDB:GetQuest(questId).isComplete` flag to true when all the objectives are complete - in short it ignores `IsComplete()` and forces a quest to be treated as complete even though `IsComplete()` returns something other than 1. Make sense?

Modules/Libs/QuestieLib.lua
```diff
@@ -143,9 +144,13 @@ function QuestieLib:GetColoredQuestName(questId, showLevel, showState, blizzLike
...
 
         if isComplete == -1 then
-            name = name .. " (" .. Questie:Colorize(l10n("Failed"), "red") .. ")"
+            name = name .. " " .. Questie:Colorize("(" .. l10n("Failed") .. ")", "red")
         elseif isComplete == 1 then
-            name = name .. " (" .. Questie:Colorize(l10n("Complete"), "green") .. ")"
+            name = name .. " " .. Questie:Colorize("(" .. l10n("Complete") .. ")", "green")
+
+        -- Quests treated as complete - zero objectives or synthetic objectives
+        elseif isComplete == 0 and QuestieDB:GetQuest(questId).isComplete == true then
```
BreakBB, line 152: So this call of `QuestieDB:GetQuest(questId).isComplete` is required and can have a different value than what `QuestieDB.IsComplete(questId)` return? @Dyaxler

Dyaxler: Correct @BreakBB

Modules/Tracker/QuestieTracker.lua
```diff
@@ -436,13 +436,14 @@ end
 -- Quest Item Button can be switched on and appear in the tracker.
 ---@param text string
 function QuestieTracker:QuestItemLooted(text)
-    local playerLoot = strmatch(text, "You receive loot")
+    local playerLoot = strmatch(text, "You receive ")
```
BreakBB, line 439: Does this work for other languages as well?

BreakBB: Perfect 👍🏻

BreakBB:
For me this PR is ready to be merged and can be shipped 🚢 

@MyTechnoHunter @Cabro do you agree?

PR #5060, opened 2023-08-06
Add Profiles system

### LINKED ISSUE #4930: Add actual "Profiles" to Questie
## Description

The settings have always been chaos and there still is no way to tell which settings are "per character" and which are "global".
To tackle this Questie needs actual "Profiles" and should allow to handle _all_ settings via Profiles.

The [Ace docs](https://www.wowace.com/projects/ace3/pages/api/ace-db-3-0) are quite clear about it and we have most things already in place since we already use AceDB for our config.

So instead of `Questie.db.char` and `Questie.db.global` we will use `Questie.db.profile` for all profile related settings. This will need a big migration.

Taken from GatherMate2:
grafik

The "Profiles" tab is added by:

```diff
    return {
        name = "Questie",
        handler = Questie,
        type = "group",
        childGroups = "tab",
        args = {
            general_tab = general_tab,
            social_tab = social_tab,
            minimap_tab = minimap_tab,
            map_tab = map_tab,
            dbm_hud_tab = dbm_hud_tab,
            tracker_tab = tracker_tab,
            nameplate_tab = nameplate_tab,
            tooltip_tab = tooltip_tab,
            advanced_tab = advanced_tab,
+           profiles = LibStub("AceDBOptions-3.0"):GetOptionsTable(Questie.db)
        }
    }
```

BreakBB:
## Issue references

Fixes #4930

## Proposed changes

- Move all settings related SVs from `char`/`global` to `profile`
- Reset all settings except for the Journey

## Things unchanged

- The compiled database still is saved in `global` space
- Completed and hidden quests are still saved in `char` space
- Most _Tracker-State` (e.g. if collapsed, what quests are hidden, ...) related things are still in `char` space
- Debug is still in `global` space
- Icons (.e.g. `ICON_OBJECT`) are still in `global` space

## TODOs

- [x] Currently the migration code does not work, but resets for every `/reload`
    -  Even though `Questie.db.profile.migrationVersion` is set to `1` on start up, it is reset to `0` on the next `/reload`
    - There are print-Statements showing this atm
    - What is even more confusing is, that other values stored in the same manner are kept correctly
- [x] We need to check what is left todo for `Questie:RefreshConfig`
    - I think Migration should run there as well
    - The Tracker position needs to be updated (looks like `QuestieTracker:Update` is not enough to do so)
- [ ] Townsfolk need to be checked, if it can stay the way it is
- [x] Remove `Questie.db.profile.globalTrackerLocation` setting as it is no longer needed at all
- [x] Fix `QuestieOptionsAdvanced.lua:146: bad argument #1 to 'band' (number expected, got nil)` which happens with a fully fresh Questie SV file


Modules/FramePool/QuestieFrame.lua
```diff
@@ -426,8 +426,8 @@ end
 ---Checks wheather the frame/icon should be hidden or not. Only for quest icons/frames.
 ---@return boolean @True if the frame/icon should be hidden and :FakeHide() should be called, false otherwise
 function _Qframe:ShouldBeHidden()
-    local questieGlobalDB = Questie.db.global
-    local questieCharDB = Questie.db.char
+    local questieGlobalDB = Questie.db.profile -- TODO: Replace questieGlobalDB and questieCharDB
```
BreakBB, line 429: Reminder for this TODO

Modules/Map/QuestieMap.lua
```diff
@@ -629,7 +627,8 @@ function QuestieMap:DrawWorldIcon(data, areaID, x, y, showFlag)
         end
 
         function iconMinimap:FadeLogic()
-            if self.miniMapIcon and self.x and self.y and self.texture and self.UiMapID and self.texture.SetVertexColor and Questie and Questie.db and Questie.db.global and Questie.db.global.fadeLevel and HBD and HBD.GetPlayerZonePosition and QuestieLib and QuestieLib.Euclid then
+            local profile = Questie.db.profile
+            if self.miniMapIcon and self.x and self.y and self.texture and self.UiMapID and self.texture.SetVertexColor and Questie and Questie.db and Questie.db.profile and Questie.db.profile.fadeLevel and HBD and HBD.GetPlayerZonePosition and QuestieLib and QuestieLib.Euclid then
```
BreakBB, line 631:
```suggestion
            if self.miniMapIcon and self.x and self.y and self.texture and self.UiMapID and self.texture.SetVertexColor and profile.fadeLevel and HBD and HBD.GetPlayerZonePosition and QuestieLib and QuestieLib.Euclid then
```

Modules/Options/QuestieOptions.lua
```diff
@@ -90,16 +90,16 @@ function QuestieOptions:OpenConfigWindow()
 end
 
 -- get option value
-function QuestieOptions:GetGlobalOptionValue(info)
-    return Questie.db.global[info[#info]]
+function QuestieOptions:GetGlobalOptionValue(info) -- TODO: Remove this?
```
BreakBB, line 93: Reminder for this TODO

Modules/Options/QuestieOptions.lua
```diff
@@ -90,16 +90,16 @@ function QuestieOptions:OpenConfigWindow()
...
+    return Questie.db.profile[info[#info]]
 end
 
 -- set option value
-function QuestieOptions:SetGlobalOptionValue(info, value)
-    if debug and Questie.db.global[info[#info]] ~= value then
-        Questie:Debug(Questie.DEBUG_SPAM, "DEBUG: global option", info[#info], "changed from '" .. tostring(Questie.db.global[info[#info]]) .. "' to '" .. tostring(value) .. "'")
+function QuestieOptions:SetGlobalOptionValue(info, value) -- TODO: Remove this?
```
BreakBB, line 98: Reminder for this TODO

Modules/QuestieInit.lua
```diff
@@ -100,7 +100,7 @@ end
 
 ---Run the validator
 local function runValidator()
-    if Questie.db.global.skipValidation then
+    if Questie.db.profile.skipValidation then
```
BreakBB, line 103:
```suggestion
    if Questie.db.global.skipValidation then
```

And yes, we still need a checkbox for this.

Modules/QuestieInit.lua
```diff
@@ -310,12 +310,12 @@ QuestieInit.Stages[3] = function() -- run as a coroutine
     Questie.started = true
 
     if (Questie.IsWotlk or Questie.IsTBC) and QuestiePlayer.IsMaxLevel() then
-        local lastRequestWasYesterday = Questie.db.char.lastDailyRequestDate ~= date("%d-%m-%y"); -- Yesterday or some day before
-        local isPastDailyReset = Questie.db.char.lastDailyRequestResetTime < GetQuestResetTime();
+        local lastRequestWasYesterday = Questie.db.profile.lastDailyRequestDate ~= date("%d-%m-%y"); -- Yesterday or some day before
```
BreakBB, line 313: This should be global as well, even though it doesn't work atm.

PR #5090, opened 2023-09-07
[fix] Hide PvP quests in Hardcore

### LINKED ISSUE #5081: [Hardcore] Questie is showing BG/pvp quests on hardcore realms.
It needs a way to differentiate between Era and Hardcore servers - realm name. Clients are the same.

BreakBB:
## Issue references

Fixes #5081

Blacklists all quests from this list https://www.wowhead.com/classic/quests/battlegrounds


MyTechnoHunter (approved):
looks good

PR #5096, opened 2023-09-09
[fix] Fix map icon tooltips not showing multiple objectives

BreakBB:
## Issue references

Fixes #

## Proposed changes

-
-

## Screenshots

Before:
image

With the fix:
image

### DIFF
--- a/Modules/Tooltips/MapIconTooltip.lua
+++ b/Modules/Tooltips/MapIconTooltip.lua
@@ -31,7 +31,6 @@ local REPUTATION_ICON_TEXTURE = "|T" .. REPUTATION_ICON_PATH .. ":14:14:2:0|t"
 local TRANSPARENT_ICON_PATH = "Interface\\Minimap\\UI-bonusobjectiveblob-inside.blp"
 local TRANSPARENT_ICON_TEXTURE = "|T" .. TRANSPARENT_ICON_PATH .. ":14:14:2:0|t"
 
-local WRAP_TEXT = 1;
 local DEFAULT_WAYPOINT_HOVER_COLOR = { 0.93, 0.46, 0.13, 0.8 }
 
 local lastTooltipShowTimestamp = GetTime()
@@ -549,14 +548,16 @@ function _MapIconTooltip:AddTooltipsForQuest(icon, tip, quest, usedText)
     for text, nameTable in pairs(tip) do
         local data = {}
         data[text] = nameTable
-        --Add the data for the first time
+        -- Add the data for the first time
         if not usedText[icon.data.Id] then
-            usedText[icon.data.Id] = {}
-
-            if not usedText[icon.data.Id][text] then
-                tinsert(quest, data)
-                usedText[icon.data.Id][text] = true
-            end
+            usedText[icon.data.Id] = {
+                [text] = true
+            }
+            tinsert(quest, data)
+            -- add another line to an existing entry
+        elseif not usedText[icon.data.Id][text] then
+            tinsert(quest, data)
+            usedText[icon.data.Id][text] = true
         else
             --We want to add more NPCs as possible candidates when shift is pressed.
             if icon.data.Name then
--- a/Modules/Tooltips/MapIconTooltip.lua
+++ b/Modules/Tooltips/MapIconTooltip.lua
@@ -356,15 +356,20 @@ function MapIconTooltip:Show()
             local defaultQuestColor = QuestieLib:GetRGBForObjective({})
             if shift then
                 local creatureLevels = QuestieDB:GetCreatureLevels(quest) -- Data for min and max level
+                local addedCreatureNames = {}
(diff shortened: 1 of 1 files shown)


Cabro:
Seems to be working, now i see multiple objectives on the world map
image

In your example, is there a way to make it show both Meat and Skin on the same NPC source without adding a new line for each objective?

BreakBB:
> In your example, is there a way to make it show both Meat and Skin on the same NPC source without adding a new line for each objective? Just make it show as it shows on the creature tooltip itself?

It wasn't like that before, was it? I mean Questie always had it on seperate lines, right?

Cabro:
> It wasn't like that before, was it? I mean Questie always had it on seperate lines, right?

I am not certain of anything anymore :D

MyTechnoHunter:
As far as I know we've always had objectives On Their Own Line I don't think we ever combined them

BreakBB:
> As far as I know we've always had objectives On Their Own Line I don't think we ever combined them

I think so too 😅

Cabro:
I am not sure what is happening here, if it's related to this PR or not, but:

image

The objective is in the form of object + custom text linked to it, and it doesn't pick up the object itself.

`[questKeys.objectives] = {nil,{{400062,"Test the Eggscilliscope Prototype"}}},`

MyTechnoHunter:
er, i am confused.. wht seems to be the issue? that its not showing rookery egg?

Cabro:
> er, i am confused.. wht seems to be the issue? that its not showing rookery egg?

image

MyTechnoHunter:
and egg freezing isnt the quest name?

still confused

Cabro:
Just like... every other quest in that list has a source for the objective, such as "Halycon", "Spire Spider Egg" etc,  the first quest is missing it. It should be another line between the quest name "[60+] Egg Freezing (4734)" and the objective "Test the Eggscilliscope Prototype" with the source of the objective - "Rookery Egg"

MyTechnoHunter:
where is the custom object referenced.. does it have a valid entry to show? nothing goofy like spaces, or not being in the relevant section of the DB?

Cabro:
```
        [4734] = {
            [questKeys.objectives] = {nil,{{400062,"Test the Eggscilliscope Prototype"}}},
        },
```

MyTechnoHunter:
is that code you added? i dont see it when i search my questie?

MyTechnoHunter:
and that makes no sense.. the comment claims its in upper blackrock spire.. and the zone says lower?

Cabro:
And an addendum to the previous comment: all those NPCs listed in the image above have different spawn points, with different quest icons on the map. I found a case where if you have mobs with the exact same spawn coords, only 1 of them gets listed: 
image

There should be 4 NPCs that drop it: Earthen, Hoary, Azure and Crimson. 

In my classicNPCFixes.lua correction file i put this:
```
        [15209] = {
            [npcKeys.zoneID] = zoneIDs.SILITHUS,
            [npcKeys.spawns] = {[zoneIDs.SILITHUS] = {{37.3,44.8},{38.29,46.46},{24.44,36.00},{27.93,30.62},{20.46,86.16},{18.23,81.12}}},
        },
        [15211] = {
            [npcKeys.zoneID] = zoneIDs.SILITHUS,
            [npcKeys.spawns] = {[zoneIDs.SILITHUS] = {{37.3,44.8},{38.29,46.46},{24.44,36.00},{27.93,30.62},{20.46,86.16},{18.23,81.12}}},
        },
        [15212] = {
            [npcKeys.zoneID] = zoneIDs.SILITHUS,
            [npcKeys.spawns] = {[zoneIDs.SILITHUS] = {{37.3,44.8},{38.29,46.46},{24.44,36.00},{27.93,30.62},{20.46,86.16},{18.23,81.12}}},
        },
        [15307] = {
            [npcKeys.zoneID] = zoneIDs.SILITHUS,
            [npcKeys.spawns] = {[zoneIDs.SILITHUS] = {{37.3,44.8},{38.29,46.46},{24.44,36.00},{27.93,30.62},{20.46,86.16},{18.23,81.12}}},
        },
```
And it only shows one of the 4 as source.

BreakBB:
> Back to this PR. Since it can already group multiple sources, why not group multiple objectives? :)

I will take another look into the code, to see if I can understand how/where that is done.

> And it only shows one of the 4 as source. Could this be because when the map icon was created, only 1 npc was linked to it? Maybe the issue is in the map icon creation?
>
> LE: Earthen Templar has the highest ID of the 4 NPCs, 15307, so most likely it is updating (as in replacing) the other NPCs tooltips for this map icon.

It can indeed be, that the last ID overrides the others. Might be the code currently uses an ID which is always the same 🤔 Need to verify that in the code.

BreakBB:
@Cabro this should be fixed with the last commit:

> Back to this PR. Since it can already group multiple sources, why not group multiple objectives? :)

Issue #3418, opened 2022-04-19
Thousand Needles - A Dip In The Moonwell - Icon moves on the minimap


SF0614R18:
In this quest you have to take control of a small robot and retrieve a sample of water from a moonwell.  If you walk beyond the moonwell, the icon on the minimap which shows the location of the moonwell begins to move.  If you keep trying to walk towards the icon on the minimap, you can never reach the icon because it keeps moving away from you.

BreakBB:
That sounds very interesting. I just did the quest a few days ago but didn't notice.

Will try to look into it with another character thanks for reporting it.

SF0614R18:
1) Take control of robot
2) Instead of going into moonwell, keep running past it
3) Pay attention to the icon in the minimap which is supposed to show where the moonwell is, it moves.

This is on the ARM version of WoW.  I’m running a MacBook.  Not sure if this matters because usually Questie runs flawlessly.

Thank you for this amazing addon.

SF0614R18:
Also this is TBC Classic.

graevy:
still an issue. when you pass character control to the robot, your character still exists in the world, and the icon moves around

BreakBB:
In the Wotlk beta I experienced a similar behaviour with another quest. I fear this is not something easy to fix for us as this is kind of a data mismatch between game API and the library we use to place icons on the map 🙈

PR #5130, opened 2023-09-19
[fix] Fix coordinates tracking while in a vehicle

BreakBB:
## Issue references

Related to #3418

## Proposed changes

-
-


BreakBB:
This doesn't seem to be possible with the current APIs. The quest in 1k Needles puts the player in a "mind controlling" state, but it is not possible to get the position of the controlled unit and/or the "pet" (the vehicle you are controlling).

PR #5150, opened 2023-10-06
Rework draw available quests

BreakBB:
## Issue references

- 

## Proposed changes

- Introduce the `AvailableQuest` module which handles `CalculateAndDrawAll`


Database/QuestieDB.lua
```diff
@@ -38,6 +38,11 @@ local QuestieQuest = QuestieLoader:ImportModule("QuestieQuest")
 ---@type QuestieQuestPrivate
 local _QuestieQuest = QuestieQuest.private
 
+--- A list of quests that will never be available, used to quickly skip quests.
+---@alias AutoBlacklistString "rep"|"skill"|"race"|"class"
+---@type table<number, AutoBlacklistString>
+QuestieDB.autoBlacklist = {}
```
BreakBB, line 44: This is used in `QuestieDB.lua` and `AvailableQuests.lua`. Not sure if there is a better place for it.

Modules/Quest/AvailableQuests.lua
```diff
@@ -0,0 +1,349 @@
...
+        end
+
+        if (
+            (not QuestieDB.IsLevelRequirementsFulfilled(questId, minLevel, maxLevel, playerLevel)) or
+            (not QuestieDB.IsDoable(questId, debugEnabled))
+        ) then
+            --If the quests are not within level range we want to unload them
+            --(This is for when people level up or change settings etc)
+            -- TODO: Is still necessary?
+            QuestieMap:UnloadQuestFrames(questId)
```
BreakBB, line 174: Could this added to the `if` below? Not sure when we are adding frames, so not sure when we need to unload them.

Technically the above should be true, because we are talking about available quests and these are tracked in `availableQuests`.

Modules/Quest/AvailableQuests.lua
```diff
@@ -0,0 +1,349 @@
...
+
+---@param callback function | nil
+function AvailableQuests.CalculateAndDrawAll(callback)
+    Questie:Debug(Questie.DEBUG_INFO, "[AvailableQuests.CalculateAndDrawAll]")
+
+    --? Cancel the previously running timer to not have multiple running at the same time
+    if timer then
+        timer:Cancel()
+    end
+    timer = ThreadLib.Thread(_CalculateAvailableQuests, 0, "Error in CalculateAndDrawAll", callback)
```
BreakBB, line 222:
```suggestion
    timer = ThreadLib.Thread(_CalculateAvailableQuests, 0, "Error in AvailableQuests.CalculateAndDrawAll", callback)
```

Modules/Quest/AvailableQuests.lua
```diff
@@ -0,0 +1,349 @@
...
+end
+
+--Draw a single available quest, it is used by the CalculateAndDrawAll function.
+---@param quest Quest
+function AvailableQuests.DrawAvailableQuest(quest) -- prevent recursion
```
BreakBB, line 227: The content of both `if`-blocks are nearly identical and should be generalized.

Database/QuestieDB.lua
BreakBB: Most likely this should be moved to `AvailableQuests.lua`, but only if `QuestieDB.IsDoable` is also moved.

Modules/QuestieEventHandler.lua
```diff
@@ -78,13 +80,13 @@ function QuestieEventHandler:RegisterLateEvents()
     -- Spell objectives
     Questie:RegisterEvent("NEW_RECIPE_LEARNED", function() -- Needed for some spells that don't necessarily appear in the spellbook, but are definitely spells
         Questie:Debug(Questie.DEBUG_DEVELOP, "[EVENT] NEW_RECIPE_LEARNED")
-        QuestieQuest.CalculateAndDrawAvailableQuestsIterative()
+       AvailableQuests.CalculateAndDrawAll()
```
BreakBB, line 83:
```suggestion
        AvailableQuests.CalculateAndDrawAll()
```

PR #5160, opened 2023-10-11
Fixed messed up ICC raid name translation

satanch:
## Proposed changes

Localization for `ptBR` and `ruRU` was mixed up with each other.


BreakBB:
Whoops 😅 Thanks a lot for this contribution @satanch ❤️

PR #5166, opened 2023-10-14
Update Comms.lua

nanjuekaien1:
~~I will complete the other missing zhCN localization documentation in the near future.~~


MyTechnoHunter:
look good, 
noticed that some names are surrounded by < > (Questie and QuestieComms)
is this normal?

Localization/Translations/DebugMessages.lua
```diff
@@ -432,7 +432,7 @@ local debugMessagesLocales = {
         ["frFR"] = false,
         ["esES"] = false,
         ["zhTW"] = "|cffff0000請注意:|r 任務位置提示插件之後的更新將會重置您的設定，建議您預先備份 Questie 相關的設定檔案，只要備份魔獸安裝資料夾內的 WTF 資料夾就可以了。",
-        ["zhCN"] = false,
+        ["zhCN"] = "|cffff0000请注意:|r 任务提示插件之后的更新将会重置您的设定，建议您提前备份<Questie>相关的设定文档，注：备份魔兽世界安装版本游戏内的WTF文件夹就可以了。",
```
BreakBB, line 435: Should this stay as `<Questie>` @nanjuekaien1 or should it be replaced with just `Questie`?

nanjuekaien1:
The <> began to increase, I was ready to add eye-catching tips, and the online test found that the effect was not good, so I restored the previous method, which was omitted and I modified.

PR #5174, opened 2023-10-20
Add manual complete overrides for Argent Tournament marker quests

BreakBB:
## Issue references

Related to #5157

## Proposed changes

- There are some "marker quests" Blizzard uses for certain quest availability states. Instead of loading all complete quests from the API every time. 
- We manually mark these "marker quests" as complete on turn in/quest accept, because they also change on Blizzards side at these points.


Modules/Quest/QuestieQuest.lua
```diff
@@ -481,6 +491,9 @@ function QuestieQuest:AcceptQuest(questId)
     end
 end
 
+local allianceChampionMarkerQuests = { [13699] = true, [13713] = true, [13723] = true, [13724] = true, [13725] = true}
+local hordeChampionMarkerQuests = { [13726] = true, [13727] = true, [13728] = true, [13729] = true, [13731] = true}
```
BreakBB, line 495:
```suggestion
local allianceChampionMarkerQuests = {[13699] = true, [13713] = true, [13723] = true, [13724] = true, [13725] = true}
local hordeChampionMarkerQuests = {[13726] = true, [13727] = true, [13728] = true, [13729] = true, [13731] = true}
```

PR #5178, opened 2023-10-24
[feature] Settings refactor

BreakBB:
## Issue references

Fixes #

## Proposed changes

-
-


Modules/Quest/QuestieQuest.lua
```diff
@@ -161,7 +164,9 @@ function _QuestieQuest:ShowManualIcons()
 end
 
 function _QuestieQuest:HideQuestIcons()
-    WorldMapButton.UpdateText()
+    if (not Questie.db.profile.enabled) then
+        Questie_Toggle:SetText(l10n("Show Questie"));
+    end
```
BreakBB, line 169:
```suggestion
    WorldMapButton.UpdateText()
```

Modules/QuestieAnnounce.lua
```diff
@@ -104,15 +104,15 @@ end
...
+    if (not message) or alreadySentBandaid[message] or Questie.db.profile.questieShutUp then
         return
     end
 
     alreadySentBandaid[message] = true
 
     if IsInRaid() or IsInGroup() then
-        SendChatMessage(message, _QuestieAnnounce.GetChatMessageChannel())
-    elseif Questie.db.char.questAnnounceLocally == true then
+        SendChatMessage(message, (IsInRaid() and "RAID") or (IsInGroup() and "PARTY"))
```
BreakBB, line 114:
```suggestion
        SendChatMessage(message, _QuestieAnnounce.GetChatMessageChannel())
```

Modules/QuestieInit.lua
```diff
@@ -239,7 +239,35 @@ QuestieInit.Stages[3] = function() -- run as a coroutine
...
+            if AtlasToggleFromWorldMap and AtlasToggleFromWorldMap:IsShown() then -- #1498
+                AtlasToggleFromWorldMap:SetScript("OnHide", function() Questie_Toggle:SetPoint('RIGHT', WorldMapFrameCloseButton, 'LEFT', 0, 0) end)
+                AtlasToggleFromWorldMap:SetScript("OnShow", function() Questie_Toggle:SetPoint('RIGHT', AtlasToggleFromWorldMap, 'LEFT', 0, 0) end)
+                Questie_Toggle:SetPoint('RIGHT', AtlasToggleFromWorldMap, 'LEFT', 0, 0);
+            else
+                Questie_Toggle:SetPoint('RIGHT', WorldMapFrameCloseButton, 'LEFT', 0, 0);
+            end
+        end
+    end)
+
```
BreakBB, line 269:
```suggestion
```

This is all handled in the `WorldMapButton` module, which is initialized a bit below.

Modules/Options/IconsTab/QuestieOptionsIcons.lua
```diff
@@ -0,0 +1,932 @@
...
+---@type QuestiePlayer
+local QuestiePlayer = QuestieLoader:ImportModule("QuestiePlayer");
+---@type QuestieTooltips
+local QuestieTooltips = QuestieLoader:ImportModule("QuestieTooltips");
+---@type QuestieMenu
+local QuestieMenu = QuestieLoader:ImportModule("QuestieMenu");
+
+QuestieOptions.tabs.icons = {...}
+local optionsDefaults = QuestieOptionsDefaults:Load()
+
```
BreakBB, line 26:
```suggestion

local _GetIconTypes
local _GetIconTypesSort
```

Modules/Options/GeneralTab/QuestieOptionsGeneral.lua
```diff
@@ -15,27 +15,24 @@ local QuestieOptionsUtils = QuestieLoader:ImportModule("QuestieOptionsUtils");
...
-local _GetQuestSoundChoices
-local _GetQuestSoundChoicesSort
-local _GetObjectiveSoundChoices
-local _GetObjectiveSoundChoicesSort
-local _GetObjectiveProgressSoundChoices
-local _GetObjectiveProgressSoundChoicesSort
+local _GetAnnounceChannels
+local _IsAnnounceDisabled
 
 local iconsHidden = true
```
BreakBB, line 37: This should also be moved to `QuestieOptionsIcons`

Questie.lua
```diff
@@ -41,6 +42,14 @@ function Questie:OnDisable()
     end
 end
 
+function Questie:RefreshConfig(db, profile)
+    -- TODO: Migration here?
+    Questie:SetIcons()
+    QuestieQuest:SmoothReset()
+    TrackerBaseFrame:OnProfileChange()
+    print("Switched Profile!")
```
BreakBB, line 50: Do we want to replace this with a regular `Questie:Print` or remove entirely?

MyTechnoHunter: Yes!.. there should be _NO_ bare print calls in the code as we want to have the ability to tell em where to print not just have it willy-nilly.. allows better compat with chat addons

Muehe:
# General tab

- [x] Enable Instant Quest Text could be moved to Interface Options box below.
- [x] ~~Maybe add Blizzard quest tracker CVar there too.~~
    * Setting this cvar is all kinds of broken, don't touch.
- [ ] Move "Show Quest IDs" option from the Advanced tab there.
- [x] ~~Isle of Quel'Danas drop down is empty (on wotlk, not sure if this is expected)~~
- [x] Quest Level Options are mutually exclusive and could be turned into a drop down. All the options there need better descriptions.
    - They are technically related to available quest icons on the maps, so should probably be moved to the related settings in the Icons tab.
- [x] Questie Map Button is a bit too bold IMHO.
- [ ] Show Map Coordinates doesn't work on the small new map frame, probably needs same reparenting as the map button.
- [ ] Enable Minimap Coordinates text disappears instantly when toggled off, but toggling on only works after the player moves. :shrug: Should be renamed Show instead of Enable to match the other buttons there.
- [ ] In Social Options box, move the Disabled option out of the channel drop down and make it a checkbox.

# Icons tab

- [ ] Improve description of Hide icons of untracked quests option. Tracked where and how?
- [x] Move "Show custom quest frame icons" from Icons tab into Interface Options box in the General tab. Everything else in the Icons tab is related to map/minimap icons, so it seems a bit out of place.
- [x] Config Tacking Icons should be replaced with a box like everything else instead of opening the menu frame from the minimap button.
- [ ] Objective icon cluster amount could use a better description.
- [ ] The Icon Colors and Icon Glow options in the Map and Minimap boxes don't use the same colours. This should be changed or they should be made mutually exclusive, because as is they don't interact well.
- [ ] The Minimap Icon Fade Distance description is really confusing. It doesn't have a on/off toggle and amount slider like the option for fading icons close to player does.
- [ ] Icon Scales box is missing a size slider for talk objective icons.
- [ ] Replace pfQuest style toggle with something, e.g. opening the style selection window, or remove it.

# Tracker tab

- [ ] Create sub-groups where it makes sense (e.g. Header and Background options)
- [ ] Review option dependencies (e.g. some options are only available for certain settings in a drop-down, we have no good way to indicate this)
- [ ] Review option names and descriptions (e.g. replace "Enable Active Quests Header" with "Enable Questie Tracker Header", remove reference to variables saved per character from "Enable Tracker" description)
- [ ] Resize sliders and drop-downs where appropriate

# Nameplate tab

- [ ] This tab seems a bit weirdly named considering it is divided into settings for nameplate icons and for target frame icons.
- [ ] Toggling nameplate icons on doesn't draw immediately, toggling off works fine.

# DBM HUD tab

- [ ] Need to get DBM. :sweat_smile:

# Advanced tab

- [ ] Reorder the sections to put things more relevant to the user further up.

PR #5196, opened 2023-11-14
[feature] Tracker Expansion State Toggle Keybind

rsquires1988:
## Proposed changes

- Add Questie drop-down and Expand/Collapse Tracker Keybind to the Keybinds menu
- Also can collapse Tracker with 'Esc'
- Edit Questie-Classic/WOTLKC/BCC.toc to include Modules/Tracker/TrackerExpansionStateToggle.lua


rsquires1988:
I haven't tested this will some of the auto expand/collapse functionality already present turned on, not sure if it would break things there, but it shouldn't if the same expand/collapse functions are being called state-dependently.

The Esc key collapse functionality should also possibly be an option, although many addons just do it. Let me know what you think.

rsquires1988:
I didn't add it to the WOTLKC.toc as I haven't tested it there at all.  I can give it a shot later today and make a new commit to add it if all's well.

rsquires1988:
Neither are working, the expected Questie drop-down doesn't show up in keybinds.  I'm not sure what the issue is, since it works as a standalone add-on.

MyTechnoHunter:
> Neither are working, the expected Questie drop-down doesn't show up in keybinds. I'm not sure what the issue is, since it works as a standalone add-on.

not working in which flavor? era or wrath?

rsquires1988:
Both era and wrath.  There's probably something I'm missing to tie it in correctly while it's actually in the Questie folder, compared to having it as a standalone add-on.

rsquires1988:
Got it working, module needed to be created and imported.  Wrath doesn't like globals so had to do a weird wrapper to globalize the local, but it works in Classic and Wrath now!

BreakBB:
Hey @rsquires1988 sorry for the late response on this.

I want to make sure I understand where this idea is coming from:

You want to have a keybind to collapse/expand the Tracker to either show all tracked quests/achievements or only show the "Tracker header". Is that correct?

The current implementation listens to the escape key to toggles the Tracker state if I read that code correctly.

Does it ALWAYS toggle, whenever the button is pressed or are there any limitations? I mean there are certainly cases where you press that key (stop casting, closing some Questie-unrelated window, ...) but do NOT want the Tracker to toggle 🤔

rsquires1988:
The Escape key is actually secondary functionality I added later to fix an annoyance I was running into, explained below.  Bindings.xml is how the Questie keybinds section is added to WoW's built-in keybinds menu.  When a keybind is set in that menu, and pressed, G_TrackerExpansionStateToggle() is called, which causes the tracker to expand or collapse, state-dependently.  

Escape will only collapse the tracker if it is expanded.  I found myself hitting escape often to collapse the tracker, as most addon windows close with Escape, which Questie didn't support for the tracker.  This means when I want a completely cleared screen, I just spam Escape.  The escape press doesn't conflict with other addons' use of the key, nor the standard in game escape menu, as it's set up as an override.  I was at first concerned that this might conflict with other addons using the same method of overloading the Escape key, but in my testing, I haven't run into any errors or weird behavior.

I agree that the Escape key functionality may not something everyone wants, though.  I'll see if I can put it behind an option.  Lemme know if you think it should be on or off by default, although I personally lean towards on.  Could also maybe have it on when a keybind is set, off when not.

rsquires1988:
Sure enough, neither the keybind nor escape worked in combat, due to taint checks.  I just added a check for combat to prevent the error message output in chat.  Functionally, it works the same as before in this regard, just without the error message.

I added an option checkbox to the TrackerTab that turns on or off the Escape key functionality.

Tested and working in Wrath and Era!

BreakBB:
Thanks for coming back to this and explaining your goal in detail 👍🏻 

Yeah, the Tracker can not be collapsed/expanded in combat and iirc that is actually an API limitation :/

I will review your code and talk with the team if we add this now or after the next major release. The next big release will revamp the settings entirely, so your feature is somewhat _conflicting_ with the current state. We'll make sure to merge it though, no worries 😁 

Thank you already for taking the time to improve Questie, we appreciate it!

rsquires1988:
Cool!  Looking forward to it :grin:

rsquires1988:
> I haven't tested this with any of the automatic expand/collapse functionality that's already present, not sure if it would break things

Should've tested this sooner.

If the tracker is expanded by clicking, or probably any other way besides the keybind, the Escape key functionality doesn't work. Looking into it.

rsquires1988:
Got it mostly fixed, just need to figure out how to globalize it for the option checkbox.

rsquires1988:
Got it! Ended up having to do more or less a complete refactor, integrating it almost entirely into TrackerHeaderFrame.lua.  Still cleaning things up, but I'll push it soon.

rsquires1988:
Threw everything I could at it to try to get it to break again, but this time I think it's good to go! Tested in Wrath and Classic.

rsquires1988:
So it works as is, but I've got a small problem with it. If other settings Windows or frames that close with Escape are opened after the tracker is expanded, the tracker will close on the first escape press. This is the consequence of using an override for escape. Ideally, I'd be adding the frame to UISpecialFrames and handling it that way, and I've been working on that, but it introduces its own problems. I'll keep you posted.

rsquires1988:
After many attempts, I'm not sure fixing that problem is possible due to the difficulty of determining whether something in UISpecialFrames is actually an escape-closable window.  I CAN have the tracker collapse on escape along with any other open windows, and it is more performant than the way I was doing it in the current code with SetOverrideBinding, but that doesn't really fix the problem.  Either way, the tracker will be closed when it doesn't feel like it should be.  It seems like there should be a way to do this, but I'm stumped at the moment.

MyTechnoHunter:
few things i noticed.. you replaced the click handlers for the header itself (manual click handling to collapse or expand..) you also removed the update at the bottom of the function..

rsquires1988:
I did, the code for setting the mode and updating the tracker has been moved to ToggleTrackerState() and tweaked a bit, although it does the same thing it did before.  This isn't necessary for anything I did to work, it was more so I could call the toggle functionality from elsewhere while I was working on it.  Clicking to collapse or expand works as it did before, but I can change it back for safety.

BreakBB:
Hey @rsquires1988 just a headsup: SoD release needs more time, than we expected (because of all the new quests, items, ... they added.

> So it works as is, but I've got a problem with it. If other settings windows or frames that close with Escape are opened after the tracker is expanded, the tracker will close on the first escape press, and not the other windows as I would like.

This indeed doesn't sound good. Since I didn't test your changes yet, I can't tell if there is a way to fix this.

rsquires1988:
I haven't had a chance to work on this again yet, but I have a plan of attack, I think the answer is probably a combination of overrides set at strategic times and UISpecialFrames.

Issue #5199, opened 2023-11-19
Custom translation support


Jakanis:
There's several addons that translate WoW to some languages (Polish, Ukrainian, Czech and some more)
It would be awesome if we could use these translations in Questie in a way similar to Auctionator - https://www.curseforge.com/wow/addons/turkish-translations-module-for-auctionator
So we'd create addon that contains our version of your lookups and translation folders and add more languages to Questie.
Could we have something like that?

Muehe:
It is technically possible to refactor the translation module (called l10n) to support more languages and it shouldn't be all that hard to do. But there are currently around 2000 strings only for the Questie UI to be translated. And that doesn't include quests, items, and so forth, for which we currently don't have a way of overwriting them in the default Blizzard UI. So refactoring that would be a lot harder, and those strings would number in the tens of thousands probably.

I wouldn't be opposed to the change if anybody wants to work on it, but it's not really a priority for us right now.

That said if you want to do it regardless of the problems outlined above we can provide guidance on how to go about it here or on our Discord.

Jakanis:
Well, we already try to handle default Blizzard UI, and so do other translators, so no worries about that :) 
We just need to translate strings used by Questie. And, if I understood correctly, these took from 'Localization' folder, which we could translate.

PR #5204, opened 2023-11-27
[feature] Custom localization support

Jakanis:
## Issue references

Feature request #5199

## Proposed changes

- Add support for addons with custom localization. 
  - Similar to how it works with Auctionator - use global `QUESTIE_LOCALES_OVERRIDE` variable with translation dictionaries
- Add translation for event in chat message (screenshot 2)
- Change locale dropdown behaviour
  - Choose custom translation on 'auto' setting if present
  - Recompile database only if locale changed
- Change DB recompilation condition check and `dbCompiledLang` setting
  - Used locale value from `l10n`
  - Won't recompile if chosen same locale on auto
  - Will recompile if custom localization addon disabled/deleted

## Known issues
- Default font doesn't support Cyrillic glyphs on non-cyryllic client 
  - LibSharedMedia sets cyryllic version of Friz Quadrata TT only for ruRU client
- Zone names/categories in tracker aren't transtaled
  - lookupZones.lua and lookupQuestCategories.lua have no localized version
- Quest objectives aren't fully translated in tooltips and tracker (only items/NPCs icons with shift key and quest icons, screenshots 3, 4, 5)
  - It's still stored in Questie DB, but I guess it's too deep for me yet

## Screenshots
Used [this addon](https://github.com/Jakanis/QuestieUkrainianTranslation) (not finished yet)
image
image
image
image
Only 'Diseased Wolf Pelt' here, NPC name handled by localization addon:
image


Jakanis:
Well, first one from me, so hello :)
Despite problems with objectives translation still looks cool. Especially with amount of translated quest and townfolks icons on map.
I hope it's acceptable :D

BreakBB:
Hey @Jakanis 

thank you very much for taking the time to look into this and also providing a PR for this! At a first glance, this looks great 🥳 

After the first days of the SoD release we'll look into this, to ship it in a later release!

Localization/l10n.lua
```diff
@@ -30,6 +30,22 @@ local supportedLocals = {
     ['koKR'] = true,
 }
 
+function l10n:InitializeLocaleOverride()
+    local overridingLocale = QUESTIE_LOCALES_OVERRIDE.locale
```
BreakBB, line 34: `QUESTIE_LOCALES_OVERRIDE` needs to be added to `.luacheckrc` otherwise other pipeline will break, as this is an "unknown global"

BreakBB (changes requested):
Hey @Jakanis this is a really great PR! 👍🏻 

Only a small thing, otherwise this looks fine. Can you take care of it?

Also I think it is worth it to mention this nice feature in the README and maybe point to the place where one can get values for `QUESTIE_LOCALES_OVERRIDE`.

What do you think, can you take care of it?

Localization/l10n.lua
Jakanis: Thanks, added

Jakanis:
Thanks, @BreakBB 

Added page on wiki and link to that in sidebar — [Localization to more languages](https://github.com/Questie/Questie/wiki/Localization-to-more-languages). Will add section with localization addons later, if you don't mind.
There's already a link to 'Contributing' page in README, so it can be found quite easily. And I guess I'll inform most localization groups anyway, so they won't miss that. :)

BreakBB:
@Jakanis thank you 👍🏻 

Feel free to also add a link pointing to these localization groups 😁

Modules/Options/AdvancedTab/QuestieOptionsAdvanced.lua
```diff
@@ -411,3 +404,23 @@ function QuestieOptions.tabs.advanced:Initialize()
         },
     }
 end
+
+_GetLanguages = function()
```
BreakBB, line 408: `_GetLanguages` is not defined yet and therefore a global. Just add a `local _GetLanguages` above `QuestieOptions.tabs.advanced:Initialize()`

Jakanis: Thanks, fixed

PR #5227, opened 2023-12-02
feat: provide french translation for file ClassicObjectives.lua

nina-alin:
## Proposed changes

- Provide French Translation for the file ClassicObjectives.lua


BreakBB:
Wow thank you very much @nina-alin for taking the time to translate these 🥳

PR #5292, opened 2023-12-14
Warlock Lake of Fire rune

EmilLindegaard:
Let me know if there should be any changes :)


Database/Corrections/sodQuestFixes.lua
```diff
@@ -666,6 +666,16 @@ function SeasonOfDiscovery:LoadQuests()
             [questKeys.requiredSpell] = -410096,
             [questKeys.zoneOrSort] = sortKeys.Shaman,
         },
+        [90032] = {
+            [questKeys.name] = "Lake Of Fire",
+            [questKeys.startedBy] = {{3537}},
+            [questKeys.finishedBy] = {{414658}},
```
BreakBB, line 672: A note on this: Since it is a "fake quest" (a quest that never is in the quest log of a player), Questie won't show the finisher (since the quest will never be "accepted" and also never will be "ready for turn in").

So basically this field is irrelevant.

BUT it is good you have the ID of what is also related to this rune.

Therefore I suggest to add another fake quest and use this ID as starter. Give the quest the same name, but maybe change the objective text to make it more clear what to do at the different points.

I think I did the same for the "Enlightment" rune for Mages, which you get in Duskwood, but also need to visit two places without any quest.

Besides that: Your PR looks great, thank you already!

EmilLindegaard: Ill look into it tomorrow 😊

EmilLindegaard: Oh and a couple of follow-up questions:
1. For Enlightenment you put in [questKeys.requiredSpell] = -415942. Is this a way to make sure the fake quest "disappears" from the map/minimap once it is acquired? If so I'll do the same.
2. Do we have a good approach for selecting requiredLevel and questLevel? Many runes are not related to quests so they can be potentially be obtained at any level (even though it may involve navigating in higher level zones). IMO it makes sense to put requiredLevel to 1 and set an approximate questLevel depending on the required activity (e.g. navigating in an area with lvl 22 mobs in the case of Lake of Fire)

EmilLindegaard: Also maybe we should consider showing these fake rune quests regardless of player level. But maybe you guys already implemented it this way or for some reason decided not to?

EmilLindegaard:
For some reason I cannot get the other Lake of Fire fake quest to show up on the map. So not sure if working yet..

BreakBB:
> For some reason I cannot get the other Lake of Fire fake quest to show up on the map. So not sure if working yet..

Did you recompile the database after the change?

EmilLindegaard:
Yes I recompiled numerous times trying out different things to see if I could fix it. No luck so far. I was able to connect it to "Scaber Stalk" object found in Darkshore just fine.

BreakBB:
Okay, so there is something wrong with the actual object you want to connect it to?
Just a small brain dump:

- Verify that object is in the sodBaseObjects file
- Verify the object data looks fine ingame (enable Debug and use the "Advanced Search" of Questie and check what the raw data looks like)
- Verify the fake quest data looks fine ingame (same as before)
- Verify the spellId to hide it is correct and your character does not know the spell already

EmilLindegaard:
I can see both the object 414658 and quest 90033 in the advanced search. I can also see that object starts the quest.
If i press "Show On Map" on the object I can also point on the object on the map and see that it starts the quest.
Futhermore when I inspect the quest i can see that I am eligible in-game to receive it.
Still no yellow question mark on the map to show it though.
Will try to inspect further tomorrow.

SpellId is not learnt 

Thank you for your help btw :)

BreakBB:
> Still no yellow question mark on the map to show it though.

If you enable the "Show Seasonal Runes" option these quests show as rune icon, not as `!`. But I guess since you know where to look you would have noticed.

> I haven't visited the object in-game so I am not sure if it an interactable object (right click-able) or only an object you can hover your cursor over. But maybe this is irrelevant anyway?

Yeah that irrelevant. As long as the object has a valid spawn it shows.

I won't have much time today, but will try to checkout your branch myself as soon as I can 👍🏻

EmilLindegaard:
That solved it! I didn't know you had to enable the option to see runes!

FYI if a NPC connected to a path (i.e. a shown as a yellow line across the map), the rune will show regardless of "Show Seasonal Runes" is enabled or not. I guess this is not intended?
The first rune I added was related to a pathing NPC and that is what lead to my confusion :)

EmilLindegaard:
Btw I thought about modifying the # [questKeys.requiredSourceItems] to be able to interpret negative IDs similarly to [questKeys.requiredSpell]. So if you have to get a items from mobs that have a lot of different spawn points, you get rid of the icons once you obtain the required item (An example is the rune warlock rune Master Channeler in Westfall where you have to get items from different mobs leading to many icons.)

BreakBB:
> FYI if a NPC connected to a path (i.e. is shown as a yellow line across the map), the rune will show regardless of "Show Seasonal Runes" is enabled or not. I guess this is not intended?

This is not intended and on our list already to be fixed 🙈

PR #5332, opened 2023-12-20
ADDED: 'Wrangling a Wild Wyvern' Quest to QuestDB

### LINKED ISSUE #5331: ADDED: 'Wild Wyvern Wrangling' Missing From QuestDB
## Description

The SoD quest 'Wild Wyvern Wrangling' is not currently in the QuestDB. 

## Wowhead link

https://www.wowhead.com/classic/quest=78121/wrangling-a-wild-wyvern

## Screenshots

N/a

## Questie & WoW version

v9.2.9
WoW Classic SoD - v1.15.052409

Sommos:
## Issue references

https://github.com/Questie/Questie/issues/5331

Fixes #5331

## Proposed changes

- ADDED: 'Wrangling a Wild Wyvern' is now in the QuestDB, and should be accessible to users.


Database/Corrections/sodQuestFixes.lua
```diff
@@ -140,6 +140,11 @@ function SeasonOfDiscovery:LoadQuests()
             [questKeys.requiredSpell] = -410111,
```
BreakBB, line 140: I just noticed that this is totally not required, since this is a real quest and not one of the "fake rune quests" we are also adding to show runes in SoD 😅 The API knows about this questId and will return `true` when using `C_QuestLog.IsQuestFlaggedCompleted(78114)` and therefore hides the quest if a player already completed it.

So feel free to remove this field for quest `78114` and `78121`.

Database/Corrections/sodQuestFixes.lua
```diff
@@ -140,6 +140,11 @@ function SeasonOfDiscovery:LoadQuests()
...
             [questKeys.zoneOrSort] = sortKeys.HUNTER,
         },
+        [78121] = {
```
BreakBB, line 143: What this quest also needs is the `preQuestSingle` field, which hides this quest until the quest in `preQuestSingle` is turned in. That field is a table of questIds and if one of the questIds is already complete, then the quest will show.

There is also `preQuestGroup` which requires ALL quests of that table to be complete in order to show the quest. But if there is just a single pre quest, we are using `preQuestSingle`.

BreakBB:
Also generally: Since these quests have a `requiredLevel = 1` you can always create a level 1 character and see if the quests show up properly.

Database/Corrections/sodQuestFixes.lua
Sommos: Fixed in: https://github.com/Questie/Questie/pull/5332/commits/6e704ae1a13649bfa253ed3bfb3358d0fe7744ad

Database/Corrections/sodQuestFixes.lua
Sommos: https://github.com/Questie/Questie/pull/5332/commits/6e704ae1a13649bfa253ed3bfb3358d0fe7744ad

Issue #5326, opened 2023-12-20
attempt to index global 'C_Calendar' (a nil value) on 1.14.2


trcrsired:
C_Calendar does not exist in 1.14.2. please add some workaround
```
Message: ...AddOns\Questie\Database\Corrections\QuestieEvent.lua:147: attempt to index global 'C_Calendar' (a nil value)
Time: Tue Dec 19 23:05:19 2023
Count: 1
Stack: ...AddOns\Questie\Database\Corrections\QuestieEvent.lua:147: attempt to index global 'C_Calendar' (a nil value)
[string "=[C]"]: ?
[string "@Interface\AddOns\Questie\Database\Corrections\QuestieEvent.lua"]:147: in function <...AddOns\Questie\Database\Corrections\QuestieEvent.lua:146>
[string "@Interface\AddOns\Questie\Database\Corrections\QuestieEvent.lua"]:199: in function <...AddOns\Questie\Database\Corrections\QuestieEvent.lua:196>
[string "@Interface\AddOns\Questie\Database\Corrections\QuestieEvent.lua"]:137: in function `Load'
[string "@Interface\AddOns\Questie\Database\Corrections\QuestieCorrections.lua"]:216: in function <...\Questie\Database\Corrections\QuestieCorrections.lua:212>

Locals: 
```
![image](https://github.com/Questie/Questie/assets/100043421/4ba6dadd-539b-4432-a872-91d5f34e8711)

trcrsired:
Here is another one
```
Message: ...face\AddOns\Questie\Modules\Tracker\TrackerUtils.lua:318: attempt to call global 'GetQuestLogCompletionText' (a nil value)
Time: Tue Dec 19 23:16:00 2023
Count: 1
Stack: ...face\AddOns\Questie\Modules\Tracker\TrackerUtils.lua:318: attempt to call global 'GetQuestLogCompletionText' (a nil value)
[string "=[C]"]: in function `GetQuestLogCompletionText'
[string "@Interface\AddOns\Questie\Modules\Tracker\TrackerUtils.lua"]:318: in function `GetCompletionText'
[string "@Interface\AddOns\Questie\Modules\Tracker\QuestieTracker.lua"]:725: in function `_UpdateQuests'
[string "@Interface\AddOns\Questie\Modules\Tracker\QuestieTracker.lua"]:1632: in function `Update'
[...]

PR #5340, opened 2023-12-21
Always return false when C_Calender does not exist for darkmoon faire

trcrsired:
Return false for darkmoonfaireactive when C_Calender is null

## Issue references

Fixes #

#5326 

## Proposed changes
always return false when C_Calender does not exist
-
-

## Screenshots

No errors any more
image

### DIFF
--- a/Database/Corrections/QuestieEvent.lua
+++ b/Database/Corrections/QuestieEvent.lua
@@ -144,6 +144,10 @@ end
 ---@param dayOfMonth number
 ---@return boolean
 _IsDarkmoonFaireActive = function(dayOfMonth)
+    local C_Calendar = C_Calendar
+    if C_Calendar == nil then
+        return false
+    end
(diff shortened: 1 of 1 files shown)


Database/Corrections/QuestieEvent.lua
```diff
@@ -144,6 +144,10 @@ end
 ---@param dayOfMonth number
 ---@return boolean
 _IsDarkmoonFaireActive = function(dayOfMonth)
+    local C_Calendar = C_Calendar
+    if C_Calendar == nil then
+        return false
+    end
```
BreakBB, line 150:
```suggestion
    if C_Calendar == nil then
        -- This is a band aid fix for private servers which do not support the `C_Calendar` API.
        -- They won't see Darkmoon Faire quests, but that's the price to pay.
        return false
    end
```

trcrsired:
ty ty

PR #5341, opened 2023-12-21
Add Priest "Void Plague" rune

EmilLindegaard:
Going to be adding some Priest Runes. Probably in a slower pace and more PRs than the Warlock runes :)

Opted for combining some of the "kill this mob" runes for each faction since I figured it does really give a better player experience to state the mob name specifically. Let me know if you disagree.


Database/Corrections/sodQuestFixes.lua
```diff
@@ -1468,6 +1468,42 @@ function SeasonOfDiscovery:LoadQuests()
...
         },
+        [90084] = {
+            [questKeys.name] = "Void Plague",
+            [questKeys.startedBy] = {{327,1211}},
+            [questKeys.finishedBy] = nil,
+            [questKeys.requiredLevel] = 1,
+            [questKeys.questLevel] = 8,
+            [questKeys.requiredRaces] = raceIDs.ALL_ALLIANCE,
+            [questKeys.requiredClasses] = classIDs.PRIEST,
+            [questKeys.objectivesText] = {"Defeat him"},
```
BreakBB, line 1479:
```suggestion
            [questKeys.objectivesText] = {"Defeat him."},
```

Database/Corrections/sodQuestFixes.lua
```diff
@@ -1468,6 +1468,42 @@ function SeasonOfDiscovery:LoadQuests()
...
+        },
+        [90085] = {
+            [questKeys.name] = "Void Plague",
+            [questKeys.startedBy] = {nil,{407734}},
+            [questKeys.finishedBy] = nil,
+            [questKeys.requiredLevel] = 1,
+            [questKeys.questLevel] = 8,
+            [questKeys.requiredRaces] = raceIDs.ALL_ALLIANCE,
+            [questKeys.requiredClasses] = classIDs.PRIEST,
+            [questKeys.objectivesText] = {"Loot Gnarlpine Cache in Ban'ethil Barrow Den"},
```
BreakBB, line 1491:
```suggestion
            [questKeys.objectivesText] = {"Loot Gnarlpine Cache in Ban'ethil Barrow Den."},
```

Database/Corrections/sodQuestFixes.lua
```diff
@@ -1468,6 +1468,42 @@ function SeasonOfDiscovery:LoadQuests()
...
+        },
+        [90086] = {
+            [questKeys.name] = "Void Plague",
+            [questKeys.startedBy] = {{3128,3129,208196}},
+            [questKeys.finishedBy] = nil,
+            [questKeys.requiredLevel] = 1,
+            [questKeys.questLevel] = 8,
+            [questKeys.requiredRaces] = raceIDs.ALL_HORDE,
+            [questKeys.requiredClasses] = classIDs.PRIEST,
+            [questKeys.objectivesText] = {"Defeat him"},
```
BreakBB, line 1503:
```suggestion
            [questKeys.objectivesText] = {"Defeat him."},
```

BreakBB (changes requested):
Most of the NPCs which drop this rune have quite a few spawns and would flood the map with icons. So we should hide these quests for now.

I am already thinking about a way to tackle this general problem.

EmilLindegaard:
I didn't think this rune was too bad, but I agree that it would be better with only 1 rune marker.

One option would be to add a fake NPC spawn to attach the quest to at the respective positions but this implies more manual work

BreakBB:
> One option would be to add a fake NPC with the same name but only 1 spawn-point to attach the quest to at the respective positions but this implies more manual work

That is also at the current state my preferred solution. That will mean though, that the player won't see any tooltips when hovering the mobs.

EmilLindegaard:
I guess it either that or create more abstraction for runes.

I means for example implement so, it will show tooltips of the mobs in the QuestKeys.startedBy but not map icons. Then add a new property related to map icon coordinates to separate the two. Probably also not a straight-forward solution since it requires separating quests from runes.

BreakBB:
Yeah, we'll figure something out.

Btw, I saw you started using the `[quest]` prefix. Keep in mind though, that every commit with such a prefix will be added to the changelog. So for the PR history it would have been good to only add it one or two commits, which should show up in the changelog and not on every change you do on this PR. Otherwise I'll squash them all to a single commit.

EmilLindegaard:
> Yeah, we'll figure something out.
> 
> 
> 
> Btw, I saw you started using the `[quest]` prefix. Keep in mind though, that every commit with such a prefix will be added to the changelog. So for the PR history it would have been good to only add it one or two commits, which should show up in the changelog and not on every change you do on this PR. Otherwise I'll squash them all to a single commit.

I'll keep that in mind next time

PR #5349, opened 2023-12-22
Add "Shifting Scale Talisman" quest

EmilLindegaard:
## Issue references

Fixes #

## Proposed changes

- Add "Shifting Scale Talisman" quest

Not related to any rune.  
Used the quest_data.lua generated from the scraper.
Added required source item from BFD to start the quest
Put ratchet as the zone similarly to other ratchet related quests

### DIFF
--- a/Database/Corrections/sodQuestFixes.lua
+++ b/Database/Corrections/sodQuestFixes.lua
@@ -280,6 +280,7 @@ function SeasonOfDiscovery:LoadQuests()
             [questKeys.specialFlags] = specialFlags.REPEATABLE,
         },
         [78909] = {
+            [questKeys.objectivesText] = {"Tip: You need to loot Handful of Shifting Scales from Old Serra'kis in Blackfathom Deeps before you can start this quest.\n\nBring N'ora Anyheart the following materials alongside the Handful of Shifting Scales to complete the talisman: A satchel of Strange Dust, five Greater Magic Essence, 2 Small Glimmering Shards and five gold pieces."},
             [questKeys.zoneOrSort] = zoneIDs.RATCHET,
             [questKeys.questFlags] = questFlags.RAID,
         },
--- a/Database/Corrections/sodQuestFixes.lua
+++ b/Database/Corrections/sodQuestFixes.lua
@@ -280,7 +280,7 @@ function SeasonOfDiscovery:LoadQuests()
             [questKeys.specialFlags] = specialFlags.REPEATABLE,
         },
         [78909] = {
-            [questKeys.objectivesText] = {"Tip: You need to loot Handful of Shifting Scales from Old Serra'kis in Blackfathom Deeps before you can start this quest.\n\nBring N'ora Anyheart the following materials alongside the Handful of Shifting Scales to complete the talisman: A satchel of Strange Dust, five Greater Magic Essence, 2 Small Glimmering Shards and five gold pieces."},
+            [questKeys.objectivesText] = {"Tip: You need to loot Handful of Shifting Scales from Old Serra'kis in Blackfathom Deeps before you can start this quest.","","Bring N'ora Anyheart the following materials alongside the Handful of Shifting Scales to complete the talisman: A satchel of Strange Dust, five Greater Magic Essence, 2 Small Glimmering Shards and five gold pieces."},
             [questKeys.zoneOrSort] = zoneIDs.RATCHET,
             [questKeys.questFlags] = questFlags.RAID,
         },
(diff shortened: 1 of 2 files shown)


Database/Corrections/sodQuestFixes.lua
```diff
@@ -279,6 +279,10 @@ function SeasonOfDiscovery:LoadQuests()
             [questKeys.startedBy] = {{213077,214070,214096,214098,214099,214101,}},
             [questKeys.specialFlags] = specialFlags.REPEATABLE,
         },
+        [78909] = {
+            [questKeys.requiredSourceItems] = {211419},
```
BreakBB, line 283: What is that for? This item is an objective so it should already show on the map 🤔

Database/Corrections/sodQuestFixes.lua
```diff
@@ -279,6 +279,10 @@ function SeasonOfDiscovery:LoadQuests()
...
+            [questKeys.zoneOrSort] = zoneIDs.RATCHET,
```
BreakBB, line 284:
```suggestion
            [questKeys.zoneOrSort] = zoneIDs.RATCHET,
            [questKeys.questFlags] = questFlags.RAID,
```

Database/Corrections/sodQuestFixes.lua
EmilLindegaard: You need to get the drop 211419 in order to be able to start the dialogue and pick up the quests (where the object is also an objective in the quest)
https://www.wowhead.com/classic/item=211419/handful-of-shifting-scales

BreakBB: I see, thanks for pointing out.

Well that field is used for items you need to collect while on a quest. Not as a real objective but for certain steps.

Not sure what's the best case to handle this case. The "pre quest" is looting the item, while the item does not start the quest directly 🤔

@Cabro do you have an idea? Maybe a fake quest which we use as pre quest and mark is as exclusiveTo?

Cabro: This is exactly the case for TF bindings quest. There is a part in code where we look for "looted" strings. Maybe we can do another if case to check the id of the item we looted, and then just make the respective quest visible?

BreakBB: > This is exactly the case for TF bindings quest. There is a part in code where we look for "looted" strings. Maybe we can do another if case to check the id of the item we looted, and then just make the respective quest visible?

There is? 🤔 Do you know where? I didn't find anything after a quick search.

EmilLindegaard: The check Cabro mentions should ideally be found in `function _QuestieQuest:DrawAvailableQuest(quest)` in QuestieQuest.lua or in `function QuestieDB.IsDoable(questId, debugPrint, returnText, returnBrief)` in QuestieDB.lua right?

Cabro: > > This is exactly the case for TF bindings quest. There is a part in code where we look for "looted" strings. Maybe we can do another if case to check the id of the item we looted, and then just make the respective quest visible?
> 
> There is? 🤔 Do you know where? I didn't find anything after a quick search.

QuestieTracker.lua:431

BreakBB:
I have to look into this again.

Database/Corrections/sodQuestFixes.lua
BreakBB: Okay so here is my idea:

For now lets just update the objective text for that quest to explain that you need to Scale in order to accept the quest.

I created #5385 to keep track of a better idea.

EmilLindegaard:
Suggestion for a modified quest objective text..

But why are the newlines ending up being formatted like this? 

image

Database/Corrections/sodQuestFixes.lua
```diff
@@ -279,6 +279,11 @@ function SeasonOfDiscovery:LoadQuests()
             [questKeys.startedBy] = {{213077,214070,214096,214098,214099,214101,}},
             [questKeys.specialFlags] = specialFlags.REPEATABLE,
         },
+        [78909] = {
+            [questKeys.objectivesText] = {"Tip: You need to loot Handful of Shifting Scales from Old Serra'kis in Blackfathom Deeps before you can start this quest.\n\nBring N'ora Anyheart the following materials alongside the Handful of Shifting Scales to complete the talisman: A satchel of Strange Dust, five Greater Magic Essence, 2 Small Glimmering Shards and five gold pieces."},
```
BreakBB, line 283: I know this is not intuitive, but blank lines are added with `""`. The objectivesText is a table and each entry will be added as line.

```suggestion
            [questKeys.objectivesText] = {"Tip: You need to loot Handful of Shifting Scales from Old Serra'kis in Blackfathom Deeps before you can start this quest.","","Bring N'ora Anyheart the following materials alongside the Handful of Shifting Scales to complete the talisman: A satchel of Strange Dust, five Greater Magic Essence, 2 Small Glimmering Shards and five gold pieces."},
```

EmilLindegaard: I see :)

PR #5381, opened 2023-12-29
[quest] Fix reputation gain from "The Crown of Will"

danez:
## Issue references

Fixes #

## Proposed changes

- The quest gives reputation with undercity not Orgrimmar, see also https://www.wowhead.com/classic/quest=518/the-crown-of-will
-

## Screenshots
Screenshot 2023-12-29 at 13 27 17

### DIFF
--- a/Database/Corrections/classicQuestFixes.lua
+++ b/Database/Corrections/classicQuestFixes.lua
@@ -262,6 +262,7 @@ function QuestieQuestFixes:Load()
         },
         [518] = {
             [questKeys.preQuestSingle] = {},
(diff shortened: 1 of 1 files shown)


danez:
Okay I changed it.

BreakBB:
Awesome, thank you!

danez:
To be fair though, most of the reputation on quests is wrong. At first I thought it is due to level scaling, but even when the quest is my level it is usually wrong.

BreakBB:
Yeah, you are absolutely right. There are lot of vanilla quests with wrong reputation values 🙈 

I opened up an issue, maybe we'll have time for it #5384

PR #5387, opened 2023-12-30
[bug] Add missing constant for undercity and replace more IDs with constants

danez:
I noticed that the constant I used in my previous PR does not even exist. 
 https://github.com/Questie/Questie/blob/master/Database/questDB.lua#L115

As the file where `factionIDs` are defined (right now only darkmoon) is not game-version specific, I wasn't sure if classic factions should be added there? I haven't played retail in a long time and I'm not sure if factions work the same through all versions?

So for now I reverted to the integer value for Undercity.

ref #5381

### DIFF
--- a/Database/Corrections/tbcQuestFixes.lua
+++ b/Database/Corrections/tbcQuestFixes.lua
@@ -24,6 +24,7 @@ function QuestieTBCQuestFixes:Load()
     local questKeys = QuestieDB.questKeys
     local raceIDs = QuestieDB.raceKeys
     local classIDs = QuestieDB.classKeys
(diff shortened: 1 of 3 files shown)


BreakBB:
Uff, I thought I already added more to that enum 🤦🏻 

Yeah, the factions are all the same. [These are the faction IDs](https://wowpedia.fandom.com/wiki/FactionID)

danez:
I added undercity and also replaced all other Ids with their constant

BreakBB:
Thanks a lot @danez for going the extra mile on this! We appreciate it ❤️

PR #5398, opened 2024-01-01
Fix reputation rewards of all classic quests

### LINKED ISSUE #5384: Wrong reputation for many quests
## Description

As stated [here](https://github.com/Questie/Questie/pull/5381#issuecomment-1872174369) there are quite a few vanilla quests where reputation reward is incorrect. Either the quest grants reputation for another faction or the value is wrong or both.

We might be able to use the scraper to fetch these values.

## Questie & WoW version

v9.2.11

danez:
## Issue references

Fixes #5384 

## Proposed changes

- This adds a new scraper that gets the reputation from wowhead and generates a correction file
- Added the correction file to the extension

Things to note:

- most changes are the reputation values that change
- some quests changed from for example `+100 Orgrimmar | +100 Trolls | +100 Undercity | +100 Thunder Bluff` to `+100 Horde` and similar for Alliance. In the game when doing such a quest you also actually see in the log `Gained +100 Reputation with Horde.` So I guess this is fine.
- Some quests are weird, for example [8857](https://www.wowhead.com/classic/quest=8857/secrets-of-the-colossus-ashi). It says that you receive 4 times +200 reputation from the same faction. I guess this is a mistake on wowheads data as I've never seen getting multiple reputation rewards from one quest for one faction. I haven't changed anything about this yet. So should I normalize this data so that `4x +200 BootyBay` becomes `1x +200 BootyBay`?
- Some rewards are now in different order, like quest 834. Should I keep them as the come from wowhead or should I sort them by factionID?
- I did not use constants in the generated file for the factions. Should I do that?
- Quests 2358 and 9378 are not found on wowhead, but these quests are on the ignore list in questie anyway.
- Some inline comments

[Diff of all changes](https://gist.github.com/danez/0730d45fa9eb1f922d89a817f8a2ed85)

[Diff of important changes (ignores changes to reputation values)](https://gist.github.com/danez/d9b69881844628ba03503399061f6762)


Database/Corrections/classicQuestFixes.lua
```diff
@@ -15,7 +15,7 @@ local l10n = QuestieLoader:ImportModule("l10n")
 -- Further information on how to use this can be found at the wiki
 -- https://github.com/Questie/Questie/wiki/Corrections
 
-function QuestieQuestFixes:Load()
+function QuestieQuestFixes:LoadMissingQuests()
```
danez, line 18: I separated this, because the order i now want is:

1. add the missing indices (`QuestieQuestFixes:LoadMissingQuests()`)
2. load reputation fixes from wowhead
3. load corrections (`QuestieQuestFixes:Load()`)

This is so that the manual corrections can still overwrite the automated ones

Database/Corrections/classicQuestFixes.lua
```diff
@@ -262,7 +264,6 @@ function QuestieQuestFixes:Load()
         },
         [518] = {
             [questKeys.preQuestSingle] = {},
-            [questKeys.reputationReward] = {{factionIDs.UNDERCITY,100}}
```
danez, line 265: All the removals here are now part of the automatically generated file.

ExternalScripts(DONOTINCLUDEINRELEASE)/scraper/quest/quest_formatter.py
```diff
@@ -35,7 +35,7 @@ def __load_json_file(self, file_name: str):
         return filtered_sorted_data
 
     def __sort_and_filter_data(self, data):
-        sorted_data = sorted(data, key=lambda x: x.get('questId', 0))
+        sorted_data = sorted(data, key=lambda x: int(x.get('questId', 0)))
```
danez, line 38: This changes the sorting so it sorts numbers as numbers like `1,2,3,10,20,100,200` and not as strings `1, 10, 100, 2, 20, 200, 3, ...`

ExternalScripts(DONOTINCLUDEINRELEASE)/scraper/quest/quest_spider.py
```diff
@@ -17,14 +18,22 @@ def __init__(self) -> None:
...
     def parse(self, response):
+        if response.url.startswith('https://www.wowhead.com/classic/quests?notFound='):
+            questID = re.search(r'https://www.wowhead.com/classic/quests\?notFound=(\d+)', response.url).group(1)
+            logging.warning('\x1b[31;20mQuest with ID {questID} not found\x1b[0m'.format(questID=questID))
+            return None
+
         result = {}
         for script in response.xpath('//script/text()').extract():
-            if script.startswith('//<![CDATA[\nWH.Gatherer.addData') and script.endswith('//]]>'):
+            if script.startswith('//<![CDATA[\nWH.Gatherer.addData') and script.find('$.extend(g_quests') >= 0 and script.endswith('//]]>'):
```
danez, line 28: This additional check ensures we look at the right script tag. Sometimes multiple script tags match the other two checks and then the script fails, because it cannot find the questId.

ExternalScripts(DONOTINCLUDEINRELEASE)/scraper/quest/quest_spider.py
```diff
@@ -17,14 +18,23 @@ def __init__(self) -> None:
         self.start_urls = [self.base_url_classic.format(quest_id) for quest_id in QUEST_IDS]
 
     def parse(self, response):
+        if response.url.startswith('https://www.wowhead.com/classic/quests?notFound='):
+            questID = re.search(r'https://www.wowhead.com/classic/quests\?notFound=(\d+)', response.url).group(1)
+            logging.warning('\x1b[31;20mQuest with ID {questID} not found\x1b[0m'.format(questID=questID))
+            return None
```
danez, line 24: Added this check to see when quests are not found, instead of silently ignoring it.

ExternalScripts(DONOTINCLUDEINRELEASE)/scraper/quest/classic/quest_ids.py
Logonz, line 1: I am thinking about if we should instead use something like this:
https://www.wowhead.com/classic/sitemap/quest?page=1
To get all the QuestIDs rather than hardcoding it like this?

Logonz:
"Some quests are weird, for example [8857](https://www.wowhead.com/classic/quest=8857/secrets-of-the-colossus-ashi). It says that you receive 4 times +200 reputation from the same faction. I guess this is a mistake on wowheads data as I've never seen getting multiple reputation rewards from one quest for one faction. I haven't changed anything about this yet. So should I normalize this data so that 4x +200 BootyBay becomes 1x +200 BootyBay?"

I think normalize them to 1x +200 Booty Bay, maybe you can generate out a simple comment "-- Normalized multiple rewards to one" or something

danez:
> like Journey

I'm not sure what you mean with Journey. What is this?

BreakBB:
> > like Journey
> 
> 
> 
> I'm not sure what you mean with Journey. What is this?

If you right click the Questie minimap icon and then click the "My Journey" button in the top right, you are taken to the "Journey" feature of Questie. The most interesting part is the "Quests by Zone" and "Advanced Search" tab, which allow you to get all the details about every quest, NPC, item and object there is in the Questie DB.

Database/Corrections/classicQuestFixes.lua
BreakBB: This is a great idea. Even though I am not sure if we even still need `LoadMissingQuests` 🤔

ExternalScripts(DONOTINCLUDEINRELEASE)/scraper/quest/quest_formatter.py
BreakBB: Perfect 👍🏻

ExternalScripts(DONOTINCLUDEINRELEASE)/scraper/quest/quest_spider.py
BreakBB: Great idea 👍🏻

ExternalScripts(DONOTINCLUDEINRELEASE)/scraper/quest/classic/quest_ids.py
BreakBB: Not sure how well that works, but sounds like a good idea.

danez:
I checked out the Journey feature, but I do not see any reputation or experience displayed there at all. Neither in the log nor when searching for quests

Cabro:
> "Some quests are weird, for example [8857](https://www.wowhead.com/classic/quest=8857/secrets-of-the-colossus-ashi). It says that you receive 4 times +200 reputation from the same faction. I guess this is a mistake on wowheads data as I've never seen getting multiple reputation rewards from one quest for one faction. I haven't changed anything about this yet. So should I normalize this data so that 4x +200 BootyBay becomes 1x +200 BootyBay?"

You can absolutely get multiple 200 reputations. For example when you kill C'thun you get 4x500

BreakBB:
> I checked out the Journey feature, but I do not see any reputation or experience displayed there at all. Neither in the log nor when searching for quests

I wasn't sure either, thanks for checking!

BreakBB:
> > "Some quests are weird, for example [8857](https://www.wowhead.com/classic/quest=8857/secrets-of-the-colossus-ashi). It says that you receive 4 times +200 reputation from the same faction. I guess this is a mistake on wowheads data as I've never seen getting multiple reputation rewards from one quest for one faction. I haven't changed anything about this yet. So should I normalize this data so that 4x +200 BootyBay becomes 1x +200 BootyBay?"
> 
> You can absolutely get multiple 200 reputations. For example when you kill C'thun you get 4x500

Do you think it's worth keeping them separated?

Cabro:
> > > "Some quests are weird, for example [8857](https://www.wowhead.com/classic/quest=8857/secrets-of-the-colossus-ashi). It says that you receive 4 times +200 reputation from the same faction. I guess this is a mistake on wowheads data as I've never seen getting multiple reputation rewards from one quest for one faction. I haven't changed anything about this yet. So should I normalize this data so that 4x +200 BootyBay becomes 1x +200 BootyBay?"
> > 
> > 
> > You can absolutely get multiple 200 reputations. For example when you kill C'thun you get 4x500
> 
> Do you think it's worth keeping them separated?

Just add them together, be done with it

ExternalScripts(DONOTINCLUDEINRELEASE)/scraper/quest/classic/quest_ids.py
danez: Here is the diff between current ids and the ones from the sitemap.
https://gist.github.com/danez/e6434243da86ba2c449357a463b6bf64

This would also include the SOD quests.

Most of the new quests are actually test/obsolete/deprecated quests. For example: https://www.wowhead.com/classic/quest=73/txt-no-reward or https://www.wowhead.com/classic/quest=1
I haven't found a single valid one, but also haven't checked all.

So we then need a ignore list to ignore these test quests. Not sure if it is worth the effort?

BreakBB: I also think it is not worth the effort.

BreakBB (approved):
I am not sure if the "ignore duplicate factions" is actually the right way to do it, since it might also be these quests indeed reward 4 times 200 rep (for example). But from what I saw most of these quests are pretty hard to test.

So I vote to finally merge this and if there is a report in the future we can still fine-tune these things 👍🏻 

Honestly impressive job on this @danez 💪🏻

PR #5417, opened 2024-01-05
[bug] Fix displaying of reputation in Tooltip for quests that do not have a description

danez:
Currently when a quest has no description the reputation reward when holding shift is not displayed.

For example a quest with a description:

Screenshot 2024-01-05 at 12 33 16

Here is a quest without description:
Screenshot 2024-01-05 at 12 42 49

and with the fix:

Screenshot 2024-01-05 at 12 33 29

All Tooltips above are with holding SHIFT

View the diff without whitespace to see the actually changes.


BreakBB (approved):
Great finding! Thanks a lot 👍🏻

PR #5416, opened 2024-01-05
[quest] ADDED: 'Alonso the Dragonslayer' Rune Quest To QuestDB

Sommos:
## Issue references

Fixes #5414

## Proposed changes

- ADDED: 'Alonso the Dragonslayer' is now in the QuestDB, and should be accessible to users.

PR #5422, opened 2024-01-07
[quest] MODIFIED: 'Trial of The Owls' Rune Quest Objectives Text To QuestDB

Sommos:
## Issue references

Fixes N/a

## Proposed changes

- MODIFIED: 'Trial of The Owls' objectivesText in in the QuestDB is now clearer, and should be accessible to users.


Sommos (commented):
This will need to have https://github.com/Questie/Questie/pull/5416 merged first.

Issue #5443, opened 2024-01-09
SoD runes


BreakBB:
## Description

# Druid

- Living Seed ✅ 
- Wild Strikes ⏳ (needs #5362)
- Fury of Stormrage ✅ 
- Survival of the Fittest ✅ 
- Starsurge ✅ 
- Lifebloom ✅ 
- Skull Bash ✅ 
- Savage Roar ✅ 
  -  Westfall location ⌛ (needs #5362)
- Sunfire ✅ 
- Lacerate ✅ 
  - Darkshore location ⌛ (needs #5362)
- Wild Growth ✅ 
- Mangle ✅ 
  - Mulgore location ⌛ (needs #5362)

# Hunter

- Heart of the Lion ✅ 
- Master Marksman ✅ 
- Lone Wolf ✅ 
- Cobra Strikes ✅ 
- Kill Command ✅ 
- Sniper Training ✅ 
- Serpent Spread ✅ 
- Flanking Strikes ✅ 
- Beast Mastery ✅ 
  - Silverpine Forest location ⌛ (needs #5362)
- Chimera Shot ✅ 
- Explosive Shot ✅ 
- Carve ⌛ (needs #5362)

# Mage

- Burnout ✅ 
- Fingers of Frost ✅ 
- Regeneration ✅ 
- Enlightment ⌛ (needs #5362)
- Icy Veins ✅ 
- Arcane Surge ✅ 
- Mass Regeneration ✅ 
- Living Flame ⌛ (needs #5362)
- Rewind Time ✅ 
- Living Bomb ⌛ (needs #5362)
- Arcane Blast ✅ 
- Ice Lance  ✅ 

# Paladin

- Seal of Martyrdom ✅ 
- Divine Storm ✅ 
- Horn of Lordaeron ✅ 
- Aegis ✅ 
- Divine Sacrifice ✅ 
- Inspiration Exemplar ✅ 
- Avenger's Shield ✅ 
- Exorcist ⌛ (needs #5362)
- Rebuke ✅ 
- Beacon of Light ✅ 
- Crusader Strike ✅ 
- Hand of Reckoning ✅ 
  - Westfall location ⌛ (needs #5362)

# Priest

- Void Plague ✅
  - Dun Morogh, Durotar and Teldrassil location ⌛ (needs #5362)
- Serendipity ✅ 
- Strength of Soul ✅ 
- Twisted Faith ✅ 
  - The Barrens location ⌛ (needs #5362)
[...]

PR #5518, opened 2024-01-16
[quest] ADDED: 'Flanking Strike' Hunter Durotar Fake Rune Quest To QuestDB

### LINKED ISSUE #5516: [quest] 'Flanking Strike' Durotar Hunter Fake Rune Quest Missing From QuestDB
## Description

The SoD Hunter fake rune quest 'Flanking Strike' Durotar is not currently in the QuestDB. 

## Wowhead link

https://www.wowhead.com/classic/guide/season-of-discovery/hunter-class-runes-discovery-location
https://www.wowhead.com/classic/spell=425762/engrave-pants-flanking-strike
https://www.wowhead.com/classic/npc=208124/raluk
https://www.wowhead.com/classic/item=207590/durotar-pig-meat

## Screenshots

N/a

## Questie & WoW version

v9.2.15
WoW Classic SoD - v1.15.052409

Sommos:
## Issue references

Fixes #5516
Linked To #5443 

## Proposed changes

- ADDED: 'Flanking Strike' Durotar Hunter Fake Rune Quest is now in the QuestDB, and is currently hidden due to too many icons because of the boars involved in this quest.


Database/Corrections/sodQuestFixes.lua
```diff
@@ -2737,6 +2737,18 @@ function SeasonOfDiscovery:LoadQuests()
             [questKeys.requiredSpell] = -425762,
             [questKeys.zoneOrSort] = sortKeys.HUNTER,
         },
+        [90170] = {
+            [questKeys.name] = "Flanking Strike",
+            [questKeys.startedBy] = {{208124,3099,3100,3225}},
```
BreakBB, line 2742: We have other quests like this as well, with the "kill any of this very unspecific mob across the zone and then go to this one place" (the other "Flanking Strike" quests above as example).

I am wondering if it is the better approach to only show the location where you need to go at the end and remove the unspecific mobs from the `startedBy`.

In this example we would only leave 208124 in, remove the rest and improve the `objectivesText` to state that you need to kill boars in Durotar 🤷🏻

Sommos: This may be a better option, however there are other quests that are just different locations for 'Flanking Strike', that have 6-7 mobs, and so listing all of those out in the objectivesText may be a bit of a pain.

BreakBB: Yeah, maybe those objectives could be better in the wording like "acquire item X which you get from A, B and C for example" and not list ALL locations, only the "best".

Sommos: Okay, a cursory solution is here: https://github.com/Questie/Questie/pull/5518/commits/6557e6b64b1ea11e8dac578879cd27c6eb14daad

As all of the NPCs required have the suffix 'Mottled Boar' I think this is the elegant way of doing it.

BreakBB: Beautiful 🥳

BreakBB: I'll go through the already added runes on `master` and see if I can do something similar for them

Database/Corrections/SeasonOfDiscovery.lua
```diff
@@ -326,6 +327,7 @@ local questsToBlacklistBySoDPhase = {
         [90167] = true, -- Hiding Hunter Flanking Strike Dun Morogh for now as there are too many icons
         [90168] = true, -- Hiding Hunter Flanking Strike Teldrassil for now as there are too many icons
         [90169] = true, -- Hiding Priest Power Word: Barrier Redridge Mountains for now as there are too many icons
+        [90170] = true, -- Hiding Hunter Flanking Strike Durotar for now as there are too many icons
```
BreakBB, line 330:
```suggestion
```

PR #5618, opened 2024-01-30
[feature] Add Tutorial for SoD runes

BreakBB:
## Issue references

Fixes #

## Proposed changes

- Add a tutorial step for SoD for runes

## Screenshots

grafik


Sommos:
LGTM so far. 

Do we want an option to show P2 runes if the user wants to see them?

BreakBB:
> LGTM so far. 
> 
> 
> 
> Do we want an option to show P2 runes if the user wants to see them?

Yes, but that should be added with the general logic for P2 runes

PR #5621, opened 2024-02-03
[feature] Add option to enable coordinates on Mini World Map

viliger2:
## Proposed changes

- Add an option to enable map coordinates on Mini World Map's title, similar to how it works currently on full sized World Map. 

## Known issues

- There is a small issue of "Show Mini World Map Coordinates" not fitting existing option structure, so I had to trim it manually to "Mini World Map Coord.".
- Map's original title shows up for a moment when changing zones and when you open a map after closing it with any other zone being selected that is not your current one. This is most likely because Blizzard UI changing title to match currently selected zone before WriteCoords has a chance to be run due to it having relatively low update interval. I don't think this is a major issue, but I understand if pull request will be rejected because of it.

## Screenshots
image
image
image

### DIFF
--- a/Modules/Options/GeneralTab/QuestieOptionsGeneral.lua
+++ b/Modules/Options/GeneralTab/QuestieOptionsGeneral.lua
@@ -298,7 +298,7 @@ function QuestieOptions.tabs.general:Initialize()
                             QuestieOptions:SetProfileValue(info, value)
 
                             if not value then
-                                QuestieCoords.ResetMapText();
+                                QuestieCoords.ResetMiniWorldMapText();
                             end
                         end,
                     },
--- a/Modules/Options/GeneralTab/QuestieOptionsGeneral.lua
+++ b/Modules/Options/GeneralTab/QuestieOptionsGeneral.lua
@@ -287,6 +287,22 @@ function QuestieOptions.tabs.general:Initialize()
                             end
                         end,
                     },
+                    miniWorldMapCoordinatesEnabled = {
+                        type = "toggle",
+                        order = 5.55,
+                        disabled = function() return not Questie.db.profile.mapCoordinatesEnabled end,
+                        name = function() return l10n('Mini World Map Coord.') end,
+                        desc = function() return l10n("Place the Player's coordinates and Cursor's coordinates on the Mini World Map's title."); end,
+                        get = function(info) return QuestieOptions:GetProfileValue(info); end,
+                        set = function (info, value)
+                            QuestieOptions:SetProfileValue(info, value)
+
+                            if not value then
+                                QuestieCoords.ResetMapText();
+                            end
+                        end,
+                    },			
+					Spacer_Range = QuestieOptionsUtils:Spacer(5.56),
                     mapCoordinatePrecision = {
                         type = "range",
                         order = 5.6,
--- a/Modules/Options/GeneralTab/QuestieOptionsGeneral.lua
+++ b/Modules/Options/GeneralTab/QuestieOptionsGeneral.lua
@@ -301,8 +301,8 @@ function QuestieOptions.tabs.general:Initialize()
                                 QuestieCoords.ResetMapText();
                             end
                         end,
-                    },			
-					Spacer_Range = QuestieOptionsUtils:Spacer(5.56),
+                    },
+                    Spacer_Range = QuestieOptionsUtils:Spacer(5.56),
                     mapCoordinatePrecision = {
                         type = "range",
                         order = 5.6,
--- a/Modules/Options/GeneralTab/QuestieOptionsGeneral.lua
+++ b/Modules/Options/GeneralTab/QuestieOptionsGeneral.lua
@@ -297,7 +297,7 @@ function QuestieOptions.tabs.general:Initialize()
                         set = function (info, value)
                             QuestieOptions:SetProfileValue(info, value)
 
-                            if not value then
+                            if (not value) then
                                 QuestieCoords.ResetMiniWorldMapText();
                             end
                         end,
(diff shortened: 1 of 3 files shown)


BreakBB:
Hey @viliger2 thanks for bringing this up! I'll look into this shortly to fully review, test and hopefully merge it 😬

Modules/QuestieCoordinates.lua
```diff
@@ -118,3 +133,13 @@ end
 function QuestieCoords:ResetMapText()
     GetMapTitleText():SetText(WORLD_MAP);
 end
+
+function QuestieCoords:ResetMiniWorldMapText()
```
BreakBB, line 137:
```suggestion
function QuestieCoords.ResetMiniWorldMapText()
```

Modules/QuestieCoordinates.lua
```diff
@@ -80,6 +89,12 @@ function QuestieCoords:WriteCoords()
         worldmapCoordsText = worldmapCoordsText.."|  Player: "..format(precision.. " X , ".. precision .." Y", posX, posY);
         -- Add text to world map
         mapTitleText:SetText(worldmapCoordsText)
+        if(Questie.db.profile.miniWorldMapCoordinatesEnabled) then
+            local miniWorldMapTitleText = GetMiniWorldMapTitleText()
+            if(miniWorldMapTitleText) then
```
BreakBB, line 94:
```suggestion
            if miniWorldMapTitleText then
```

Modules/QuestieCoordinates.lua
```diff
@@ -118,3 +133,13 @@ end
 function QuestieCoords:ResetMapText()
     GetMapTitleText():SetText(WORLD_MAP);
 end
+
+function QuestieCoords:ResetMiniWorldMapText()
+    local currentMapId = WorldMapFrame:GetMapID();
+    if(currentMapId) then
```
BreakBB, line 139:
```suggestion
    if currentMapId then
```

Modules/QuestieCoordinates.lua
```diff
@@ -118,3 +133,13 @@ end
...
+        local info = C_Map.GetMapInfo(currentMapId);
+        if(info) then
```
BreakBB, line 141:
```suggestion
        if info then
```

Modules/QuestieCoordinates.lua
```diff
@@ -80,6 +89,12 @@ function QuestieCoords:WriteCoords()
         worldmapCoordsText = worldmapCoordsText.."|  Player: "..format(precision.. " X , ".. precision .." Y", posX, posY);
         -- Add text to world map
         mapTitleText:SetText(worldmapCoordsText)
+        if(Questie.db.profile.miniWorldMapCoordinatesEnabled) then
```
BreakBB, line 92:
```suggestion
        if Questie.db.profile.miniWorldMapCoordinatesEnabled then
```

Modules/QuestieCoordinates.lua
```diff
@@ -27,6 +27,15 @@ local function GetMapTitleText()
     end
 end
 
+local function GetMiniWorldMapTitleText()
+    local regions = {WorldMapFrame.MiniBorderFrame:GetRegions()}
+    for i = 1, #regions do
+        if (regions[i].SetText) then
```
BreakBB, line 33:
```suggestion
        if regions[i].SetText then
```

Modules/Options/GeneralTab/QuestieOptionsGeneral.lua
```diff
@@ -287,6 +287,22 @@ function QuestieOptions.tabs.general:Initialize()
...
+                        type = "toggle",
+                        order = 5.55,
+                        disabled = function() return not Questie.db.profile.mapCoordinatesEnabled end,
+                        name = function() return l10n('Mini World Map Coord.') end,
+                        desc = function() return l10n("Place the Player's coordinates and Cursor's coordinates on the Mini World Map's title."); end,
+                        get = function(info) return QuestieOptions:GetProfileValue(info); end,
+                        set = function (info, value)
+                            QuestieOptions:SetProfileValue(info, value)
+
+                            if not value then
```
BreakBB, line 300:
```suggestion
                            if (not value) then
```

Modules/Options/GeneralTab/QuestieOptionsGeneral.lua
```diff
@@ -287,6 +287,22 @@ function QuestieOptions.tabs.general:Initialize()
                             end
                         end,
                     },
+                    miniWorldMapCoordinatesEnabled = {
+                        type = "toggle",
+                        order = 5.55,
+                        disabled = function() return not Questie.db.profile.mapCoordinatesEnabled end,
+                        name = function() return l10n('Mini World Map Coord.') end,
+                        desc = function() return l10n("Place the Player's coordinates and Cursor's coordinates on the Mini World Map's title."); end,
```
BreakBB, line 295: These two new strings need a corresponding translation entry in https://github.com/Questie/Questie/blob/master/Localization/Translations/Options/General.lua

BreakBB (changes requested):
Hey @viliger2 

I added a few really minor points about code style, which you can ignore if you want to.

A big question from my side is: Do we even need a new option? As a user I would honestly not expect to have an option to show the title on the fullscreen map while I can choose to not show it on the small version. I think we are good to have a single option to either show it or not and use it for both, fullscreen and "mini world map".

Maybe you had something differently in mind?

BreakBB:
> > Maybe you had something differently in mind?
> 
> Mostly the fact that some people might get uppity if you change current functionality. You know how it is when you change stuff.

Honestly I think the change was when Blizzard introduced the mini world map and Questie didn't add the title to it 🤔 I personally think it's enough to simply always show the title on the mini version, if the option is enabled.

Also: Thanks for the quick reply! 👍🏻

viliger2:
I am fine with merging options. I'll update the PR with Russian localisation.

BreakBB:
Sounds good @viliger2 👍🏻

viliger2:
I've removed separate option for mini world map. And since that option no longer exists, we don't need additional localisation.

BreakBB:
Thank you very much @viliger2 - Tested and verified 👍🏻

BreakBB (approved):

PR #5633, opened 2024-02-08
Update zhTW translations and change font to show Chinese characters.

class2u:
Update zhTW translations of tutorial

Change the font to STANDARD_TEXT_FONT so that Chinese characters can be displayed.


anon1231823:
@class2u you should join the discord channel for translators. They had the strings there before Phase 2 came out so we knew ahead of time

class2u:
@anon1231823 Thanks for your suggestion, I have joined.

BreakBB:
Thanks a lot for this @class2u 👍🏻

PR #5789, opened 2024-03-16
Sod translations

### LINKED ISSUE #5445: Add translations for SoD quests and runes
## Description

Currently all new SoD quests and manually created fake quests for the runes can not be translated.

We need a proper way to handle this.

BreakBB:
## Issue references

Fixes #5445

## Proposed changes

-
-


anon1231823:
Since SoD is "over", wonder if there is any sense continuing with this.

BreakBB:
I am honestly a bit ashamed I never finished this... I also don't think it's time well spent at this point.

BreakBB:
We won't finish this.

Issue #4266, opened 2022-09-30
Add some way to show quest chain information


Muehe:
## Description

There are many quest that depend on other quest(-s) to be finished first, we need a way to indicate this, especially when group members are on different stages of the chain.

![User Story](https://cdn.discordapp.com/attachments/1016938047285248011/1016938047872442438/unknown.png)

BreakBB:
In addition the "Missing Pre Quest" section of the "Quests by Zone" tab should be improved to not only show a pop up of the missing pre quest, but actually contain the entire quest chain.

Current state:
![Image](https://github.com/user-attachments/assets/42898592-ca49-496e-a735-82629d8dd0e1)

anon1231823:
fixed by https://github.com/Questie/Questie/pull/5803

Logonz:
Should be fixed

PR #5803, opened 2024-03-31
[feature] Add information about remaining quests in chain in expanded map tooltips

jenswaage:
## Issue references

 Related to #4266 

## Proposed changes
- List remaining quests in a chain in the expanded map tooltip (shift key) w/ option to toggle
- Minor refactors and comments

## Screenshots

Remaining quests in chain for a single quest
single_quest

Remaining quests in chains for multiple quests
multiple quests

New option added
Screenshot 2024-04-01 at 00 29 02


Cabro:
Looks very good, however i know for sure the DB doesn't have the correct next in chain quest id for even half the quests :-s

jenswaage:
@Cabro isn’t that a separate issue? Or do you think it reduces the usefulness of the feature to the extent that it must be solved first?

MyTechnoHunter:
Does it honor the show questID option like the shift tooltips do?

MyTechnoHunter:
Nice, will help in reporting i would think

Cabro:
> @MyTechnoHunter added support Screenshot 2024-04-06 at 17 48 33

How are the next quests in the chain available if you are a few steps before those?

MyTechnoHunter:
> 
> How are the next quests in the chain available if you are a few steps before those? I'd put a different word instead of "Available" or just remove that completely.
> 
> Another thing, there are many quest chains where you can only get to the next quests if you gain a level, or several. How is it handling that case?

Beyond scope i would think..
Requirements of the chained quests can be handled by lookup, the fact the current quest _is_ a chain quest is more than helpful

jenswaage:
@Cabro 
> How are the next quests in the chain available if you are a few steps before those? I'd put a different word instead of "Available" or just remove that completely.

I agree with this. Changed it to only show when the quest is a group or dungeon quest.
Screenshot 2024-04-06 at 20 41 06
Screenshot 2024-04-06 at 20 19 05

> Another thing, there are many quest chains where you can only get to the next quests if you gain a level, or several. How is it handling that case?

This feature just naively lists all remaining quests that are part of a direct series regardless of level or other requirements. The intention of the feature is just to have a quick way to see if a quest is part of a chain and how long that chain is. I agree with @MyTechnoHunter that any additional information is better handled via some sort of lookup so as to not overload the tooltip with information.

Modules/Tooltips/MapIconTooltip.lua
```diff
@@ -441,28 +493,35 @@ function _MapIconTooltip:IsMinimapInside()
...
 end
 
-function _MapIconTooltip:GetAvailableOrCompleteTooltip(icon)
-    local tip = {};
-    if icon.data.Type == "complete" then
-        tip.type = "(" .. l10n("Complete") .. ")";
+---@param quest Quest
+---@return string tag
+--- Get the quest tag to display in the tooltip
+function GetQuestTag(quest)
```
BreakBB, line 499:
```suggestion
--- Get the quest tag to display in the tooltip
---@param quest Quest
---@return string tag
local function _GetQuestTag(quest)
```

Modules/Tooltips/MapIconTooltip.lua
```diff
@@ -441,28 +493,35 @@ function _MapIconTooltip:IsMinimapInside()
...
         else
-            tip.type = "(" .. l10n("Available") .. ")";
+            return "(" .. l10n("Available") .. ")";
         end
     end
+end
+
+function _MapIconTooltip:GetAvailableOrCompleteTooltip(icon)
+    local tip = {};
+    tip.type = GetQuestTag(icon.data)
```
BreakBB, line 524:
```suggestion
    tip.type = _GetQuestTag(icon.data)
```

BreakBB (approved):
Thank you very much for this @jenswaage and sorry for not looking into this earlier.

I liked that you extracted `_GetQuestTag` 👍🏻

Modules/Tooltips/MapIconTooltip.lua
```diff
@@ -257,6 +260,53 @@ function MapIconTooltip:Show()
                     end
                 end
 
+                if shift and Questie.db.profile.enableTooltipsNextInChain then
+                    -- add quest chain info
+                    local nextQuest = QuestieDB.GetQuest(QuestieDB.GetQuest(questData.questId).nextQuestInChain)
+                    local firstInChain = true;
+                    while nextQuest ~= nil do
+
+                        local nextQuestTitleString = "";
```
BreakBB, line 269:
```suggestion
                        local nextQuestTitleString;
```

Cabro:
image

This needs some adjusting to not show blacklisted quests.

Issue #5802, opened 2024-03-30
Unavailable quests SOD


GeodeBlast:
As a gnome mage i can see both [[40] The Art of the Armorsmith (5283)], [[40] The Way of the Weaponsmith (5284)] as available quests on the map, but there is no quests to be picked up there. From what I see on Wowhead they are restricted to Humans, Dwarfs, and Night elves.

BreakBB:
Hey @GeodeBlast you are a Blacksmith on that character though, aren't you? Are you playing SoD or Era? Also what level do you have?

GeodeBlast:
Yes this was in sod phase 2, I was max blacksmithing and lvl 40

Den tis 9 apr. 2024 21:52Björn Böing ***@***.***> skrev:

BreakBB:
Thank you very much for your fast reply. I am quite sure Blizzard simply changed the requiredLevel for this quest to be above level 40, because that is how they handled profession quests in earlier phases.

Would be great if you could confirm (if you already played that char in P3 and leveled blacksmithing 😁 )

GeodeBlast:
Hello! I missed an RP moment before being eligible for the quest where I
have to talk to Bengus Deepforge, then Myolor Sunderfury which then made me
eligible for [[40] The Art of the Armorsmith (5283)] as a gnome mage lvl50
with 230 blacksmithing. I wonder if this was still possible during p2.

Den tis 9 apr. 2024 kl 22:01 skrev Björn Böing ***@***.***>:

PR #5810, opened 2024-04-09
[quest] MODIFIED: requiredRaces & Classes for quests

Sommos:
## Issue references

#5802 

## Proposed changes

- ADDED: `requiredRaces` & `requiredClasses` overrides to 'The Way Of the Armorsmith' & The 'Way of the Weaponsmith' to show only on correct races


BreakBB:
I also think there is no class or race requirement for this quest. However, I think the requiredLevel changed to level 41 (or higher) for SoD, because that is what Blizzard did with other profession quests.

PR #5844, opened 2024-04-17
Implement objective icon DB overrides

Yttrium-tYcLief:
This feature allows us to specify an icon in the third field of a monster, object, or item objective in order to override the icon that gets drawn on the map for that objective.

For instance, if you have an item objective, but the item is obtained by *talking* to an NPC, previously our only method of implementing that was to 'fake' it by **making a correction where the NPC *drops* the item**.

```lua
-- sodItemFixes.lua
[219770] = { -- Gemeron's Field Report
    [itemKeys.npcDrops] = {221482},
},
```
```lua
-- sodQuestFixes.lua
[81727] = { -- Recover Incursion Field Report: Ashenvale
    [questKeys.objectives] = {nil,nil,{{219770}}}
},
```

Then, Questie would display a 'loot' icon on the NPC.

**Now, we can tell it to display a 'talk' icon instead**, like so:
```lua
-- sodQuestFixes.lua
[81727] = { -- Recover Incursion Field Report: Ashenvale
    [questKeys.objectives] = {nil,nil,{{219770, nil, Questie.ICON_TYPE_TALK}}}
},
```

In terms of data, the NPC is still registered as 'dropping' the item, but at the very least players will have a better suggestion as to what to do for each objective.

This feature is also useful for situations where the raw objective of a quest is of type `monster`, and it's using a kill bunny NPC ID, even though you're actually supposed to perform some sort of action. The action 'credits' you with a kill for that monster, so from a code standpoint, drawing the 'slay' icon was appropriate, but it wasn't very intuitive for users. Now we can display 'event' or 'object' or 'talk' icons, or whatever we feel is appropriate.


BreakBB (approved):
Looks good to me code wise 👍🏻

PR #5887, opened 2024-04-28
Cataclysm lookups

EKE00372:
## Proposed changes

- add cata lookups: NPCs, Quests and Items
- update classic lookups/Quests/esMX


Localization/lookups/Classic/lookupQuests/esMX.lua
```diff
@@ -1,11 +1,7 @@
-if GetLocale() ~= "esMX" then
-    return
-end
-
----@type l10n
```
BreakBB, line 5: This needs to be added again.

PR #6008, opened 2024-05-25
[quest] Hide "Heeding the Call" in Cataclysm

SacredBagel:
## Proposed changes

- This hides the druid quest "Heeding the Call" in Cata as it has been removed: https://www.wowhead.com/cata/quest=5924/heeding-the-call


BreakBB:
Hey @SacredBagel thanks a lot taking the time to do this PR 💪🏻

PR #6009, opened 2024-05-25
[quest] Fix "Conscript of the Horde" exclusivity

SacredBagel:
## Proposed changes

- This fixes the exclusivity of the "Conscript of the Horde" quest. This quest can be picked up with either "Preserving the Barrens" or "Warchief's Command: Northern Barrens!", but not both. (the latter two are still exclusive to each other)
- Updates the breadcrumbs between these quests for the entire chain
- Removes an old quest that was previously referenced in the chain, but has been removed from Cata

## Screenshots
Here's a screenshot where I have both of these in my quest log showing that this is possible!

WoWScrnShot_052524_165237

### DIFF
--- a/Database/Corrections/QuestieQuestBlacklist.lua
+++ b/Database/Corrections/QuestieQuestBlacklist.lua
@@ -1503,6 +1503,7 @@ function QuestieQuestBlacklist:Load()
         [819] = QuestieCorrections.CATA_HIDE, -- Removed with cata
         [830] = QuestieCorrections.CATA_HIDE, -- Removed with cata
         [832] = QuestieCorrections.CATA_HIDE, -- Removed with cata
+        [842] = QuestieCorrections.CATA_HIDE, -- Removed with cata
         [882] = QuestieCorrections.CATA_HIDE, -- Removed with cata
         [883] = QuestieCorrections.CATA_HIDE, -- Removed with cata
         [884] = QuestieCorrections.CATA_HIDE, -- Removed with cata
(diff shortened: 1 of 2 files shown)


BreakBB:
In the current state of your fix "Preserving the Barrens" can still not be picked up when you already have "Conscript of the Horde" in your quest log. I think you want to remove that exclusiveness as well, correct?

Hit me up on the Questie Discord, if you have any questions!

BreakBB:
Oh and it also looks like none of the quests is correctly marked as breadcrumb quest to https://www.wowhead.com/cata/quest=871/in-defense-of-far-watch

SacredBagel:
> In the current state of your fix "Preserving the Barrens" can still not be picked up when you already have "Conscript of the Horde" in your quest log. I think you want to remove that exclusiveness as well, correct?

That's right - I think this already exists here as a previous fix where Preserving the Barrens is only exclusive to Warchief's Command and not Conscript of the Horde: https://github.com/Questie/Questie/blob/dcf54911d5c8a8d6e38213309d0a0b0548f143c7/Database/Corrections/cataQuestFixes.lua#L2368-L2370

I'll look into updating that breadcrumb data 👍

BreakBB:
You are absolutely right, thanks! No clue, why I did only fix one side of the quest and not both 😅

SacredBagel:
Alright I think I've figured it out. I updated the links between quests in the "Conscript of the Horde" [chain](https://www.wowhead.com/cata/quest=840/conscript-of-the-horde). Also added another removed [quest](https://www.wowhead.com/cata/quest=842/crossroads-conscription) that was previously referenced here.

BreakBB:
Good job and very detailed PR, thanks 👍🏻

Very nice of you to also fix the `nextQuestInChain` fields. These are empty for way too many quests in the cata DB 😩

PR #6023, opened 2024-05-28
[quest] Removal of Pre-Cataclysm Mage Quests

JMZ-UK:
## Proposed changes

- Initial Removal of all Pre-Cataclysm Mage Quests.
  (Pre-reqs and race-reqs still to fix)

- Fix of PR6022


BreakBB:
Thanks for this PR @JMZ-UK 🙏🏻 

If you plan to provide more PRs, I suggest you create a branch on your fork, before adding commits, so they are not on `master`. That way it will be much easier for you to sync your fork.

PR #6097, opened 2024-07-09
Update the issue template to add Cata to the Wowhead links

renodr:
## Issue references

Fixes # N/A

## Proposed changes

- Adds a link to https://wowhead.com/cata into the list of Wowhead links.

## Screenshots

N/A


BreakBB:
Good catch @renodr , thanks!

PR #6193, opened 2024-07-19
[Quest] Cata Classic - Fix The Stolen Keg quest chain

gewoonnico:
Fix tags for the first Wetlands quest chain, starting with "The Stolen Keg" and ending at "Assault on Menethil Keep".
There are two breadcrumb quests for this chain: "Hero's Call: Wetlands!" and "Checking on the Boys".

## Proposed changes
- Properly assign breadcrumb quests "Hero's Call: Wetlands!" and "Checking on the Boys" for "The Stolen Keg"
  - Remove _preQuestSingle_ from "The Stolen Keg"
  - Add _nextInQuestChain_ to breadcrumb quests
- Add _nextInQuestChain_ tags for entire chain, up to "Assault on Menethil Keep"
  - Remove _preQuestSingle_ from "Assault on Menethil Keep" since pre-quest seems to be another breadcrumb (could start that quest without having done any of the previous quests)

### DIFF
--- a/Database/Corrections/cataQuestFixes.lua
+++ b/Database/Corrections/cataQuestFixes.lua
@@ -2903,6 +2903,10 @@ function CataQuestFixes.Load()
             [questKeys.objectives] = {{{39999,nil,Questie.ICON_TYPE_INTERACT}}},
             [questKeys.extraObjectives] = {{nil,Questie.ICON_TYPE_EVENT,l10n("Escort the Injured Fawn back home"),0,{{"monster",39930}}}},
         },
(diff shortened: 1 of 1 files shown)


BreakBB:
Great PR, thanks for adding these nice little details @gewoonnico 🥳

Cabro:
I see you used `[questKeys.requiredMaxLevel] = 24,`, which is awesome, the alliance ones were on my list!
However, following the logic from horde ones, I was expecting the last level for this to be available to be 23. Can you still get that quest at 24?

PR #6194, opened 2024-07-19
[Quest] Cata Classic - Fix The Third Fleet quest chain

gewoonnico:
Fix the Menethil Harbor quest chain starting with "The Third Fleet" and ending at "The Eye of Paleth".

## Proposed changes
- Add _nextQuestInChain_ for entire chain
- Fix _npcDrops_ ID on Intrepid Strongbox Key for "Lifting the Curse"

### DIFF
--- a/Database/Corrections/cataQuestFixes.lua
+++ b/Database/Corrections/cataQuestFixes.lua
@@ -3284,6 +3284,19 @@ function CataQuestFixes.Load()
         [25810] = { -- The Hatchery Must Burn
             [questKeys.startedBy] = {{41003}},
         },
+        [25815] = { -- The Third Fleet
+            [questKeys.nextQuestInChain] = 25816,
+        },
+        [25816] = { -- Cursed To Roam
+            [questKeys.nextQuestInChain] = 25817,
+        },
+        [25817] = { -- The Cursed Crew
+            [questKeys.nextQuestInChain] = 25818,
+        },
+        [25818] = { -- Lifting the Curse
+            [questKeys.nextQuestInChain] = 25819,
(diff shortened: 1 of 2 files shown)


Database/Corrections/cataQuestFixes.lua
```diff
@@ -3284,6 +3284,19 @@ function CataQuestFixes.Load()
...
+        },
+        [25816] = { -- Cursed To Roam
+            [questKeys.nextQuestInChain] = 25817,
+        },
+        [25817] = { -- The Cursed Crew
+            [questKeys.nextQuestInChain] = 25818,
+        },
+        [25818] = { -- Lifting the Curse
+            [questKeys.nextQuestInChain] = 25819,
+            [questKeys.objectives] = {{{41429,nil,Questie.ICON_TYPE_SLAY}}},
```
BreakBB, line 3298: This is not the correct way to fix the issue.

The original objective is: `{nil,nil,{{2629}},nil,},` which matches the ingame objective (loot the item 2629, which is the key). When then checking the entry for that item you notice the NPC currently dropping the item is wrong:

```lua
[2629] = {'Intrepid Strongbox Key',{1160},nil,nil,nil,nil,2048,nil,1,0,0,13,0},
```

1160 is the old version of "Captain Halyndor". On top of that the cata DB has no spawn location for that NPC (because it is no longer in the game).

So the correct fix is to update the item, to drop from the new version of "Captain Halyndor" (ID: 41429, which you already got for the first fix).

BreakBB (changes requested):

BreakBB (approved):
Looking fine now, good job! 👍🏻

PR #6208, opened 2024-07-19
Feature: Re-click Questie icon to close dropdown menu

### LINKED ISSUE #6180: Questie Icon Dropdown Menu Stays Open When It Should Close
## Bug description

When opening the full sized map in game and then right clicking on the Questie icon in the top right corner of it, the same dropdown menu pops up as when you left click on the icon when the map isnt open. In both cases, re-clicking on the Questie Icon should close that dop down menu. Additionally, if the full map is open, closing the map should also close this dropdown menu. Currently the only way to close this menu is to left click the cancel option at the bottom of the dropdown menu. 

## Screenshots

image

image

## Questie & WoW version

Questie Version: 10.3.5
WoW Version: 4.4.0.55639
Cataclysm

BrandonJack9:
re-clicking the minimap icon with left click closes the dropdown menu, and re-clicking the icon with right click in the worldmap will close the dropdown menu as well

## Issue references

Fixes #6180

## Proposed changes

-Left clicking the Questie minimap icon will close the dropdown menu if its open
-Right clicking the Questie icon in the world map window will close the dropdown menu if its open
-This feature will work in classic and SOD, I will need to update the library to get it working in CATA 

## Screenshots

https://github.com/user-attachments/assets/033d18fa-1122-49b1-b7c5-e7b1672b8c6a

https://github.com/user-attachments/assets/187d3130-31e1-43a5-a228-cdb56b5ec72f

### DIFF
--- a/Libs/LibUIDropDownMenu/LibUIDropDownMenu.lua
+++ b/Libs/LibUIDropDownMenu/LibUIDropDownMenu.lua
@@ -118,6 +118,10 @@ function lib:UIDropDownMenu_InitializeHelper(frame)
     frame:SetHeight(L_UIDROPDOWNMENUQUESTIE_BUTTON_HEIGHT * 2);
 end
 
+function lib:getOpen() 
+    return L_UIDROPDOWNMENUQUESTIE_OPEN_MENU
+end
+
 function lib:UIDropDownMenuButton_ShouldShowIconTooltip(self)
     if self.Icon and (self.iconTooltipTitle or self.iconTooltipText) and (self.icon or self.mouseOverIcon) then
         return GetMouseFocus() == self.Icon;
--- a/Libs/LibUIDropDownMenu/LibUIDropDownMenu.lua
+++ b/Libs/LibUIDropDownMenu/LibUIDropDownMenu.lua
@@ -118,8 +118,9 @@ function lib:UIDropDownMenu_InitializeHelper(frame)
     frame:SetHeight(L_UIDROPDOWNMENUQUESTIE_BUTTON_HEIGHT * 2);
 end
 
+--Questie addition, to know when menus are showing
 function lib:getOpen() 
-    return L_UIDROPDOWNMENUQUESTIE_OPEN_MENU
+    return L_UIDROPDOWNMENUQUESTIE_OPEN_MENU ~= nil
 end
 
 function lib:UIDropDownMenuButton_ShouldShowIconTooltip(self)
--- a/Modules/MinimapIcon.lua
+++ b/Modules/MinimapIcon.lua
@@ -36,8 +36,11 @@ function _MinimapIcon:CreateDataBrokerObject()
             if (not Questie.started) then
                 return
             end
-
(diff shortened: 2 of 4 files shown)


Modules/MinimapIcon.lua
```diff
@@ -57,6 +57,8 @@ function _MinimapIcon:CreateDataBrokerObject()
                 end
 
                 return;
+            elseif button == "LeftButton" and QuestieMenu:isOpen() then
+                QuestieMenu:Hide()
```
BreakBB, line 61: I think I'd prefer if this will be added to the previous case.

So that it is somewhat like:

```lua
if button == "LeftButton" then
    if QuestieMenu.IsOpen() then
        QuestieMenu:Hide()
        return
    end
    ...
elseif button == "RightButton" then
    ...
end
```

Libs/LibUIDropDownMenu/LibUIDropDownMenu.lua
```diff
@@ -118,6 +118,10 @@ function lib:UIDropDownMenu_InitializeHelper(frame)
     frame:SetHeight(L_UIDROPDOWNMENUQUESTIE_BUTTON_HEIGHT * 2);
 end
 
+function lib:getOpen() 
+    return L_UIDROPDOWNMENUQUESTIE_OPEN_MENU
+end
```
BreakBB, line 123: This is full on nitpicking: The value of `L_UIDROPDOWNMENUQUESTIE_OPEN_MENU` is pretty big table and we actually don't need it. We just need to know if it is `nil` or not. Also a small comment for the future is great, as we generally try to avoid changing dependencies and we need to make sure we are not losing that function. 

```suggestion
-- Questie addition, to know when menues are showing
function lib:getOpen() 
    return L_UIDROPDOWNMENUQUESTIE_OPEN_MENU ~= nil
end
```

BreakBB (changes requested):
Hey @BrandonJack9 

Thanks a lot for this PR! I added some comments and suggestions. It seems a lot, but it is absolutely not, great job on your first PR 💪🏻 

I'd try to explain what the requested changes are for, but feel free to ask if something is not clear 😃

Modules/WorldMapButton/WorldMapButton.lua
```diff
@@ -39,8 +39,10 @@ QuestieWorldMapButtonMixin = {
         if button == "LeftButton" then
             Questie.db.profile.enabled = (not Questie.db.profile.enabled)
             QuestieQuest:ToggleNotes(Questie.db.profile.enabled)
-        elseif button == "RightButton" then
+        elseif button == "RightButton" and not QuestieMenu.IsOpen() then
             QuestieMenu:Show()
+        elseif  button == "RightButton" and QuestieMenu.IsOpen() then 
+            QuestieMenu:Hide()
```
BreakBB, line 45:
```suggestion
        elseif button == "RightButton" then
            if QuestieMenu.IsOpen() then
                QuestieMenu:Hide()
            else
                QuestieMenu:Show()
            end
```

BreakBB (approved):

PR #6364, opened 2024-08-07
[Quest] Cata Classic - Add clear objective for "Razorbeak Friends"

gewoonnico:
Add Trained Razorbeak NPC location on map through an extra objective for "Razorbeak Friends", so the player actually knows where to find them on this quest.

## Proposed changes

- Add `extraObjectives` of Trained Razorbeak NPC to "Razorbeak Friends" quest
- Add localization entry for CataObjectives, fill in French as well based on localized NPC name

## Screenshots

After changes (before there was _nothing_ indicated on the map):
image


Database/Corrections/cataQuestFixes.lua
```diff
@@ -4531,6 +4531,9 @@ function CataQuestFixes.Load()
         [26540] = { -- Dangerous Compassion
             [questKeys.extraObjectives] = {{nil, Questie.ICON_TYPE_TALK, l10n("Go back to the ship"),0,{{"monster",43188}}}},
         },
+        [26546] = { -- Razorbeak Friends
+            [questKeys.extraObjectives] = {{nil, Questie.ICON_TYPE_INTERACT, l10n("Feed Trained Razorbeak"), 0, {{"monster",2657}}}},
```
BreakBB, line 4535: Hey @gewoonnico actually it's better to just change the NPC that is listed in the `objectives` to the correct one. That way players will also see the progress on the tooltip.

You should also override the icon used for it:

```suggestion
            [questKeys.objectives] = {{{2657,nil,Questie.ICON_TYPE_INTERACT}}},
```

The new translation can also be removed then, because with `nil` Questie uses the string from the game.

BreakBB (changes requested):

PR #6371, opened 2024-08-08
"The Wild Gods" ZoneID Quest fix [82043]

BrandonJack9:
## Proposed changes

-Just fixes the issue with "The Wild Gods" quest, where the quest was showing "Unkown Zone" for it. 
NOTE: I was unable to check if the fixed worked myself due to not having a character that is a high enough level to accept the quest

## Screenshots
image


BrandonJack9:
didn't properly update questie before creating this pull request. Sorry!

PR #6373, opened 2024-08-08
"The Wild Gods" Quest Correction in sodQuestFixes [82043]

BrandonJack9:
## Proposed changes

-Fixes "The Wild Gods" zoneOrSort field and updates it to Felwood instead of Uknown Zone

## Screenshots
image


BreakBB:
Hey @BrandonJack9 I just noticed there is a follow up quest with the same problem. Can you fix that one as well? Feel free to just add it to this PR (commit and push another change on the existing branch)

https://www.wowhead.com/classic/quest=82044/the-wild-gods

BrandonJack9:
I can totally do that!

PR #6399, opened 2024-08-17
[fix] Custom localization initialization fix

Jakanis:
## Issue

When locales initialized on ADDON_LOADED event, QUESTIE_LOCALES_OVERRIDE variable is not set by localization addon, as it is not loaded yet.

## Proposed changes

- Move locale initialization to PLAYER_LOGIN stages. 

It may increase load time only if QUESTIE_LOCALES_OVERRIDE is set and l10n:InitializeLocaleOverride() is called, so seems like no impact for most users with that.


BreakBB:
Hey @Jakanis thanks a lot for addressing this and sorry for breaking this in the first place!

PR #6414, opened 2024-08-29
Fix for non-string calls to GetTooltip

Tanoh:
## Proposed changes

Make sure `GetTooltip` argument is a string before trying to call `sub` on it. Looks like it is due to interaction with Plater that is causing it. I always get this error whenever a hostile nameplate is shown. This fix fixes it by making sure that the argument is a string.

You could argue that Plater should be fixed as well, which might be true but Questie shouldn't break either if sent arguments it can't handle.

## Screenshots
image


BreakBB:
Hey @Tanoh thanks for taking the time to look into this. You are the first one to report this, so it might be some sort of weird/rare conflict 🤷🏻 There is no harm to add this check though 👍🏻

PR #6488, opened 2024-10-09
[quest] Mark "The Thunderdrome!" as breadcrumb quest

TorpedoCXC:
Marked as breadcrumb quest for 25067 Thunderdrome: The Ginormus!

### DIFF
--- a/Database/Corrections/cataQuestFixes.lua
+++ b/Database/Corrections/cataQuestFixes.lua
@@ -5804,11 +5804,9 @@ function CataQuestFixes.Load()
             [questKeys.exclusiveTo] = {26893},
         },
         [26895] = { -- The Thunderdrome!
-            [questKeys.exclusiveTo] = {25067},
             [questKeys.nextQuestInChain] = 25067,
         },
         [26896] = { -- The Thunderdrome!
-            [questKeys.exclusiveTo] = {25067},
             [questKeys.nextQuestInChain] = 25067,
         },
         [26897] = { -- Blackfathom Deeps
--- a/Database/Corrections/cataQuestFixes.lua
+++ b/Database/Corrections/cataQuestFixes.lua
@@ -5805,9 +5805,11 @@ function CataQuestFixes.Load()
         },
         [26895] = { -- The Thunderdrome!
             [questKeys.exclusiveTo] = {25067},
+            [questKeys.nextQuestInChain] = 25067,
         },
         [26896] = { -- The Thunderdrome!
             [questKeys.exclusiveTo] = {25067},
+            [questKeys.nextQuestInChain] = 25067,
         },
         [26897] = { -- Blackfathom Deeps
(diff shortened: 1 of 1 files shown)


BreakBB:
Hey @TorpedoCXC three things:

1. As these quests already have the `exclusiveTo` set, they shouldn't show up in the first place. How did you encounter an issue?
2. It is correct to use `nextQuestInChain` which then replaces the current `exclusiveTo` correction.
3. Could you try to use the quest names in the PR title instead of the IDs? Maybe even follow the format used in [previous commits](https://github.com/Questie/Questie/commits/master/)? So for this I'd prefer "[quest] Mark "The Thunderdrome!" as breadcrumb quest".
 
Side note for 3.: The PR title will become the commit message and we use the commit messages to generate the changelog for each release. If not, that is no problem, as we always have the chance to update the commit message before merging.

TorpedoCXC:
1.) The only issue was that the alliance and horde quests "The Thunderdrome!" were not marked as breadcrumbs for "Thunderdrome: The Ginormus!". I did not encounter the issue that the breadcrumb quest still showed as available when the main quest was already done. Or to be precise: I did not test this at all.
3.) I will try to remember this. 👍 

My question on Discord was just so I learn how to mark breadcrumb quests in the future and to learn about the attributes `exclusiveTo` and `nextQuestInChain` and their differences in general.

BreakBB:
> The only issue was that the alliance and horde quests "The Thunderdrome!" were not marked as breadcrumbs for "Thunderdrome: The Ginormus!"

Well this is strange, because both correction entries already have the `exclusiveTo` set which work very similar to `nextQuestInChain` 🤔 Once you accept "Thunderdrome: The Ginormus!" then "The Thunderdrome!" should no longer be available 🙈 It's better to use `nextQuestInChain` though 🤷🏻 👍🏻 

> I will try to remember this. 👍

Thanks! Don't stress yourself too much with it though <3

Database/Corrections/cataQuestFixes.lua
```diff
@@ -5805,9 +5805,11 @@ function CataQuestFixes.Load()
         },
         [26895] = { -- The Thunderdrome!
             [questKeys.exclusiveTo] = {25067},
+            [questKeys.nextQuestInChain] = 25067,
```
BreakBB, line 5808:
```suggestion
            [questKeys.nextQuestInChain] = 25067,
```

Database/Corrections/cataQuestFixes.lua
```diff
@@ -5805,9 +5805,11 @@ function CataQuestFixes.Load()
...
         },
         [26896] = { -- The Thunderdrome!
             [questKeys.exclusiveTo] = {25067},
+            [questKeys.nextQuestInChain] = 25067,
```
BreakBB, line 5812:
```suggestion
            [questKeys.nextQuestInChain] = 25067,
```

Issue #6490, opened 2024-10-09
Quest 25533 Pirate Accuracy Increasing shown as available when it is not


TorpedoCXC:
### Description

I tested the prerequisites for this quest and can confirm the accuracy of this comment: https://www.wowhead.com/cata/quest=25533/pirate-accuracy-increasing#comments:id=1187234

Missing prequest, but quest is shown as available alrady:
![WoWScrnShot_100624_203430](https://github.com/user-attachments/assets/2e78ced0-9915-4fdd-b9ac-32c9e94b9fb2)

Last prequest done:
![WoWScrnShot_100624_203827](https://github.com/user-attachments/assets/5bb15df4-1414-4a18-b31c-9eac857bdb1c)

### Questie version

v10.6.0

### Game flavor

Cata

TorpedoCXC:
There is also an alliance version of this quest with also probably 3 prerequisite quests:

See here: https://www.wowhead.com/cata/quest=25532/pirate-accuracy-increasing#comments:id=1180461:reply=194533

PR #6491, opened 2024-10-09
[quest] Add 25533 Pirate Accuracy Increasing

TorpedoCXC:
Add 3 prerequisite quests that all need to be completed before.

Alliance version still needs to be done.
See here: https://github.com/Questie/Questie/issues/6490

### DIFF
--- a/Database/Corrections/cataQuestFixes.lua
+++ b/Database/Corrections/cataQuestFixes.lua
@@ -3824,6 +3824,9 @@ function CataQuestFixes.Load()
             [questKeys.preQuestSingle] = {25505},
             [questKeys.requiredSourceItems] = {54821},
         },
+        [25533] = { -- Pirate Accuracy Increasing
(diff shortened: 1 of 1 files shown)


TorpedoCXC:
Do I have to add `[questKeys.preQuestSingle] = {},` to the quest first?

Cabro:
Only if there is an entry for prequestsingle in DB, you need to negate that one

TorpedoCXC:
> Only if there is an entry for prequestsingle in DB, you need to negate that one

There is indeed:
`[25533] = {"Pirate Accuracy Increasing",{{40487},nil},nil,38,40,946,nil,{"Get a River Boat from Griznak and then use the Frost Cannon to extinguish 15 fires."},nil,{{{40727}},{{203009}},nil,nil,},nil,nil,{25516,},nil,nil,nil,400,nil,nil,nil,nil,nil,262152,nil,nil,{{1133,5}},nil,nil,nil,nil},`

PR #6609, opened 2024-10-19
[quest] Loramus Thalipedes Awaits - add prerequisite quests

### LINKED ISSUE #6595: Quest 25700 Loramus Thalipedes Awaits showing as available when it is not
### Description

https://www.wowhead.com/cata/quest=25700/loramus-thalipedes-awaits

WoWScrnShot_101824_155512

### Questie version

v10.6.1

### Game flavor

Cata

TorpedoCXC:
## Issue references

Fixes #6595


BreakBB:
I think it's good to add the missing `objectives` in the same to fully fix the referenced issue.

TorpedoCXC:
I would fix it myself, but I am unsure how to do this.

One needs to talk with the NPC 7506 (Bloodmage Lynnore) first. Afterwards, the quest can be completed at NPC 7783 (Loramus Thalipedes).

This is how the quest is currently stored in the database:

```
[25700] = {"Loramus Thalipedes Awaits",{{7506},nil},{{7783}},54,56,946,nil,{"Speak with Bloodmage Lynnore and participate in her ritual."},nil,nil,nil,nil,nil,nil,nil,nil,4,nil,nil,nil,nil,nil,10,nil,nil,nil,nil,nil,nil,nil},
```

Is there any kind of documentation / example for this case? @BreakBB

BreakBB:
I'll try to explain what's the underlying issue here and will also provide an example:

So the objective in the game is "Participate in ritual". From the naming that seems to be an objective of type "event", which would then result us to fix the quest by adding the `triggerEnd` field. This field is mostly used for situations where you have to reach a certain location and it is not directly related to any NPC/object (like [escort quests](https://github.com/Questie/Questie/blob/master/Database/Corrections/cataQuestFixes.lua#L96)).

**BUT** since you said that you simply need to talk to the NPC to fulfill "Participate in ritual", then you can also fix this by adding an objective fix for type "monster" and override the icon. We did this for [Emergecy Aid](https://github.com/Questie/Questie/blob/master/Database/Corrections/cataQuestFixes.lua#L5564) for example.

TorpedoCXC:
> **BUT** since you said that you simply need to talk to the NPC to fulfill "Participate in ritual", then you can also fix this by adding an objective fix for type "monster" and override the icon. We did this for [Emergecy Aid](https://github.com/Questie/Questie/blob/master/Database/Corrections/cataQuestFixes.lua#L5564) for example.

To clarify: You need to start talking by right clicking on the NPC 7506 (Bloodmage Lynnore) and then click on a special chat option with the NPC (the one that are marked with a speech bubble (pretty sure you know what I mean)). The NPC starts doing some stuff then (the ritual) and afterwards one can complete the quest at NPC 7783 (Loramus Thalipedes). Not sure if this makes a difference in your evaluation of this.

TorpedoCXC:
The example looks good by the way. Thx for that.

BreakBB:
> To clarify: You need to start talking by right clicking on the NPC 7506 (Bloodmage Lynnore) and then click on a special chat option with the NPC (the one that are marked with a speech bubble (pretty sure you know what I mean)). The NPC starts doing some stuff then (the ritual) and afterwards one can complete the quest at NPC 7783 (Loramus Thalipedes). Not sure if this makes a difference in your evaluation of this.

I personally would still use `objectives` instead of `triggerEnd` to make it more clear to the player what's to do.

TorpedoCXC:
Please check if I understood the example correctly @BreakBB 

One more thing. The description in the file `cataQuestDB.lua` (and others) is helpful.

```
 ['objectives'] = 10, -- table
        --['creatureObjective'] = 1, -- table {{creature(int), text(string)},...}, If text is nil the default "<Name> slain x/y" is used
        --['objectObjective'] = 2, -- table {{object(int), text(string)},...}
        --['itemObjective'] = 3, -- table {{item(int), text(string)},...}
        --['reputationObjective'] = 4, -- table: {faction(int), value(int)}
        --['killCreditObjective'] = 5, -- table: {{{creature(int), ...}, baseCreatureID, baseCreatureText}, ...}
```

**But**, there are lots of ellipses in the documentation. I assume that this means other options follow. May I suggest to compile a complete list of possible options? Like for example in this PR I seem to have set the third option of `creatureObjective` to `Questie.ICON_TYPE_TALK`. Without your advice, I would've never known that there is such an option.

If such a documentation already exists, I would appreciate a link. Thx.

BreakBB:
> But, there are lots of ellipses in the documentation.

Yeah the documentation is far from perfect and missing quite some stuff we added at some point 🙈 I'll see to add these override icons right away.

PR #6613, opened 2024-10-20
[db] Amulet of Allistarj / Grol - fix drop issues

TorpedoCXC:
Amulet of Allistarj drops from https://www.wowhead.com/cata/object=203229/allistarjian-vault now.
Amulet of Grol drops from https://www.wowhead.com/cata/npc=41267/spirit-of-grol instead of https://www.wowhead.com/cata/npc=7665/grol-the-destroyer

### DIFF
--- a/Database/Corrections/QuestieNPCBlacklist.lua
+++ b/Database/Corrections/QuestieNPCBlacklist.lua
@@ -6,6 +6,8 @@ local QuestieCorrections = QuestieLoader:ImportModule("QuestieCorrections")
 ---@return table<NpcId, boolean>
 function QuestieNPCBlacklist:Load()
     return {
+        [3032] =  QuestieCorrections.CATA_HIDE, -- Beram Skychaser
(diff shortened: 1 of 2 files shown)


Database/Corrections/QuestieNPCBlacklist.lua
```diff
@@ -6,6 +6,8 @@ local QuestieCorrections = QuestieLoader:ImportModule("QuestieCorrections")
 ---@return table<NpcId, boolean>
 function QuestieNPCBlacklist:Load()
     return {
+        [3032] =  QuestieCorrections.CATA_HIDE, -- Beram Skychaser
+        [7666] =  QuestieCorrections.CATA_HIDE, -- Archmage Allistarj
```
BreakBB, line 10: Nice addition 👍🏻

TorpedoCXC: Thx. :)

PR #6722, opened 2024-12-03
[feature] Add Questie Rescale Townsfolk Icon option to Icon Options

### LINKED ISSUE #6708: Icon Scale/size option for vendors, traders, trainers, inn-keeps, flight masters etc.
### Description

Thank you for the awesome addon!

I love having the world Map icons for the various vendors, trader, trainer and service NPCs in Anniversary WoW.
However their Icon sizes on the world Map are currently way too large and there seems to be no direct option to adjust specifically their size/scale.

I was able to get a somewhat okay appearance by tremendously reducing the size of ALL icons and then maxing out the size of Quest related icons - but it would be a lot simpler if there was a Scale option for just the Icons related to vendors/trainers/innkeeps/flight masters and other such NPCs.

Thanks for considering this suggestion.
image

Sommos:
## Issue references

Fixes #6708 

## Proposed changes

- Add a Townsfolk Scaling option to the Icon Scales in the Icons tab of the Questie UI that allows users to only scale Townsfolk for enhanced clarity.

## Screenshots

image


Modules/Options/IconsTab/QuestieOptionsIcons.lua
```diff
@@ -611,10 +611,25 @@ function QuestieOptions.tabs.icons:Initialize()
...
+                        name = function() return l10n('Map Townsfolk Icons'); end,
+                        desc = function() return l10n('How large the townsfolk map icons are.\n(Default: %s)', optionsDefaults.profile.globalTownsfolkScale); end,
+                        width = 1.55,
+                        min = 0.01,
+                        max = 4,
+                        step = 0.01,
+                        get = function(info) return QuestieOptions:GetProfileValue(info); end,
+                        set = function (info, value)
+                            QuestieMap:RescaleTownsfolkIcons()
+                            QuestieOptions:SetProfileValue(info, value)
```
BreakBB, line 626: I think this needs to happen in the opposite order.

Modules/Options/QuestieOptionsDefaults.lua
```diff
@@ -15,6 +15,7 @@ function QuestieOptionsDefaults:Load()
             monsterScale = 1,
             objectScale = 1,
             globalScale = 0.6,
+            globalTownsfolkScale = 0.6,
```
BreakBB, line 18: You introduced this, but does not use it at any place.

Sommos: This is used in the `globalTownsfolkScale` in `QuestieOptionsIcons.lua` to set the default value to 0.6 on the slider.

BreakBB: True. But it is not used to change the scaling of any icon.

Sommos:
This has been moved to https://github.com/Questie/Questie/pull/7209.

PR #6740, opened 2024-12-07
Update Kalimdor.lua

anon1231823:
## Issue references

Fixes #

## Proposed changes

-
-


Localization/Translations/Zones/Kalimdor.lua
```diff
@@ -225,20 +225,20 @@ local kalimdorLocales = {
         ["koKR"] = "썬더 블러프",
         ["esMX"] = "Cima del Trueno",
         ["enUS"] = true,
-        ["frFR"] = "Thunder Bluff",
+        ["frFR"] = "Les Pitons-du-Tonnerre",
```
BreakBB, line 228: Hmmm... This is a hard one. In Vanilla this is "Thunder Bluff" and got a localized name only later (WotLK?).

BreakBB: Scratch that. We are using the post vanilla translations all over the place already 😂 🤷🏻

PR #6759, opened 2024-12-15
Add option to change delay before quest sounds

jsb:
Currently, most quest sounds (objective progress, objective completion) are played with a delay of 0.5 seconds after the actual event. This delay does not feel great for short progress sounds which play noticeably after the on-screen message appears.

This PR adds an option slider to configure a custom delay time for quest sounds. Default is 0.5, as before.

image

### DIFF
--- a/Localization/Translations/Options/Sounds.lua
+++ b/Localization/Translations/Options/Sounds.lua
@@ -182,10 +182,10 @@ local soundsOptionsLocales = {
         ["esES"] = false,
         ["frFR"] = false,
     },
-    ["Delay (in seconds) between the event and playing the sound."] = {
+    ["Delay (in seconds, default: %s) for playing objective progress and completion sounds. Increase this if you hear double sounds."] = {
         ["ptBR"] = false,
         ["ruRU"] = false,
-        ["deDE"] = "Verzögerung (in Sekunden) zwischen dem Ereignis und dem Abspielen des Sounds.",
+        ["deDE"] = "Verzögerung (in Sekunden, Voreinstellung: %s) beim Abspielen von Quest-Ziel-Fortschritts-Sounds. Diesen Wert erhöhen, falls Sounds doppelt abgespielt werden.",
         ["koKR"] = false,
         ["esMX"] = false,
         ["enUS"] = true,
--- a/Localization/Translations/Options/Sounds.lua
+++ b/Localization/Translations/Options/Sounds.lua
@@ -170,6 +170,30 @@ local soundsOptionsLocales = {
         ["esES"] = "El sonido que escuchas cuando avanzas en un objetivo de misión",
         ["frFR"] = "Le son que vous entendez lorsque vous progressez sur un objectif de quête",
     },
+    ["Sound Delay"] = {
+        ["ptBR"] = false,
+        ["ruRU"] = false,
+        ["deDE"] = "Sound-Verzögerung",
+        ["koKR"] = false,
+        ["esMX"] = false,
+        ["enUS"] = true,
+        ["zhCN"] = false,
+        ["zhTW"] = false,
+        ["esES"] = false,
+        ["frFR"] = false,
+    },
+    ["Delay (in seconds) between the event and playing the sound."] = {
+        ["ptBR"] = false,
+        ["ruRU"] = false,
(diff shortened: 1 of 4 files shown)


BreakBB:
Hey @jsb 

thank you very much for taking the time to look into this!

One important thing is: The delay on "objective progress sound" is there, to make sure users don't hear "quest progress" and "quest complete" at the same time for the same progress event. I like the idea to add a setting to change this delay, so users with _good_ hardware can improve their experience.

Modules/Sounds.lua
```diff
@@ -44,9 +44,11 @@ function Sounds.PlayQuestComplete()
         return
     end
 
-    shouldPlayObjectiveSound = false
     shouldPlayObjectiveProgress = false
-    PlaySoundFile(Sounds.GetSelectedSoundFile(Questie.db.profile.questCompleteSoundChoiceName), "Master")
+    shouldPlayObjectiveComplete = false
+    C_Timer.After(Questie.db.profile.soundDelay, function ()
+        PlaySoundFile(Sounds.GetSelectedSoundFile(Questie.db.profile.questCompleteSoundChoiceName), "Master")
+    end)
```
BreakBB, line 51: We don't want a delay for this. As soon as Questie knows the quest is complete, play the sound.

Localization/Translations/Options/Sounds.lua
```diff
@@ -170,6 +170,30 @@ local soundsOptionsLocales = {
...
+        ["deDE"] = "Sound-Verzögerung",
+        ["koKR"] = false,
+        ["esMX"] = false,
+        ["enUS"] = true,
+        ["zhCN"] = false,
+        ["zhTW"] = false,
+        ["esES"] = false,
+        ["frFR"] = false,
+    },
+    ["Delay (in seconds) between the event and playing the sound."] = {
```
BreakBB, line 185: I think it would be good to explain this setting with some more details. It is only for progress events and should only be changed on "good" hardware. Once users hear double sounds, they need to increase it.

Localization/Translations/Options/Sounds.lua
```diff
@@ -170,6 +170,30 @@ local soundsOptionsLocales = {
         ["esES"] = "El sonido que escuchas cuando avanzas en un objetivo de misión",
         ["frFR"] = "Le son que vous entendez lorsque vous progressez sur un objectif de quête",
     },
+    ["Sound Delay"] = {
```
BreakBB, line 173: I suggest to name it "Progress Sound Delay"

jsb:
Thanks @BreakBB, could you clarify how the "double sounds" issue can happen on slow hardware?

Is the issue that the "quest progress" and "quest complete" events could arrive at different times? In my testing, that never seemed to be the case: Even when I turned the delay timer down all the way to 0, I would never hear the "objective progress", "objective complete", nor "quest complete" sounds at the same time.

BreakBB:
First of all: Thanks for coming back to this @jsb 

To your question:

These are the places where "objective complete" and "objective progress" is called:

https://github.com/Questie/Questie/blob/master/Modules/Quest/QuestLogCache.lua#L110-L116

and this it the place where the "quest complete" sound is called:

https://github.com/Questie/Questie/blob/master/Modules/Quest/QuestLogCache.lua#L211-L213

From the perspective of the code flow Questie checks every objective of a quest for changes and plays the "objective complete" OR "objective progress" (both with a 0.5sec delay atm). Then after Questie checked every objective it checks if all objectives are complete and plays the "quest complete" sound if so (without a delay) and cancels the potential objective sound timer.

The way this works, means that _technically_ (without any timer) the "objective complete" sound would always play right away and shortly afterwards the "quest complete" sound, if the objective was the last one to complete.

So the timers make sure that there is enough delay, so that Questie can check if the quest is complete, before it plays any sound at all.

In reality even adding a timer object with 0 delay can be enough delay to make sure to reach the point to check the quest completion state and hinder the other sound from playing. And the API is also limiting the output of multiple sounds at the same time iirc.

All in all: Keep the delay timer for "objective progress" and "objective complete" sound, but don't add one to "quest complete" 😁 And I am happy to merge your config as it totally makes sense to have such an option for dedicated users, going through all the settings Questie offers 👍🏻

BreakBB (approved):
Looks good to me @jsb 

Thanks a lot for looking into this and provide this PR 🎉

PR #6761, opened 2024-12-15
[locale] Fix translations to follow proper German grammar

Metternich11:
Issue references:
No specific issue. This Pull Request fixes errors in the German translations, particularly in the file Townsfolk.lua.

Proposed changes:
Corrected German translations in Townsfolk.lua:
Replaced incorrect terms like "Händler:in" with grammatically correct forms, such as "Händler".
Ensured consistent and proper use of German grammar in the localization file.


BreakBB:
Hey @Metternich11 thanks for looking into this.

The translations currently in place are meant as gender neutral wording. Questie can not determine the gender of an NPC and therefore uses the [Gender-Doppelpunkt](https://de.wikipedia.org/wiki/Gender-Doppelpunkt).

Metternich11:
Hi @BreakBB,

The masculine plural form is traditionally used as a generic form to include all genders in German. This approach aligns with how languages with grammatical gender, such as French, handle similar situations.

The "Gender-Doppelpunkt" is not recognized as part of official German grammar and should therefore be regarded as non-standard.

BreakBB:
> The masculine plural form is traditionally used as a generic form to include all genders in German.

Personally "traditionally" is what is triggering me here.

But you are absolutely right, that there is no (official) standard for this yet in the German language, so we should change it.

Thank you very much 👍🏻

PR #6789, opened 2024-12-19
Update esMX.lua

anon1231823:
## Issue references

Fixes #

## Proposed changes

-
-


anon1231823:
I don't know if this is right, I am using Notepad (not ++) and it says Unix (LF) but why is it showing 4372 additions/deletions?

PR #6841, opened 2025-01-18
Locale update

anon1231823:
## Issue references

Fixes #

## Proposed changes

-
-


Localization/Translations/ExtraObjectives/ClassicObjectives.lua
```diff
@@ -623,7 +623,19 @@ local classicObjectiveLocales = {
         ["enUS"] = true,
         ["frFR"] = "Utilisez le Draco-Incarcinatrix 900 et tire sur un Rejeton draconien dans les Steppes ardentes.",
         ["esES"] = "Usa el Draco-Incarcinatrix 900 y dispara a una cría de dragonante en Las Estepas Ardientes.",
-        ["zhTW"] = false,
+        ["zhTW"] = "使用龍靈採集器900型並擊敗龍人。",
+        ["zhCN"] = false,
+    },
+	 ["Close the portal"] = {
```
BreakBB, line 629:
```suggestion
    ["Close the portal"] = {
```

Cabro: I moved that string to cataobjectives, since apparently classic is different than cata for that particular quest (there's another version in cata where mobs dont spawn until you close the portal, but in classic it's not needed).

PR #6870, opened 2025-02-05
Update zhCN/zhTW.

NORPG:
Complete the missing translations in zhCN.


Localization/Translations/Options/General.lua
```diff
@@ -213,7 +213,7 @@ local generalOptionsLocales = {
         ["koKR"] = "플레이어 레벨 오프셋",
         ["esMX"] = "Compensación de nivel",
         ["enUS"] = true,
-        ["zhCN"] = false,
+        ["zhCN"] = "玩家等级偏移",
```
NORPG, line 216: I'm not sure what this offset means, so I translate it like this for now.

BreakBB: By default Questie uses `GetQuestGreenRange` to determine which quests to show. Everything that is below the range of that API is considered "trivial" and therefore not shown.

You can however use this setting to change this behavior and instead use a manual offset.

BreakBB:
Small advice @NORPG you should always work on a branch on your fork. Otherwise it will be very hard for you to keep your work synced with the Questie repo.

NORPG:
> Small advice @NORPG you should always work on a branch on your fork. Otherwise it will be very hard for you to keep your work synced with the Questie repo.

Thanks for your suggestion, but I develop directly on VScode. In the case of daul remote, I usually pull and rebase the changes of upstream and push them to my fork, so it is not a problem.
Of course creating a new branch is also a good idea :) .

BreakBB (approved):

PR #6892, opened 2025-02-15
[Locale] Fix q432/q433 in zhTW

NORPG:
## Proposed changes

- The localization of q432 and q433 in zhTW is incorrect. This patch updates the localization and fixes this error.


anon1231823:
@NORPG Just as a heads up in the future, Questie isn't using the second argument (Quest description) at all. So if you want to save yourself the pain of fixing the quests, I would focus efforts on the Quest Name and Quest Objective

PR #6900, opened 2025-02-18
adding support for Ace3 treeWidth

MyTechnoHunter:
initial stages. support is there but we are not using tree correctly. instead we are using tab with a container more work needed

## Issue references

Fixes #

## Proposed changes

-
-


MyTechnoHunter:
this is the original post from the author

applebanditwow

Hello,

 

I wanted to change the width of a tree group container in AceConfig, but found that there was no option to do this.

 

I have written the code below to add this feature. Is it possible to add this to Ace3?

 

### AceConfigDialog-3.0.lua, after line 1721 ###

if (group.treeWidth ~= nil) then
	tree:SetTreeWidth(group.treeWidth)
end


### AceConfigRegistry-3.0.lua, update typedkeys.group ###

group={
	args=istable,
	plugins=opttable,
	inline=optbool,
	cmdInline=optbool,
	guiInline=optbool,
	dropdownInline=optbool,
	dialogInline=optbool,
	childGroups=optstring,
	treeWidth=optnumber,
},


### Example of usage ###

myConfigDialog = {
	name = "myAddon",
	handler = myAddon,
	type = "group",
	childGroups = "tree",
	treeWidth = 210,
	args = {}
}

MyTechnoHunter:
Capture

BreakBB:
This is related to #6901 - I don't think we need this PR on its own though. What do you think @MyTechnoHunter ?

MyTechnoHunter:
> This is related to #6901 - I don't think we need this PR on its own though. What do you think @MyTechnoHunter ?

It's not ready for sure.. keep as a draft?

Issue #7064, opened 2025-07-15
Four Winds Phasing To Do List


coldpasta123:
### Description

NPC 56312 at {78.19,32.80} after completing {29916,29917}. On acceptance moves to {74.68,34.59}

### Questie version

v11

### Game flavor

MoP

BreakBB:
It's a great idea to start listing stuff in here and not add time to it now. We can implement the phasing logic later on 👍🏻

PR #7093, opened 2025-07-24
More Pandaria Quest Fixes

coldpasta123:
## Issue references

#7064 

## Proposed changes

-
-


Modules/Phasing.lua
```diff
@@ -320,6 +320,10 @@ local phases = {
     SUNA_AT_CAMP_OSUL = 1248,
     BAN_AT_OUTPOST = 1249,
     BAN_AT_CAMP_OSUL = 1250,
+    BEFORE_MANTID_INVASION = 1251,
+    AFTER_MANTID_INVASION = 1252,
```
BreakBB, line 324: This phasing ID is not handled in the code below, so there is logic missing to it.

BreakBB (changes requested):

Modules/Phasing.lua
coldpasta123: @BreakBB added

PR #7106, opened 2025-07-29
More mop daily fixes

coldpasta123:
## Issue references

Fixes #

## Proposed changes

-
-


Database/Corrections/mopNPCFixes.lua
```diff
@@ -1370,6 +1370,10 @@ function MopNpcFixes.Load()
                 {44.41,16.8,phases.SKEER_IN_KLAXXI_VEES}
             }},
         },
+        [63123] = { -- Mogu Ritual Destroyed Credit
```
BreakBB, line 1373: The actual NPC IDs are 63087, 63088, 63089 and 63090. So I would use a killCreditObjective. For quest `[30283] = { -- A Smashing Impression` that is.

Database/Corrections/mopNPCFixes.lua
```diff
@@ -1814,6 +1818,11 @@ function MopNpcFixes.Load()
                 [zoneIDs.ISLE_OF_GIANTS] = {{67,75.2},{69.6,71.4},{69.8,71.6}}},
                 [zoneIDs.ISLE_OF_THUNDER] = {{50.4,42.8},{50.8,43.4},{51,43.6}},
         },
+        [70414] = { -- Skumblade Pillager
+            [npcKeys.name] = "Skumblade Pillager",
+            -- [npcKeys.spawns] = {[zoneIDs.VALE_OF_ETERNAL_BLOSSOMS] = {{}}}, Unknown location, not part of quest causing error
+            [npcKeys.zoneID] = zoneIDs.VALE_OF_ETERNAL_BLOSSOMS,
```
BreakBB, line 1824: wowhead says this NPC is added with 5.2.0 and is located in the "Troves of the Thunder King" Scenario 🤔 Just remove this correction and since the NPC is no longer listed in the objectives.

However I need to adjust the validations, to make sure the killCreditObjectives also don't have missing entries (like this one).

Database/Corrections/mopNPCFixes.lua
coldpasta123: @BreakBB I'm not sure I agree. The objective isn't to interact with the objects, but to be at the location when the event is triggered. The trigger for the event might be the objects (hence extra objective), but it is sufficient to be there when someone else triggers it and not actually do anything.

PR #7111, opened 2025-07-30
Mop quest fixes

coldpasta123:
## Issue references

Fixes #

## Proposed changes

-
-


Localization/Translations/ExtraObjectives/MopObjectives.lua
```diff
@@ -940,6 +964,7 @@ local mopObjectiveLocales = {
     },
 }
 
+
```
BreakBB:
```suggestion
```

PR #7139, opened 2025-08-12
Update tillers 2nd locations

puggan:
## Issue references

Fixes #7138

## Proposed changes

Add dual locations for the tillers thats missing it.

## Source
Location data copied from wowhead:

57298: Fung coordinates: 48.2 33.8 or 52.8 51.6
57402: Haohan coordinates: 44.4 34.0 or 53.0 51.6
58647: Ellas coordinates: 31.6 58.0 or 53.0 51.6
58705: Fish coordinates: 41.6 30.0 or 52.8 51.8
58707: Hillpaw coordinates: 31.0 53.0 or 53.0 51.8
58708: Sho coordinates: 29.4 30.4 or 53.0 52.0
58709: Chee coordinates: 34.4 46.8 or 53.0 52.0
58761: Tina coordinates: 45.0 33.8 or 53.0 51.8

PR #7140, opened 2025-08-12
Update tillers, home locations only

### LINKED ISSUE #7138: Ella (and other) tiller locations
### Description

The tillers have dual locations, they can be at home or at the market.

For some of them we have both locations, some have only the home location, and Ella have only her market location.

This make Ellas home harder to find on the map.

I don't know if we want the dual location registered for all, or if we only want the home-location registered to not make the market to crowded with quest-markers.

I'll make a PR where I add the dual location for all tillers.

### Questie version

v11.2.13

### Game flavor

MoP

puggan:
## Issue references

Fixes #7138 (Alternative solution to PR #7139)

## Proposed changes

Move Ella to her home-location, and remove the market location for the other Tillers,
so the quest-marker end up at there home instead of all crowed together at the market.

(If crowded together at the market is preferred, use PR #7139 instead)


puggan:
Better solution in: https://github.com/Questie/Questie/commit/d46d23c109ea64de6f553a171d38f2e26edba685

PR #7165, opened 2025-09-13
Attempt to handle Nomi's hidden faction

Cabro:
Nomi


Modules/QuestieReputation.lua
```diff
@@ -155,6 +180,9 @@ function QuestieReputation.GetReputationReward(questId)
     local rewards = {}
     local knowsMrPopularityRank1 = IsSpellKnown(78634)
     local knowsMrPopularityRank2 = IsSpellKnown(78635)
+    local hasDarkmoonFaireReputationBuff = IsSpellKnown(46668)
```
BreakBB: This won't work for the DMF-Buff. `IsSpellKnown` only checks for spells that show up in the spell book.

PR #7202, opened 2025-10-05
[feature] Add new dropdown to see all quests in all zones

### LINKED ISSUE #6930: Re-open last selected quest when re-opening the Journey with "Quests by Zone"
### Description

This was requested [on Discord](https://discord.com/channels/263036731165638656/909847511089418280/1347082001659985972) and would actually quite nice.

Questie should not only remember the last selected Journey tab, when re-opening it, but should also select the previously selected quest. So that the Journey frame is in the exact same state as when the user closed it the last time.

### LINKED ISSUE #6958: "Quests by Zone" functionality but for ALL zones
### Description

So I mainly play WoW Classic anniversary, and I'm a completionist and like to do every quest in the world before they turn gray if I can. While Questie is useful for that, it still always reaches a point where I have to keep hopping though zones to find the lowest level quest so I can switch zones to do the lowest ones before they turn gray.

It would be helpful to have something similar to where when you use the Quests by Zone window, you can select All Zones from the dropdown and now you have all quests in the world sorted by level. You could still have Available, Missing Pre Quest, etc.

Combined with that it would probably be better to turn Event Quests into one of the filters so that way it's optional if you want to include event quests in that same quest flow.

Sommos:
## Issue references

Fixes #6958
Fixes #6930

## Proposed changes

- Add a new dropdown that allows a user to search by 'All Zones' (Collects all quests in all zones and displays it as a list to the user), or 'By Continent' where they can continue to use the existing functionality, and sort journey quest list by zone.
- Add persistence so that when a user selects a quest within a zone, closes the journey window, and later reopens it, their selected quest remains saved and restored.
- Modify sort algorithm to include secondary sorting using suffixes when quest level remains the same between two indexes.

Tested only in Classic Era, ~~NOT Mists of Pandaria Classic~~.

This now works in MoP Classic (See 3rd & 4th image)

## Screenshots

image
image
image
image

### DIFF
--- a/Modules/Libs/QuestieLib.lua
+++ b/Modules/Libs/QuestieLib.lua
@@ -398,37 +398,38 @@ function QuestieLib:SanitizePattern(pattern)
     return sanitize_cache[pattern]
 end
 
+local suffixPriority = {
+    [""] = 1, -- No suffix (normal quests) - should come first
+    ["+"] = 2, -- Elite
+    ["S"] = 3, -- Scenario
+    ["D"] = 4, -- Dungeon
+    ["H"] = 5, -- Heroic
+    ["R"] = 6, -- Raid
+    ["++"] = 7, -- Legendary
+    ["A"] = 8, -- Account
+    ["C"] = 9, -- Celestial
+}
+
+local function compareQuestsByLevelAndType(a, b)
+    if a[1] ~= b[1] then
+        return a[1] < b[1]
+    end
+
+    -- if levels are the same, compare by suffix priority
+    local suffixA = a[3] or ""
+    local suffixB = b[3] or ""
+    local priorityA = suffixPriority[suffixA] or 999
+    local priorityB = suffixPriority[suffixB] or 999
+
+    if priorityA ~= priorityB then
+        return priorityA < priorityB
+    end
+
+    return a[2] < b[2]
+end
+
 function QuestieLib:SortQuestIDsByLevel(quests)
     local sortedQuestsByLevel = {}
-    local suffixPriority = {
-        [""] = 1, -- No suffix (normal quests) - should come first
-        ["+"] = 2, -- Elite
-        ["S"] = 3, -- Scenario
-        ["D"] = 4, -- Dungeon
-        ["H"] = 5, -- Heroic
-        ["R"] = 6, -- Raid
-        ["++"] = 7, -- Legendary
-        ["A"] = 8, -- Account
-        ["C"] = 9, -- Celestial
-    }
-
-    local function compareQuestsByLevelAndType(a, b)
-        if a[1] ~= b[1] then
-            return a[1] < b[1]
-        end
-
-        -- if levels are the same, compare by suffix priority
-        local suffixA = a[3] or ""
-        local suffixB = b[3] or ""
-        local priorityA = suffixPriority[suffixA] or 999
-        local priorityB = suffixPriority[suffixB] or 999
-
-        if priorityA ~= priorityB then
-            return priorityA < priorityB
-        end
-
-        return a[2] < b[2]
-    end
 
     for q in pairs(quests) do
         local questLevel, _ = QuestieLib.GetTbcLevel(q)


Modules/Journey/tabs/QuestsByZone/QuestsByZoneTab.lua
```diff
@@ -128,7 +161,181 @@ _CreateZoneDropdown = function()
...
+                -- multiple pre quests are missing
+                elseif not QuestieDB:IsPreQuestGroupFulfilled(preQuestGroup) then
+                    table.insert(allZoneTree[2].children, temp)
+                    prequestMissingCounter = prequestMissingCounter + 1
+                -- repeatable quests
+                elseif QuestieDB.IsRepeatable(questId) then
+                    table.insert(allZoneTree[4].children, temp)
+                    repeatableCounter = repeatableCounter + 1
+                -- quests which require you to NOT have learned a spell (most likely a fake quest for SoD runes)
+                elseif requiredSpell and requiredSpell < 0 and (IsSpellKnownOrOverridesKnown(requiredSpell) or IsPlayerSpell(requiredSpell)) then
```
BreakBB: This line is raising an error:

> Modules/Journey/tabs/QuestsByZone/QuestsByZoneTab.lua:295: Usage: IsSpellKnownOrOverridesKnown(spellID[, isPet])

The problem is that `requiredSpell` is negative 😅 Which is why the original code uses `math.abs`

Modules/Journey/tabs/QuestsByZone/QuestsByZoneTab.lua
```diff
@@ -147,7 +354,6 @@ _HandleContinentSelection = function(key, _)
                 end
             end
         end
-        local text = l10n('Select Your Profession')
```
BreakBB: Why was this removed?

Modules/Libs/QuestieLib.lua
```diff
@@ -400,16 +400,71 @@ end
...
+                return "H"
+            elseif questTagId == questTagIds.SCENARIO then
+                return "S"
+            else
+                return ""
+            end
+        else
+            local char = stringSub(questTagName, 1, 1)
+            return char
+        end
```
BreakBB: Instead of copying most parts of `QuestieLib:GetLevelString` I suggest to add a new function to the `QuestieLib` which has more of less the code you have here. And then use that new function here and in `QuestieLib:GetLevelString`.

Otherwise we'll surely miss one of the places once something needs to be changed.

Modules/Libs/QuestieLib.lua
```diff
@@ -400,16 +400,71 @@ end
...
+    end
+
+    local function compareQuestsByLevelAndType(a, b)
```
BreakBB: This is long overdue, good job! 💪🏻

Modules/Journey/tabs/QuestsByZone/QuestsByZoneTab.lua
```diff
@@ -128,7 +161,181 @@ _CreateZoneDropdown = function()
     return dropdown
 end
 
+_HandleAllZonesSelection = function()
```
BreakBB, line 150: Do we actually need this code all over again? On a short look it looks very similar to the code in `QuestsByZone.lua`. If there are major differences, then maybe some parts that are the same can be extracted and be reused?

Modules/Journey/QuestieJourneyUtils.lua
```diff
@@ -14,6 +14,10 @@ local GetItemIcon = C_Item.GetItemIconByID or GetItemIcon
 local AceGUI = LibStub("AceGUI-3.0")
 
 function QuestieJourneyUtils:GetSortedZoneKeys(zones)
+    if type(zones) ~= "table" then
+        return {}
+    end
```
BreakBB: Why is this needed?

Modules/Journey/tabs/QuestsByZone/QuestsByZone.lua
```diff
@@ -38,6 +37,58 @@ function _QuestieJourney.questsByZone:ManageTree(container, zoneTree)
...
+                            master:AddChild(scrollFrame)
+
+                            _QuestieJourney:DrawQuestDetailsFrame(scrollFrame, quest)
+                        end
+                    end
+                end)
+            else
+                _QuestieJourney.lastZoneSelection[3] = nil
+            end
+        end
```
BreakBB: Maybe this can be a function, so we have a proper name for the code? Just to make it clear what this code is for (which is re-selecting the previous selection, right?)

Also this is very similar now to the code in `_QuestieJourney:RestoreQuestSelection` 🤔

Localization/Translations/Journey/QuestsByZone.lua
```diff
@@ -170,6 +170,42 @@ local questsByZoneLocales = {
         ["zhCN"] = "前置任务",
         ["zhTW"] = "前置任務",
     },
+    ["All Zones"] = {
```
BreakBB: Remember to ask for translations once it's more clear what is needed and what not.

Modules/Journey/QuestieJourneyUtils.lua
Sommos: It's completely unnecessary - Just a holdover from testing something else.

6bab331

Modules/Journey/tabs/QuestsByZone/QuestsByZoneTab.lua
Sommos: 6bab331

Modules/Journey/tabs/QuestsByZone/QuestsByZoneTab.lua
Sommos: It never gets assigned or read after this, so it's just dead code, no?

Sommos: d27d697

Modules/Libs/QuestieLib.lua
Sommos: d27d697

Modules/Journey/tabs/QuestsByZone/QuestsByZoneTab.lua
Sommos: d27d697

Modules/Journey/tabs/QuestsByZone/QuestsByZone.lua
Sommos: d27d697

Modules/Journey/tabs/QuestsByZone/QuestsByZone.lua
```diff
@@ -25,6 +24,57 @@ local l10n = QuestieLoader:ImportModule("l10n")
...
+                scrollFrame:SetFullHeight(true)
+                master:AddChild(scrollFrame)
+
+                _QuestieJourney:DrawQuestDetailsFrame(scrollFrame, quest)
+            end
+        end)
+    else
+        _QuestieJourney.lastZoneSelection[3] = nil
+    end
+end
```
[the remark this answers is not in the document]
Sommos: 586eca6

Modules/Libs/QuestieLib.lua
```diff
@@ -226,49 +226,54 @@ function QuestieLib.GetTbcLevel(questId, playerLevel)
...
 
-    return "[" .. levelString .. char .. "] "
+---@param questId QuestId
+---@param level Level @The quest level
+---@return string levelString @String of format "[40+]"
+function QuestieLib:GetLevelString(questId, level)
+    local levelString = tostring(level)
+    local suffix = QuestieLib:GetQuestTypeSuffix(questId)
+    return "[" .. levelString .. suffix .. "] "
 end
```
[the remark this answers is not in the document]
Sommos: 64ca272

Modules/Journey/tabs/QuestsByZone/QuestsByZoneTab.lua
```diff
@@ -128,6 +161,178 @@ _CreateZoneDropdown = function()
...
+                    table.insert(allZoneTree[3].children, temp)
+                    completedCounter = completedCounter + 1
+                -- available quests
+                else
+                    table.insert(allZoneTree[1].children, temp)
+                    availableCounter = availableCounter + 1
+                end
+            end
+        end
+    end
```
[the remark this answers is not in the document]
Sommos: 64ca272

Modules/Journey/tabs/QuestsByZone/QuestsByZoneTab.lua
```diff
@@ -128,6 +161,178 @@ _CreateZoneDropdown = function()
...
+
+    local totalCounter = availableCounter + completedCounter + prequestMissingCounter
+    allZoneTree[1].text = allZoneTree[1].text .. ' [ '..  availableCounter ..'/'.. totalCounter ..' ]'
+    allZoneTree[2].text = allZoneTree[2].text .. ' [ '..  prequestMissingCounter ..'/'.. totalCounter ..' ]'
+    allZoneTree[3].text = allZoneTree[3].text .. ' [ '..  completedCounter ..'/'.. totalCounter ..' ]'
+    allZoneTree[4].text = allZoneTree[4].text .. ' [ '..  repeatableCounter ..' ]'
+    allZoneTree[5].text = allZoneTree[5].text .. ' [ '..  unobtainableCounter ..' ]'
+
```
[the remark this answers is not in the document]
Sommos: 64ca272

Modules/Libs/QuestieLib.lua
```diff
@@ -402,16 +406,42 @@ end
...
+        [""] = 1,    -- No suffix (normal quests) - should come first
+        ["+"] = 2,   -- Elite
+        ["S"] = 3,   -- Scenario
+        ["D"] = 4,   -- Dungeon
+        ["H"] = 5,   -- Heroic
+        ["R"] = 6,   -- Raid
+        ["++"] = 7,  -- Legendary
+        ["A"] = 8, -- Account
+        ["C"] = 9, -- Celestial
+    }
```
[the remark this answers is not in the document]
Sommos: This is expected behaviour.

The 'Heroic' suffix refers to heroic dungeons, not heroic raids, so placing it before raids matches normal player progression.

The suffix priority is also a UI only, stylistic choice, not a strict difficulty ranking. In the 'Quests By Zone' view we sort by level first, then by suffix so players see normal -> scenario -> dungeon, which is the intended and more readable ordering.

Modules/Libs/QuestieLib.lua
```diff
@@ -226,49 +226,50 @@ function QuestieLib.GetTbcLevel(questId, playerLevel)
...
+    elseif questTagId == questTagIds.ACCOUNT then
+        return "A"
+    elseif questTagId == questTagIds.CELESTIAL then
+        return "C"
+    end
+
+    -- Fallback: use first character of quest tag name for unknown tags
+    -- This preserves backward compatibility with existing UI/tests
+    return stringSub(questTagName, 1, 1)
+end
```
Sommos, line 261: 255c1ed

BreakBB:
I started this on MoP and it errors right away for me.

> 2x ...odules/Journey/tabs/QuestsByZone/QuestsByZoneTab.lua:208: attempt to index upvalue 'contDropdown' (a nil value)
[tail call]: ?
[Questie/Modules/Journey/tabs/QuestsByZone/QuestsByZoneTab.lua]:208: in function <...odules/Journey/tabs/QuestsByZone/QuestsByZoneTab.lua:169>
[Questie/Modules/Journey/tabs/QuestsByZone/QuestsByZoneTab.lua]:96: in function <...odules/Journey/tabs/QuestsByZone/QuestsByZoneTab.lua:86>
[Questie/Modules/Journey/tabs/QuestsByZone/QuestsByZoneTab.lua]:51: in function 'DrawTab'
[Questie/Modules/Journey/QuestieJourneyPrivates.lua]:30: in function 'HandleTabChange'
[Questie/Modules/Journey/QuestieJourney.lua]:153: in function <...ceQuestie/Modules/Journey/QuestieJourney.lua:153>

The problem is, that `contDropdown.frame:Hide()` inside `_HandleAllZonesSelection` is called before `contDropdown = _CreateContinentDropdown()`.

anon1231823:
This PR is nearly ready to go. There are a couple things missing.

1. The Scourge Invasion event quests are still hidden, so it's not showing up at all. It should be in the unobtainable category.
2. Fishing Tournament quests show as available, and technically should be unobtainable. The quests 8228 and 8229 should be available on saturday before the tournament, and the tournament quests during the tournament.
3. There is a discussion if when you complete a follow up to a breadcrumb quest if the breadcrumb should go into the unobtainable category. Currently, it is marked as complete. But it is not actually complete. If you do `/run print(C_QuestLog.IsQuestFlaggedCompleted(nnnn))` for the breadcrumb quest after completing it's follow up, it would show as `false`.

BreakBB:
>     1. The Scourge Invasion event quests are still hidden, so it's not showing up at all. It should be in the unobtainable category.

I am not sure about that. We don't show them in the current QbZ logic, do we?

>     2. Fishing Tournament quests show as available, and technically should be unobtainable. The quests 8228 and 8229 should be available on saturday before the tournament, and the tournament quests during the tournament.

I think available is more correct than unobtainable in this case. That's also where they show right now in the Fishing professions list in QbZ, don't they?

>     3. There is a discussion if when you complete a follow up to a breadcrumb quest if the breadcrumb should go into the unobtainable category. Currently, it is marked as complete. But it is not actually complete. If you do `/run print(C_QuestLog.IsQuestFlaggedCompleted(nnnn))` for the breadcrumb quest after completing it's follow up, it would show as `false`.

Interesting idea. But that is something that shouldn't be part of this PR as that also would need to be reflected in QbZ.

I'd love to get this finally merged and free @Sommos from this hell of a PR 😆

Modules/Journey/tabs/QuestsByZone/QuestsByZone.lua
```diff
@@ -247,6 +338,22 @@ function _QuestieJourney.questsByZone:CollectZoneQuests(zoneId)
...
+                    tinsert(zoneTree[6].children, temp)
+                    unobtainableCounter = unobtainableCounter + 1
+                -- AQ War Effort quests (one-time world event that has ended for all realms)
+                elseif QuestieQuestBlacklist.AQWarEffortQuests[questId] then
+                    tinsert(zoneTree[6].children, temp)
+                    unobtainableCounter = unobtainableCounter + 1
+                -- Scourge Invasion quests (one-time world event that has ended for all realms)
+                elseif QuestieQuestBlacklist.ScourgeInvasionQuests[questId] then
+                    tinsert(zoneTree[6].children, temp)
+                    unobtainableCounter = unobtainableCounter + 1
```
BreakBB: I don't think we should handle this case specially at all.

@Sommos what is your take on it?

Sommos: We should hold off on that for the moment. When we don't have special tracking for these kinds of things and SoD still has them available, we should avoid mucking around with it too much.

2afa1ff

anon1231823: > I don't think we should handle this case specially at all.
> 
> @Sommos what is your take on it?

So why treat AQ war effort differently from scourge invasion?

Modules/Journey/tabs/QuestsByZone/QuestsByZone.lua
```diff
@@ -247,6 +338,18 @@ function _QuestieJourney.questsByZone:CollectZoneQuests(zoneId)
...
+                -- Event quests where the event is not currently active
+                elseif QuestieEvent.IsEventQuest(questId) and not QuestieEvent.IsEventActiveForQuest(questId) then
+                    tinsert(zoneTree[6].children, temp)
+                    unobtainableCounter = unobtainableCounter + 1
+                -- AQ War Effort quests (one-time world event that has ended for all realms)
+                elseif QuestieQuestBlacklist.AQWarEffortQuests[questId] then
+                    tinsert(zoneTree[6].children, temp)
+                    unobtainableCounter = unobtainableCounter + 1
```
BreakBB, line 352: War efforts are still available for SoD

```suggestion
                -- AQ War Effort quests (one-time world event that has ended for all realms)
                elseif (not Questie.IsSoD) and QuestieQuestBlacklist.AQWarEffortQuests[questId] then
                    tinsert(zoneTree[6].children, temp)
                    unobtainableCounter = unobtainableCounter + 1
```

Modules/Libs/QuestieLib.lua
```diff
@@ -162,15 +162,32 @@ end
...
+    {1, 0.25, 1}, --Pink Flamingo
+    {1, 1, 0}, --Yellow
+    {0.16015625, 0.65234375, 0}, --Slimy Green
+    {0, 0.66015625, 1}, --Deep Sky Blue
+    {0.87109375, 0.87109375, 0.56640625}, --Primrose
+    {0, 0.5859375, 0}, --Vine Green --G67 default
+    {0, 0.3, 1}, --Navy Blue
+    {0, 0.97265625, 0}, --Lime
+    {0, 1, 1}, --Aqua
+    {1, 0.1484375, 0}, --Scarlet
```
BreakBB: Whoops, a merge went wrong here:


```suggestion
    -- Light (200)         Standard (500)         -- Family
    {0.99, 0.73, 0.73},    {0.94, 0.19, 0.19},    -- Red
    {0.99, 0.81, 0.59},    {0.98, 0.46, 0.05},    -- Orange
    {0.99, 0.93, 0.54},    {0.92, 0.68, 0.05},    -- Yellow
    {0.73, 0.96, 0.80},    {0.13, 0.77, 0.36},    -- Green
    {0.75, 0.87, 0.99},    {0.23, 0.55, 0.94},    -- Blue
    {0.78, 0.82, 0.99},    {0.39, 0.45, 0.94},    -- Indigo
    {0.87, 0.82, 1.00},    {0.55, 0.35, 0.96},    -- Violet
    {0.99, 0.76, 0.89},    {0.93, 0.16, 0.55},    -- Pink
```

Modules/Libs/QuestieLib.lua
```diff
@@ -217,49 +234,56 @@ function QuestieLib.GetTbcLevel(questId, playerLevel)
...
+        return stringSub(questTagName, 1, 1)
     end
+end
 
-    return "[" .. levelString .. char .. "] "
+---@param questId QuestId
+---@param level Level @The quest level
+---@return string levelString @String of format "[40+]"
+function QuestieLib:GetLevelString(questId, level)
+    local levelString = tostring(level or "?")
```
BreakBB: There should never be a case where level is `nil`, is there?

```suggestion
    local levelString = tostring(level)
```

Modules/Libs/QuestieLib.lua
```diff
@@ -393,16 +417,42 @@ end
...
+        local priorityB = suffixPriority[suffixB] or 999
+
+        if priorityA ~= priorityB then
+            return priorityA < priorityB
+        end
 
-    local function compareTablesByIndex(a, b)
-        return a[1] < b[1]
+        return a[2] < b[2]
     end
```
BreakBB: We can move this outside of `QuestieLib:SortQuestIDsByLevel` to not re-define it every time the function is called. Makes `QuestieLib:SortQuestIDsByLevel` more readable as well.

BreakBB:
The feature seems to work fine 👍🏻 🥳

PR #7205, opened 2025-10-06
[feature] Add dungeon names alongside quests

### LINKED ISSUE #6953: Flag dungeons quests with the dungeon they're associated with
### Description

It would be helpful, specifically in classic wow to know which dungeons quests are associated with. If possibly, add the name of the dungeon in parenthesis or something.

Sommos:
## Issue references

Fixes #6953

## Proposed changes

- Add ability to see what dungeon is associated with a quest when holding shift and hovering over a quest on the map.

### DIFF
--- a/Localization/Translations/TrackerUI.lua
+++ b/Localization/Translations/TrackerUI.lua
@@ -509,6 +509,18 @@ local trackerUILocales = {
         ["zhCN"] = "暴雪计时器启动！",
         ["zhTW"] = "暴雪計時器啟動！",
     },
+    ["Dungeon"] = {
+        ["enUS"] = true,
+        ["deDE"] = false,
+        ["esES"] = false,
+        ["esMX"] = false,
+        ["frFR"] = false,
+        ["koKR"] = false,
+        ["ptBR"] = false,
+        ["ruRU"] = false,
+        ["zhCN"] = false,
+        ["zhTW"] = false,
+    },
 }
 
 for k, v in pairs(trackerUILocales) do
--- a/Modules/Tooltips/MapIconTooltip.lua
+++ b/Modules/Tooltips/MapIconTooltip.lua
@@ -21,6 +21,8 @@ local QuestieComms = QuestieLoader:ImportModule("QuestieComms")
 local l10n = QuestieLoader:ImportModule("l10n")
 ---@type QuestXP
 local QuestXP = QuestieLoader:ImportModule("QuestXP")
+---@type ZoneDB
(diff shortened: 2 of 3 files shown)


Localization/Translations/TrackerUI.lua
```diff
@@ -509,6 +509,18 @@ local trackerUILocales = {
         ["zhCN"] = "暴雪计时器启动！",
         ["zhTW"] = "暴雪計時器啟動！",
     },
+    ["Dungeon"] = {
```
BreakBB: We don't have a translation for this yet? 🤔 

It could also be this string is available as global. Give `/dump DUNGEON` a shot ingame.

BreakBB:
@Cabro any thoughts on this PR? This is really a great addition to the tooltips for dungeon quests imo. I am just not 100% sure about the position

Localization/Translations/TrackerUI.lua
Sommos: We already have this in QuestStates.lua. Good catch! :)

Fixed in: https://github.com/Questie/Questie/pull/7205/commits/1a1a188d1782cdf7c6b413187d24a3ae56892491

Cabro:
If anything, the dungeon should absolutely be stated in QuestLinks

BreakBB:
> If anything, the dungeon should absolutely be stated in QuestLinks

So you are against adding it to the tooltip? Or want it to **also** show in quest links?

Cabro:
Also show it **especially** in QuestLinks!

Modules/Tooltips/MapIconTooltip.lua
```diff
@@ -241,6 +243,16 @@ function MapIconTooltip:Show()
                             self:AddDoubleLine(TRANSPARENT_ICON_TEXTURE .. " " .. questData.title, rewardString, 1, 1, 1, 1, 1, 0);
                         end
                     end
+                    -- Add dungeon information if this is a dungeon quest
+                    if shift and quest then
+                        local zoneOrSort = quest.zoneOrSort
+                        if zoneOrSort and zoneOrSort > 0 then
+                            local dungeonName = ZoneDB:GetDungeonName(zoneOrSort)
+                            if dungeonName then
+                                self:AddLine("  " .. l10n("Dungeon") .. ": " .. dungeonName, 0.7, 0.7, 0.7)
```
BreakBB: In French there is an additional space before the `:`. It might be worth to extract a local function which can be reused for this case and the one below.

Sommos: 10353db

PR #7213, opened 2025-10-15
[feature] Add moonwell tracking in tailoring submenu

### LINKED ISSUE #6925: Show Moonwell locations on map/minimap for Tailors
### Description

Not sure if this is something that's within the scope of Questie, I know I've struggled to find an addon/weakaura that does this, still haven't found one.
We could have a Tailoring sub-menu in the Questie main dropdown with a checkbox for Moonwell locations. This might be applicable to other professions too maybe, like Blacksmithing anvils? I've never trained Blacksmithing so I don't know if this already exists in Questie.

Sommos:
## Issue references

Fixes #6925 

## Proposed changes

- Add a 'Moonwell' toggle option under the new Tailoring submenu to track the location of Moonwells used in crafting Mooncloth

Tested working on Classic Era

## Screenshots

image

image

### DIFF
--- a/Localization/Translations/Moonwell.lua
+++ b/Localization/Translations/Moonwell.lua
@@ -5,14 +5,14 @@ local moonwellLocales = {
     ["Moonwell"] = {
         ["enUS"] = true,
         ["deDE"] = "Mondbrunnen",
-        ["esES"] = false,
-        ["esMX"] = false,
-        ["frFR"] = false,
-        ["koKR"] = false,
-        ["ptBR"] = false,
-        ["ruRU"] = false,
-        ["zhCN"] = false,
-        ["zhTW"] = false,
+        ["esES"] = "Poza de la Luna",
+        ["esMX"] = "Poza de la Luna",
+        ["frFR"] = "Puits de lune",
+        ["koKR"] = "달샘",
+        ["ptBR"] = "Poço Lunar",
+        ["ruRU"] = "Лунный колодец",
+        ["zhCN"] = "月亮井",
+        ["zhTW"] = "月井",
     },
 }
 


anon1231823:
I think it would be cool if the icon is the Moonwell icon from warcraft 3

https://wowpedia.fandom.com/wiki/Moon_Well_(Warcraft_III)

I wonder if that icon is accessible

BreakBB:
> I think it would be cool if the icon is the Moonwell icon from warcraft 3

While that would be a cool reference, I don't think that is actually a good icon. Especially since not every WoW player actually played WC3.

So I vote for the Mooncloth icon that is currently used.

Localization/Translations/Moonwells.lua
```diff
@@ -0,0 +1,33 @@
+---@type l10n
+local l10n = QuestieLoader:ImportModule("l10n")
+
+local moonwellLocales = {
+    ["Moonwell"] = {
+        ["enUS"] = true,
+        ["deDE"] = false,
```
BreakBB:
```suggestion
        ["deDE"] = "Mondbrunnen",
```

Localization/Translations/Moonwells.lua
```diff
@@ -0,0 +1,33 @@
...
+        ["frFR"] = false,
+        ["koKR"] = false,
+        ["ptBR"] = false,
+        ["ruRU"] = false,
+        ["zhCN"] = false,
+        ["zhTW"] = false,
+    },
+    ["Moonwells"] = {
+        ["enUS"] = true,
+        ["deDE"] = false,
```
BreakBB:
```suggestion
        ["deDE"] = "Mondbrunnen",
```

Modules/Map/Moonwells.lua
```diff
@@ -0,0 +1,59 @@
+---@class Moonwells
+local Moonwells = QuestieLoader:CreateModule("Moonwells")
+
+---@type l10n
+local l10n = QuestieLoader:ImportModule("l10n")
+---@type QuestieMap
+local QuestieMap = QuestieLoader:ImportModule("QuestieMap")
+
+Moonwells.data = {
```
BreakBB: In general I prefer to use `[<id>] = true` for these kind of tables, so that it's a `Set` instead of a basic array. The benefits I see are:

1. You make sure each ID is only once in the list
2. You can use `pairs` instead of `ipairs` which is _slightly_ more performant (not relevant for this small table though 😅 )

Modules/Map/QuestieMap.lua
```diff
@@ -20,6 +20,8 @@ local ZoneDB = QuestieLoader:ImportModule("ZoneDB")
 local l10n = QuestieLoader:ImportModule("l10n")
 ---@type WeaponMasterSkills
 local WeaponMasterSkills = QuestieLoader:ImportModule("WeaponMasterSkills")
+---@type Moonwells
+local Moonwells = QuestieLoader:ImportModule("Moonwells")
```
BreakBB: This is no longer needed right? Also the code change below seems obsolete.

Modules/QuestieMenu/QuestieMenu.lua
```diff
@@ -1,4 +1,4 @@
----@class QuestieMenu
+---@type QuestieMenu
```
BreakBB: This needs to be reverted. `class` is used as **declaration**, while `type` references the previously declared `class`.

BreakBB (commented):
Great PR 👍🏻

Sommos:
> I think it would be cool if the icon is the Moonwell icon from warcraft 3
> 
> https://wowpedia.fandom.com/wiki/Moon_Well_(Warcraft_III)
> 
> I wonder if that icon is accessible

I've managed to extract the Warcraft 3 moonwell icon and got it working with Questie:

image

I'm going to agree with @BreakBB that it's not a great icon for this purpose and would prefer to stay with the current mooncloth implementation.

Localization/Translations/Moonwells.lua
Sommos: Added in: https://github.com/Questie/Questie/pull/7213/commits/8659c3d6d5ea922c85f661e702a4ab57eefc9e95

Localization/Translations/Moonwells.lua
Sommos: Added in: https://github.com/Questie/Questie/pull/7213/commits/8659c3d6d5ea922c85f661e702a4ab57eefc9e95

Modules/Map/QuestieMap.lua
Sommos: 100% correct, the changes to this file are not needed. 

Fixed in: https://github.com/Questie/Questie/pull/7213/commits/c993f3557332d0df0f4cd6f0b7c323821edc4dd5

Modules/QuestieMenu/QuestieMenu.lua
Sommos: Ahhh, I didn't catch this. Unsure how that managed to get changed!

Fixed in: https://github.com/Questie/Questie/pull/7213/commits/c993f3557332d0df0f4cd6f0b7c323821edc4dd5

Modules/Map/Moonwells.lua
Sommos: 100%. Didn't think of this!

Fixed in: https://github.com/Questie/Questie/pull/7213/commits/c993f3557332d0df0f4cd6f0b7c323821edc4dd5

Localization/Translations/Moonwell.lua
```diff
@@ -0,0 +1,33 @@
+---@type l10n
+local l10n = QuestieLoader:ImportModule("l10n")
+
+local moonwellLocales = {
+    ["Moonwell"] = {
```
BreakBB, line 5: Remember to ask the translators for the other ones here

anon1231823:
```
    ["Moonwell"] = {
        ["enUS"] = true,
        ["deDE"] = "Mondbrunnen",
        ["esES"] = "Poza de la Luna",
        ["esMX"] = "Poza de la Luna",
        ["frFR"] = "Puits de lune",
        ["koKR"] = "달샘",
        ["ptBR"] = "Poço Lunar",
        ["ruRU"] = "Лунный колодец",
        ["zhCN"] = "月亮井",
        ["zhTW"] = "月井",
    },
```

Localization/Translations/Moonwell.lua
Sommos: Fixed in: https://github.com/Questie/Questie/pull/7213/commits/0041032c21299b59908d8dbe6f91f894aacf8932

redcatthethird:
OMG, thank you so much! This is great!

PR #7226, opened 2025-10-22
Update interface version to 11508

Gogo1951:
Update for 1.15.8 game client.


BreakBB:
Ah, thanks for the reminder 👍🏻

PR #7230, opened 2025-10-26
[fix] Dungeon translation in tooltips & UI frame check

Sommos:
## Proposed changes

- FIX: Dungeon translation in tooltips & quest links to accurately translate the dungeon name
- FIX: UI base frame being checked for existance before generating additional layer frames

### DIFF
--- a/Database/Zones/zoneDB.lua
+++ b/Database/Zones/zoneDB.lua
@@ -151,6 +151,7 @@ function ZoneDB:GetLocalizedDungeonName(areaId)
     else
         local alternativeDungeonAreaId = alternativeDungeonAreaIdToDungeonAreaId[areaId]
         if alternativeDungeonAreaId then
+            areaId = alternativeDungeonAreaId
             dungeonName = dungeons[alternativeDungeonAreaId][1]
         end
     end


Modules/Journey/QuestieJourney.lua
```diff
@@ -142,8 +142,13 @@ function QuestieJourney:IsShown()
     return isWindowShown
 end
 
+-- There are ways to toggle this function before the frame has been created
 function QuestieJourney:ToggleJourneyWindow()
-    -- There are ways to toggle this function before the frame has been created
+    if not QuestieJourneyFrame then
```
BreakBB: I suggest you make use of `Questie.started` and then simply do the `print` and return instead of calling `BuildMainFrame`. The Questie init process is quite complex and `BuildMainFrame` is only a small portion of `QuestieJourney:Initialize` which is also only a small portion of the entire Questie init.

That way it _should_ also be possible to remove the `if QuestieJourneyFrame then` below to only have a single `print`.

Modules/QuestLinks/Link.lua
```diff
@@ -204,8 +204,9 @@ _AddDungeonInfo = function(quest)
     if zoneOrSort and zoneOrSort > 0 then
         local dungeonName = ZoneDB:GetDungeonName(zoneOrSort)
         if dungeonName then
+            local localizedDungeonName = C_Map.GetAreaInfo(zoneOrSort) or dungeonName
```
BreakBB: I suggest to move this into `ZoneDB:GetDungeonName` and rename that function to `ZoneDB:GetLocalizedDungeonName`. We only use that function here for the map icon tooltip and in the link.

Modules/QuestLinks/Link.lua
```diff
@@ -348,11 +349,11 @@ _AddPlayerQuestProgress = function(quest, starterName, starterZoneName, finisher
         else
             if finisherName then
                 _AddTooltipLine(" ")
-                _AddTooltipLine((l10n("Ended by")..": " .. Questie:Colorize(finisherName, "gray")))
+                _AddTooltipLine((l10n("Ended by")..": " .. Questie:Colorize(l10n(finisherName), "gray")))
```
BreakBB: This is a dead call because we don't load quest, NPC, object and item translations into `l10n` but they get baked into the DB itself at compile time. So they will always return what you put int (the default behavior of `l10n`).

Modules/QuestLinks/Link.lua
```diff
@@ -378,10 +379,10 @@ _AddPlayerQuestProgress = function(quest, starterName, starterZoneName, finisher
         else
             if starterName then
                 _AddTooltipLine(" ")
-                _AddTooltipLine((l10n("Started by")..": " .. Questie:Colorize(starterName, "gray")))
+                _AddTooltipLine((l10n("Started by")..": " .. Questie:Colorize(l10n(starterName), "gray")))
```
BreakBB: Same as above, NPC/object names are not in `l10n`.

Sommos: 18efd3d

Modules/QuestLinks/Link.lua
Sommos: 18efd3d

Modules/QuestLinks/Link.lua
Sommos: 18efd3d

Modules/Journey/QuestieJourney.lua
Sommos: 18efd3d

PR #4743, opened 2023-03-30
[feature] Add Quests by Faction tab

BreakBB:
## Issue references

Fixes #3344

## Proposed changes

- Adds a new section "Quests by Faction" tab to the Journey

Big props to @scleal who developed the base functionality for this in #3419

## Screenshots
grafik

PR #7240, opened 2025-11-15
Implemented a Quests by Faction tab in the journey

IAmChills:
## Issue references
3 years ago I asked for a new feature to be able to filter quests by reputation rewards. It was played with but never added. I have since learned LUA and hope to see this merged.

This is a new PR based on the outdated quests-by-faction PR from BreakBB (thank you)
https://github.com/Questie/Questie/pull/4743 

## Proposed changes

This is a complete rewrite/update based on the latest build of Questie to introduce a new Tab within the My Journey window. This will add a Quests by Faction tab that allows the user to select their expansion and reputation to view a list of quests that reward reputation with the chosen faction. This tab automatically filters out quests that are not available to your race, class, or faction.

I have also introduced some new details around the reputation reward value to be included in the descriptions and titles.

## Screenshots

image

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -2822,6 +2822,11 @@ globals = {
     "WOW_PROJECT_CATACLYSM_CLASSIC",
     "WOW_PROJECT_MISTS_CLASSIC",
     "WOW_PROJECT_MAINLINE",
+    "EXPANSION_NAME0",
+    "EXPANSION_NAME1",
+    "EXPANSION_NAME2",
+    "EXPANSION_NAME3",
+    "EXPANSION_NAME4",
     "FACTION_STANDING_CHANGED",
     -----------------------------------------------------> External Addons
     "AtlasToggleFromWorldMap",


Modules/Journey/QuestieJourney.lua
```diff
@@ -112,6 +113,10 @@ function QuestieJourney:BuildMainFrame()
                 text = l10n("Quests by Zone"),
                 value="zone"
             },
+            {
+                text = l10n("Quests by Faction"),
```
BreakBB: This string also needs to be added to the translation files. See `Localization/Translations/Journey`

Modules/Journey/tabs/QuestsByFaction/QuestsByFactions.lua
```diff
@@ -0,0 +1,684 @@
...
+local factionQuestMap
+
+local factionIDs = QuestieDB.factionIDs
+
+local expansionDefinitions = {
+    { key = "classic", label = l10n("Classic"), order = Expansions.Era },
+    { key = "tbc", label = l10n("Burning Crusade"), order = Expansions.Tbc },
+    { key = "wotlk", label = l10n("Wrath of the Lich King"), order = Expansions.Wotlk },
+    { key = "cata", label = l10n("Cataclysm"), order = Expansions.Cata },
+    { key = "mop", label = l10n("Mists of Pandaria"), order = Expansions.MoP },
```
BreakBB: All these expansions also needs translations. Feel free to add a new file for it to `Localization/Translations`. Remember to add it to `Localization/Translations/Translations.xml`.

In general I think it's worth going through the changes and find all new usages of `l10n` and see if the translations for it already exist or if a placeholder entry is required.

Modules/Journey/tabs/QuestsByFaction/QuestsByFactions.lua
```diff
@@ -0,0 +1,684 @@
...
+}
+
+local expansionKeyByOrder = {}
+local expansionOrderByKey = {}
+for _, expansion in ipairs(expansionDefinitions) do
+    expansionKeyByOrder[expansion.order] = expansion.key
+    expansionOrderByKey[expansion.key] = expansion.order
+end
+
+local expansionFactionCandidates = {
```
BreakBB: I think it would be good to move this table (and maybe the loop below) into a new file.

Modules/Journey/tabs/QuestsByFaction/QuestsByFactions.lua
```diff
@@ -0,0 +1,684 @@
...
+            ) or {}
+
+            local requiredRaces = queryResult[10]
+            local requiredClasses = queryResult[11]
+
+            if _IsQuestAvailableToPlayer(requiredRaces, requiredClasses) then
+                local questName = QuestieLib:GetColoredQuestName(questId, Questie.db.profile.enableTooltipsQuestLevel, false)
+                local rewardValue = _GetFactionReputationRewardValue(questId, factionId)
+                if rewardValue and rewardValue ~= 0 then
+                    questName = questName .. " (" .. rewardValue .. ")"
```
BreakBB: I think it would be nice to show the rep reward this similar to the map icons:

Image

So with a `+` and in the same blue tone. I'd skip the icon though.

Modules/Journey/tabs/QuestsByFaction/QuestsByFactions.lua
```diff
@@ -0,0 +1,684 @@
...
+    end)
+
+    container:AddChild(factionTreeFrame)
+end
+
+---Build the quest tree for a faction grouping
+---@param factionId number
+---@return table<number, any> | nil
+function _QuestieJourney.questsByFaction:CollectFactionQuests(factionId)
+    _EnsureFactionQuestData()
```
BreakBB: Why is this done here and not in `QuestieJourney:Initialize` where `zoneMap` is generated?

Modules/Journey/tabs/QuestsByFaction/QuestsByFactions.lua
```diff
@@ -0,0 +1,684 @@
...
+                    "nextQuestInChain",
+                    "parentQuest",
+                    "preQuestSingle",
+                    "preQuestGroup",
+                    "requiredMinRep",
+                    "requiredMaxRep",
+                    "requiredSpell",
+                    "requiredMaxLevel",
+                    "requiredRaces",
+                    "requiredClasses"
```
BreakBB: I think `_EnsureFactionQuestData` should handle these two fields and simply don't add quests for the opposite faction to `QuestieJourney.factionMap`.

BreakBB (commented):
This looks pretty great already! Thank you so much for taking the time to tackle this feature again 🥳 

There are a few things I added notes to. Most importantly we need to make sure that every string handed to `l10n` has at least a place holder entry in our translation files, so that our translators can go through them. Please don't use AI to add non-trivial translations.

Modules/Journey/tabs/QuestsByFaction/QuestsByFactions.lua
anon1231823: There are global strings for expansion names so these don't need manual translation it should use the global strings

https://wago.tools/db2/GlobalStrings?filter%5BBaseTag%5D=EXPANSION_NAME&page=1

IAmChills: Thanks for the the feedback, this was adjusted. I also see there are a few global strings available for some of the tree:

AVAILABLE_QUESTS = "Available Quests"
TRACKER_FILTER_COMPLETED_QUESTS = "Completed Quests"

But I maintained consistency with the QuestsByZone file.

Modules/Journey/tabs/QuestsByFaction/QuestsByFactions.lua
IAmChills: This was moved!

Modules/Journey/tabs/QuestsByFaction/QuestsByFactions.lua
IAmChills: I am not 100% sure how you would like this to look but I did utilize the rewardBlue coloring and ended up providing the full string in parenthesis?

image

Modules/Journey/tabs/QuestsByFaction/QuestsByFactions.lua
IAmChills: My thought process was to lazy load it on request but I have moved it to initialize.

Modules/Journey/tabs/QuestsByFaction/QuestsByFactions.lua
IAmChills: This was moved to the mapping!

IAmChills:
@BreakBB Thank you for all the great feedback. I hope the adjustments I made are progress. I have never tackled something to this scale before, its a big AddOn! I appreciate all your help.

Modules/Journey/tabs/QuestsByFaction/QuestsByFactions.lua
anon1231823: The wording should be consistent for what exists in 'Quest by Zone' as far as "Select your Faction" etc

Does having expansion even make sense? And if I'm playing classic, I shouldn't see TBC reputations. 

I don't know if you did this already, but when using global strings I know those have to be declared somewhere

IAmChills: That's exactly why I made an expansion drop down, so you can only interact with the quests available to you. It uses Expansions.lua module to determine your `WOW_PROJECT_ID` and only shows the available expansions in the list. Classic will never see TBC +. It seemed cleaner to me to be able to focus down on specific expansions so the reputation list isn't a mile long if you are looking for something in MoP for example.

IAmChills: I did end up removing the "Your" part of the strings, so it says "Select Expansion and Faction"

IAmChills:
Added function to identify if the player is tracking a reputation and set to the default selection if so.

Cabro:
First time running the PR:

1 - I don't see anywhere "select expansion" being used. I just have the dropdown with the expansions. Neat. Clean.
image

2 - Missing "Quests By Faction" locale strings
image

3 - Either the string translation is missing or this one has an extra space at the end
image

4 - It populates every faction, even if it has 0 quests. Probably it should not show repuations from the opposite faction.
Obviously, a character faction change should still be detected and only show the reputations for your active faction. Could not test this case.
image

5 - I'm not sure how the tracked rep show as default should work when you have more expansions available. Should it show it by default as soon as you open MJ's tab for Quests by Faction? Or should it show only after you select the expansion where the faction is added?

IAmChills:
@Cabro 

1. It is just a default, habits. I removed it under the assumption there will never be an error fetching expansions.
2. Added missing locale.
3. Added missing locale.
4. Fixed this to now filter out factions that do not apply to the player. I had only filtered out the quests previously.
5. I feel it should immediately be the default expansion and faction if you are tracking it. It is only persistent between sessions so your saved state remains on close. I feel this is a quality of life function. If you are tracking it, chances are you want to work on that reputation.

Cabro:
I still have issues with 4.
image

It populates for example Ravasaur Trainers/Sunreaver Onslaught/Forest Hozen etc as alliance, and i keep seeing factions with 0 quests. Ravasaur Trainers and Shado-Pan Assault for example.

5. Fully agree with chose behavior!

6. Also i think Shang Xi's Academy should be Pandaren only, right?

Cabro:
I'm not entirely sure what "Horde" tab should show, since it's either missing a ton of quests, or something is broken. Notice this is for `Classic` expansion.
image

I assume "Alliance" one is the same

Modules/Journey/tabs/QuestsByFaction/QuestsByFactions.lua
BreakBB: This looks great to me 👌🏻 

What do you think @anon1231823 @Cabro ?

Modules/Journey/tabs/QuestsByFaction/QuestsByFactions.lua
BreakBB: I see 🤔 

Maybe that's something worth doing for both tabs after this feature is merged.

Cabro: Is this the cause for sections with 0/0/0/0/0?

IAmChills:
@Cabro

So I think I fixed some of the issues you were seeing with factions unavailable to you but I am am also potentially seeing an issue... A second set of eyes would be great. Unfortunately I don't play MoP so it's not as easy for me to track this down. It's working in Classic!

Some quests that should be restricted to Race/Class have a nil or 0 value in the table? Examples:

`[32508] = "Welcome to the Island"` has a `0` in the requiredRaces argument when it should be `ALL_HORDE = 33555378`
`[29404] = "Much to Learn"` has a `0` in the requiredRaces argument when it should be `PANDAREN = 8388608`
I am not sure why Ravasaur Trainers even show up, is that even an actual faction? I didn't map it at all...

I wonder if this is the discrepancy?

Additionally, the Classic Horde / Alliance subsections, to my understanding are just mapped to increase your reputation with any classic major cities. It is part of the quests table and just shows you any quest that is mapped to a major city.

Cabro:
29404 is actually blacklisted and shouldn't show up in any table.
`[29404] = Expansions.Current >= Expansions.MoP, -- Not in the game` - in QuestieQuestBlacklist.lua

32508 is something that should've been in Phase3 in Questie\Database\Corrections\ContentPhases\MistsOfPandaria.lua
We took most of those from whatever private DBs we could find, or what info wowhead had at the time (which obviously is improved nowadays). Since we are not yet in that phase, we have no way of checking if that is a quest that the players can actually take.

Ravasaur Trainers is a faction added in Cataclysm, the equivalent for Wintersaber Trainers. The whole faction has 1 NPC which starts 3 quests that give no reputation for the faction. I am just as clueless as you are :D https://www.wowhead.com/npc=11701/morvek

For the Horde/Alliance tab, shouldn't it only show quests that actually give reputation then? We have some corrections with `factionIDs.HORDE` but for MoP only atm. The quest reward reputation values haven't actually been the center of our focus until very recently, so whatever values in DB we have actually are from the private DBs, since Blizzard does not disclose these numbers. As you can probably imagine, most of it is absolutely wrong.
Furthermore, even wowhead cannot make up its mind what these values should actually be. The Tillers NPCs quests have different reputation rewards across the expansions (per wowhead), and I have a quest that is alliance only, with a horde reputation reward for all expansions - https://www.wowhead.com/quest=29803/ears-of-the-lich-king

Modules/Journey/tabs/QuestsByFaction/QuestsByFactions.lua
```diff
@@ -0,0 +1,540 @@
...
+
+    for questId in pairs(QuestieDB.QuestPointers) do
+        local queryResult = QuestieDB.QueryQuest(questId, queryFields) or {}
+        local requiredMinRep = queryResult[1]
+        local requiredMaxRep = queryResult[2]
+        local reputationReward = queryResult[3]
+        local requiredRaces = queryResult[4]
+        local requiredClasses = queryResult[5]
+
+        if _IsQuestAvailableToPlayer(requiredRaces, requiredClasses) then
```
BreakBB, line 281: To (fully?) fix the "Factions without quests are showing up" issue, we should move the filtering of hidden quests to this place.

So the check that is currently done further below when iterating `sortedQuestByLevel` needs to happen here and is then no longer required below.

Database/Corrections/ContentPhases/MistsOfPandaria.lua
```diff
@@ -294,6 +294,7 @@ local questsToBlacklistByPhase = {
         [32503] = true, -- Desecrate this Fire!
         [32506] = true, -- A Wing to Fly On
         [32507] = true, -- Skin of the Saurok
+        [32508] = true, -- Welcome to the Island
```
BreakBB, line 297: Looks like @Cabro pushed some stuff to the wrong branch 😂

BreakBB:
> I'm not entirely sure what "Horde" tab should show, since it's either missing a ton of quests, or something is broken. Notice this is for Classic expansion.

I am also unsure what the best approach towards this is. I mean we already have the single factions in the list (e.g. "Undercity").

Modules/Journey/tabs/QuestsByFaction/QuestsByFactions.lua
IAmChills: I moved the filtering to this section.

IAmChills:
> > I'm not entirely sure what "Horde" tab should show, since it's either missing a ton of quests, or something is broken. Notice this is for Classic expansion.
> 
> I am also unsure what the best approach towards this is. I mean we already have the single factions in the list (e.g. "Undercity").

I just committed a few of the suggestions and I also added a check to only populate the "Horde" and "Alliance" tabs with quests that actually provide reputation. I personally kind of like this option as it allows you to sort of just go wild on the overall faction if that is your goal. Instead of cross-referencing each major city and running all around the world you can sort of narrow in on a zone and know you are taking care of all major factions, which is a very reasonable grind. At the end of the day you can still just single out specific city if you want to as well.

Modules/Journey/tabs/QuestsByFaction/QuestsByFactionsTab.lua
```diff
@@ -0,0 +1,194 @@
...
+    if not selectedExpansionKey then
+        selectedExpansionKey = _GetDefaultExpansion()
+    end
+    local factions = selectedExpansionKey and QuestieJourney.factionsByExpansion[selectedExpansionKey]
+
+    if _QuestieJourney.lastFactionSelection[2] then
+        currentFactionId = _QuestieJourney.lastFactionSelection[2]
+    else
+        currentFactionId = RESET
+    end
```
BreakBB:
```suggestion
    if not selectedExpansionKey then
        selectedExpansionKey = _GetDefaultExpansion()
    end
    local factions = selectedExpansionKey and QuestieJourney.factionsByExpansion[selectedExpansionKey]

    local currentFactionId = RESET
    if _QuestieJourney.lastFactionSelection[2] then
        currentFactionId = _QuestieJourney.lastFactionSelection[2]
    end
```

BreakBB:
Alright so @Cabro also agreed this looks great 🥳 

Now to make the CI happy as well:

1. `EXPANSION_NAME0` and so need to be added to the [.luacheckrc](https://github.com/Questie/Questie/blob/master/.luacheckrc#L40) globals
2. I left another comment, to fix the other problem about unused value on the `currentFactionId` variable

IAmChills:
Awesome! I am not completely familiar with commenting on PR's, I can't seem to figure out how to post code snippets or reply to your comments but let me know if there is something else I need to do!

anon1231823:
> Awesome! I am not completely familiar with commenting on PR's, I can't seem to figure out how to post code snippets or reply to your comments but let me know if there is something else I need to do!

The EXPANSION global variables need to be added to luacheck

IAmChills:
> > Awesome! I am not completely familiar with commenting on PR's, I can't seem to figure out how to post code snippets or reply to your comments but let me know if there is something else I need to do!
> 
> The EXPANSION global variables need to be added to luacheck

Oh I see. Does it really matter where? I put it within the Blizzard Data below the Project variables.

anon1231823:
> > > Awesome! I am not completely familiar with commenting on PR's, I can't seem to figure out how to post code snippets or reply to your comments but let me know if there is something else I need to do!
> 
> > 
> 
> > The EXPANSION global variables need to be added to luacheck
> 
> 
> 
> Oh I see. Does it really matter where? I put it within the Blizzard Data below the Project variables.

@BreakBB

BreakBB:
Let's see what the pipeline says 😁 

To me your changes look good @IAmChills 👌🏻

BreakBB:
Looks fine as well.

Thank you so much @IAmChills for taking all the time to finally finish QuestsByFaction. This surely is an awesome addition to Questie 🥳 

If you ever want to tackle something else, your PRs are highly appreciated. Also feel free to hit us up on Discord if you want 👍🏻

PR #7250, opened 2025-11-21
Update API calls for TBC Anniversary 2.5.5 compatibility

NDRWhun:
- Updated interface version to 20505 in Questie-BCC.toc
- Updated chat APIs to use C_ChatInfo namespace (AceComm-3.0)
- Replaced ChatFrame_OnHyperlinkShow with ChatFrameMixin:OnHyperlinkClick
- Added ChatFrameUtil.AddMessageEventFilter compatibility with safe wrappers
- Added nil-safe WOW_PROJECT_ID comparisons
- Added nil checks for ContentPhases and tracker line operations
- Fixed UTF-8 encoding in build.py for TOC file reading
- TBC Anniversary uses same database as regular TBC (WOW_PROJECT_ID = 5)

All changes verified against WoW UI source 2.5.5

## Issue references

Fixes #

## Proposed changes

-
-

## Screenshots
{04086281-12D6-4CDE-8869-FB0796179011}
{B8793105-169D-43C0-8BB1-5823D50B23B1}
{A35A7300-99C4-4BD8-BFB5-B97EF87B7384}


Modules/QuestieInit.lua
```diff
@@ -171,6 +171,7 @@ QuestieInit.Stages[1] = function() -- run as a coroutine
     end
 
     -- Check if the DB needs to be recompiled
+    -- Note: TBC Anniversary (2.5.5) uses WOW_PROJECT_ID = 5 (same as regular TBC), so it uses the same database
```
BreakBB, line 174:
```suggestion
```

Database/Corrections/QuestieCorrections.lua
```diff
@@ -253,6 +253,7 @@ function QuestieCorrections:Initialize(validationTables)
     _LoadCorrections("itemData", QuestieItemFixes:Load(), QuestieDB.itemKeysReversed, validationTables)
     _LoadCorrections("objectData", QuestieObjectFixes:Load(), QuestieDB.objectKeysReversed, validationTables)
 
+    -- TBC corrections apply to both regular TBC and TBC Anniversary (2.5.5) since they use the same WOW_PROJECT_ID (5)
```
BreakBB, line 256:
```suggestion
```

Libs/HereBeDragons/HereBeDragons-2.0.lua
```diff
@@ -16,11 +16,11 @@ HereBeDragons.worldMapData     = HereBeDragons.worldMapData or {}
...
-local WoWClassic = (WOW_PROJECT_ID == WOW_PROJECT_CLASSIC)
-local WoWBC = (WOW_PROJECT_ID == WOW_PROJECT_BURNING_CRUSADE_CLASSIC)
-local WoWWrath = (WOW_PROJECT_ID == WOW_PROJECT_WRATH_CLASSIC)
-local WoWCata = (WOW_PROJECT_ID == WOW_PROJECT_CATACLYSM_CLASSIC)
-local WoWMists = (WOW_PROJECT_ID == WOW_PROJECT_MISTS_CLASSIC)
+local WoWClassic = ((WOW_PROJECT_ID or 0) == (WOW_PROJECT_CLASSIC or 2))
+local WoWBC = ((WOW_PROJECT_ID or 0) == (WOW_PROJECT_BURNING_CRUSADE_CLASSIC or 5))
+local WoWWrath = ((WOW_PROJECT_ID or 0) == (WOW_PROJECT_WRATH_CLASSIC or 11))
+local WoWCata = ((WOW_PROJECT_ID or 0) == (WOW_PROJECT_CATACLYSM_CLASSIC or 14))
+local WoWMists = ((WOW_PROJECT_ID or 0) == (WOW_PROJECT_MISTS_CLASSIC or 19))
```
BreakBB, line 23: Why was this added? 🤔

Database/Corrections/QuestieQuestBlacklist.lua
```diff
@@ -6960,7 +6960,9 @@ function QuestieQuestBlacklist:Load()
     if Questie.IsSoD then
         Questie:Debug(Questie.DEBUG_DEVELOP, "Blacklisting quests for SoD...")
         questsToBlacklist = ContentPhases.BlacklistSoDQuestsByPhase(questsToBlacklist, ContentPhases.activePhases.SoD)
-    elseif Questie.IsAnniversary or Questie.IsAnniversaryHardcore then
+    elseif (Questie.IsAnniversary or Questie.IsAnniversaryHardcore) and ContentPhases.activePhases and ContentPhases.activePhases.Anniversary and ContentPhases.BlacklistAnniversaryQuestsByPhase then
+        -- Only apply Anniversary blacklist for Classic Anniversary, not TBC Anniversary
+        -- TBC Anniversary uses the same database as regular TBC and doesn't need phase-based blacklisting
```
BreakBB, line 6965: Yeah, we'll need to add proper content phases for TBC so quests will unlock correctly. And I agree this needs to be disabled for Anniversary TBC. However the change here shouldn't be necessary once Blizzard fixes the `WOW_PROJECT_ID` to be updated for TBC and isn't `2`.

Modules/QuestLinks/Link.lua
```diff
@@ -389,22 +389,51 @@ _AddPlayerQuestProgress = function(quest, starterName, starterZoneName, finisher
...
+                    end
                 end
             end
         end
     end
-end)
+end
+
+-- Try new API first (2.5.5+)
+if ChatFrameMixin and ChatFrameMixin.OnHyperlinkClick then
```
BreakBB, line 416: Great solution to keep compatibility for non 2.5.5 clients 👌🏻

Modules/Tracker/QuestieTracker.lua
```diff
@@ -1879,14 +1879,24 @@ function QuestieTracker:UpdateHeight()
...
+        local currentLine = TrackerLinePool.GetCurrentLine()
+        local firstLine = TrackerLinePool.GetFirstLine()
 
-        if TrackerLinePool.GetCurrentLine().mode == "zone" then
-            -- If a single zone is the only line in the tracker then don't add pixel padding
-            trackerQuestFrame.ScrollChildFrame:SetHeight((TrackerLinePool.GetFirstLine():GetTop() - TrackerLinePool.GetCurrentLine():GetBottom()))
-        else
-            -- Add 3 pixels to bottom of tracker to account for text that traverses beyond the GetStringHeight() function such as lower case "g".
-            trackerQuestFrame.ScrollChildFrame:SetHeight((TrackerLinePool.GetFirstLine():GetTop() - TrackerLinePool.GetCurrentLine():GetBottom() + 3))
+        if currentLine and firstLine then
```
BreakBB, line 1885: Sweet cleanup. The Tracker code desperately needs it 😅 

This `nil` check shouldn't be required though, is it?

Database/Corrections/ContentPhases/Anniversary.lua
```diff
@@ -775,9 +775,14 @@ local questsToBlacklistByPhase = {
...
+    if not questsToBlacklistByPhase or not contentPhase then
+        return questsToBlacklist
+    end
     for phase = contentPhase + 1, #questsToBlacklistByPhase do
-        for questId in pairs(questsToBlacklistByPhase[phase]) do
-            questsToBlacklist[questId] = true
+        if questsToBlacklistByPhase[phase] then
+            for questId in pairs(questsToBlacklistByPhase[phase]) do
+                questsToBlacklist[questId] = true
+            end
```
BreakBB, line 785: With a proper solution/work around for `Questie.IsTBC` this isn't needed.

In general we plan to do some renaming so that the current "Anniversary" will become "Era" and it can be better distinguished to "TBC" (there won't be "Anniversary Era" one TBC goes live, will it?). I don't think there will be "Anniversary TBC" naming wise as there is no competitive TBC flavor.

Database/Corrections/Holidays/QuestieEvent.lua
```diff
@@ -141,7 +141,7 @@ function QuestieEvent:Load()
     end
 
     -- TODO: Also handle WotLK which has a different starting schedule
-    if Questie.IsClassic and (((not Questie.IsAnniversary) and (not Questie.IsAnniversaryHardcore)) or (ContentPhases.activePhases.Anniversary >= 3)) then
+    if Questie.IsClassic and (((not Questie.IsAnniversary) and (not Questie.IsAnniversaryHardcore)) or (ContentPhases and ContentPhases.activePhases and ContentPhases.activePhases.Anniversary and ContentPhases.activePhases.Anniversary >= 3)) then
```
BreakBB, line 144: Same as with the other comment above: If we have proper `Questie.IsTBC` we won't need this change.

Modules/QuestieShutUp.lua
```diff
@@ -3,6 +3,44 @@ local QuestieShutUp = QuestieLoader:CreateModule("QuestieShutUp")
 ---@type l10n
 local l10n = QuestieLoader:ImportModule("l10n")
 
+-- Compatibility: 2.5.5+ uses ChatFrameUtil.AddMessageEventFilter/RemoveMessageEventFilter instead of ChatFrame_AddMessageEventFilter/RemoveMessageEventFilter
+local ChatFrameAddMessageEventFilter = ChatFrameUtil and ChatFrameUtil.AddMessageEventFilter or ChatFrame_AddMessageEventFilter
+local ChatFrameRemoveMessageEventFilter = ChatFrameUtil and ChatFrameUtil.RemoveMessageEventFilter or ChatFrame_RemoveMessageEventFilter
```
BreakBB, line 8: Good catch that we also need this change here!

I am thinking about moving this to the [QuestieCompat](https://github.com/Questie/Questie/blob/master/Modules/QuestieCompat.lua) module, so we have it at a single place. Otherwise I can promise I forget about this place if I change something at the other...

Modules/VersionCheck.lua
```diff
@@ -48,23 +48,24 @@ Questie._sessionWarnings = {}
...
 ---@type boolean
-Questie.IsWotlk = WOW_PROJECT_ID == WOW_PROJECT_WRATH_CLASSIC
+Questie.IsWotlk = (WOW_PROJECT_ID or 0) == (WOW_PROJECT_WRATH_CLASSIC or 11)
 
---- Addon is running on Classic TBC client
+--- Addon is running on Classic TBC client (includes TBC Anniversary 2.5.5+)
+--- Note: TBC Anniversary uses the same WOW_PROJECT_ID (5) as regular TBC, so it uses the same database
 ---@type boolean
-Questie.IsTBC = WOW_PROJECT_ID == WOW_PROJECT_BURNING_CRUSADE_CLASSIC
+Questie.IsTBC = (WOW_PROJECT_ID or 0) == (WOW_PROJECT_BURNING_CRUSADE_CLASSIC or 5)
```
BreakBB, line 64: I suggest we temporarily use `GetBuildInfo` to determine TBC build and then use `Questie.IsTBC` to disable `Questie.IsAnniversary`. What do you think @NDRWhun ?

Libs/HereBeDragons/HereBeDragons-2.0.lua
NDRWhun: It was a nil check for debugging/safety. Can be removed. I have cleaned most of the nil checks. (I've added them where I saw a lua error, to find out what's causing it.

Database/Corrections/QuestieQuestBlacklist.lua
NDRWhun: Sure but for PTR it's necessary to "bypass". Later can be reworked ^^

Modules/Tracker/QuestieTracker.lua
NDRWhun: Can be removed. it's unnecessary now.

Database/Corrections/ContentPhases/Anniversary.lua
NDRWhun: Good point, I was just wanted to separate old things from the new one. But was good for comparisons.

Modules/VersionCheck.lua
NDRWhun: Good idea can be done

PR #7265, opened 2025-12-07
Add Questie.API namespace

BreakBB:
## Issue references

Fixes #

## Proposed changes

- Add `Questie.API` as official namespace for other addons to interact with Questie
  - `Questie.API.isReady` needs to be `true` before the API returns correct data
  - `Questie.API.RegisterOnReady` can be used to register a callback which is fired once `Questie.API.isReady` is true
  - Add `Questie.API.GetQuestObjectiveIconForUnit`
  - Add `Questie.API.RegisterForQuestUpdates`

### DIFF
--- a/Modules/QuestieNameplate.lua
+++ b/Modules/QuestieNameplate.lua
@@ -74,23 +74,21 @@ function QuestieNameplate:UpdateNameplate()
 
     for guid, token in pairs(activeGUIDs) do
         local unitName, _ = UnitName(token)
-        if (not unitName) then
-            return
-        end
+        if unitName then
+            local icon = QuestieNameplate.GetIcon(guid)
 
-        local icon = QuestieNameplate.GetIcon(guid)
-
-        if icon then
-            local frame = _QuestieNameplate.GetFrame(guid)
-            -- check if the texture needs to be changed
-            if frame.lastIcon ~= icon then
-                frame.lastIcon = icon
-                frame.Icon:SetTexture(icon)
+            if icon then
+                local frame = _QuestieNameplate.GetFrame(guid)
+                -- check if the texture needs to be changed
+                if frame.lastIcon ~= icon then
+                    frame.lastIcon = icon
+                    frame.Icon:SetTexture(icon)
+                end
+            else
+                -- tooltip removed but we still have the frame active, remove it
+                activeGUIDs[guid] = nil
+                _QuestieNameplate.RemoveFrame(guid)
             end
-        else
-            -- tooltip removed but we still have the frame active, remove it
-            activeGUIDs[guid] = nil
-            _QuestieNameplate.RemoveFrame(guid)
         end
     end
 end


Logonz:
A general comment, should we keep the API in the `Modules/` Folder or should we do something to allow more in the future.

Imagine something like "External/", "Public" or just "API/" (Or any name)
Then we can split the files into maybe "Events" and/or "Quest" - I think you understand what you are getting at 👊 

"Public/PublicEnum.lua"
```
QuestieAPI.Enums = {
    ---@enum QuestUpdateTriggerReason
    QuestUpdateTriggerReason = {
        QUEST_ACCEPTED = 1,
        QUEST_UPDATED = 2,
        QUEST_TURNED_IN = 3,
        QUEST_ABANDONED = 4,
    }
}
```

"Public/API.lua"
RegisterOnReady
PropagateOnReady

"Public/PublicQuest.lua"
GetQuestObjectiveIconForUnit
RegisterForQuestUpdates
PropagateQuestUpdate

Something like that maybe, don't stare yourself blind at the same suggestions :D

---

I wonder if we should run can call the callbacks in coroutines and only do one call per "frame" in case the user of our API use a ton of performance in one frame for some reason but this is just a thought. We might also need to do more security through stuff like setfenv to block the addons from extracting internal Questie data that we don't want it to do, as they could in theory just call things from our namespace using this.

Modules/API.lua
```diff
@@ -0,0 +1,92 @@
...
+    for _, callback in pairs(questUpdateCallbacks) do
+        if next(objectiveIndices) then
+            for _, objectiveIndex in pairs(objectiveIndices) do
+                xpcall(function() callback(questId, objectiveIndex, triggerReason) end, CallErrorHandler)
+            end
+        else
+            xpcall(function() callback(questId, nil, triggerReason) end, CallErrorHandler)
+        end
+    end
+end
```
Logonz: It is a bit confusing here for me initially that one is a public function and one is a private function.

Questie.API is obviously public and Questie. is not. and then we have QuestieAPI.

I think we should make it clearer how each is used and if i was to implement a new event it should be obvious how to implement it.

BreakBB: Yeah, good thought. I just placed all that stuff into a single file for simplicity.

Modules/EventHandler/QuestEventHandler.test.lua
```diff
@@ -164,7 +171,12 @@ describe("QuestEventHandler", function()
 
         QuestEventHandler.QuestRemoved(QUEST_ID)
 
-        _G.C_Timer = {NewTicker = function(_, callback) callback() return {} end}
+        _G.C_Timer = {
+            NewTicker = function(_, callback)
+                callback()
+                return {}
+            end
+        }
```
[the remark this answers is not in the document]
BreakBB: This is okay, as only a partial Ticker is used.

Modules/QuestieNameplate.lua
```diff
@@ -78,15 +73,12 @@ function QuestieNameplate:UpdateNameplate()
...
 
     for guid, token in pairs(activeGUIDs) do
-
         local unitName, _ = UnitName(token)
-        local _, _, _, _, _, npcId, _ = strsplit("-", guid)
-
-        if (not unitName) or (not npcId) then
+        if (not unitName) then
             return
         end
```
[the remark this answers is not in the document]
BreakBB: Good catch.

Public/Enums.lua
```diff
@@ -0,0 +1,13 @@
...
+QuestieAPI.Enums = {
+    ---@enum QuestUpdateTriggerReason
+    QuestUpdateTriggerReason = {
+        QUEST_ACCEPTED = 1,
+        QUEST_UPDATED = 2,
+        QUEST_TURNED_IN = 3,
+        QUEST_ABANDONED = 4,
+    }
+}
+Questie.API.Enums = QuestieAPI.Enums
```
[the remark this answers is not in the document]
BreakBB: The load order in the TOC files make sure `Questie.API` is always initialized before any API file is reached.

Public/RegisterForQuestUpdates.lua
```diff
@@ -0,0 +1,33 @@
+---@class QuestieAPI
+local QuestieAPI = QuestieLoader:ImportModule("QuestieAPI")
+
+---@type table<number, fun(questId: number, objectiveIndex: number?, triggerReason: QuestUpdateTriggerReason)>
```
Logonz: ---@type fun(questId: number, objectiveIndex: number?, triggerReason: QuestUpdateTriggerReason)[]

Public/GetQuestObjectiveIconForUnit.lua
```diff
@@ -0,0 +1,18 @@
+---@type QuestieNameplate
+local QuestieNameplate = QuestieLoader:ImportModule("QuestieNameplate")
+
+--- Returns the icon path for the quest objective marker for the given unit GUID.
+--- Only NPC GUIDs are valid inputs.
+---@param guid string
+---@return string | nil -- Path to the icon texture, or nil if no icon is available.
```
Logonz, line 7: Discussion of wanting to do string|nil or string? ;)
I doesn't actually matter so i think we can just mix them.

Public/RegisterOnReady.lua
```diff
@@ -0,0 +1,34 @@
+---@class QuestieAPI
+local QuestieAPI = QuestieLoader:ImportModule("QuestieAPI")
+
+---@type table<number, fun()>
```
Logonz: ---@type fun()[]

Public/GetQuestObjectiveIconForUnit.lua
BreakBB: I think for return types it's good to have them explicit, while for parameter's I prefer the `?` writing.

PR #7266, opened 2025-12-08
[feature] Add a toggle for fading quest lines in tracker

### LINKED ISSUE #7263: ACCESSIBILITY - Add option to disable fading out of tracker when hovering mouse cursor over it
### Description

(WOW Classic) 
Moving my mouse cursor over the tracker window causes all quests but the one the cursor is highlighting to fade out.  This makes it very difficult to read the other quests.  95% of the time I move my cursor over the tracker my intention is to find a certain quest (by using mouse scroll wheel to scroll through the list) and fading the text out defeats this purpose.

I've reviewed tracker options in the app and searched for an answer online but was unable to find an existing solution.

While I'm sure some people appreciate the transparency feature, it makes it very difficult for people with not-perfect vision to read anything.  Please add a toggle in the options to disable this behavior as an accessibility feature.

Sommos:
## Issue references

Fixes #7263

## Proposed changes

- Add a new toggle that will disable the hover fade in the tracker

## Screenshots

image


Modules/Options/QuestieOptionsDefaults.lua
```diff
@@ -84,6 +84,7 @@ function QuestieOptionsDefaults:Load()
             hideTrackerInPetBattles = true,
             trackerFadeMinMaxButtons = false,
             trackerFadeQuestItemButtons = false,
+            trackerDisableHoverFade = false,
```
BreakBB, line 87: We also need a migration for this.

Modules/Options/TrackerTab/QuestieOptionsTracker.lua
```diff
@@ -498,10 +498,23 @@ function QuestieOptions.tabs.tracker:Initialize()
...
                         type = "toggle",
                         order = 6,
                         width = 1.5,
+                        name = function() return l10n("Disable Quest Hover Fade") end,
+                        desc = function() return l10n("When this is checked, the other quests in the tracker will stay fully opaque while hovering a quest.") end,
+                        disabled = function() return not Questie.db.profile.trackerEnabled end,
+                        get = function() return Questie.db.profile.trackerDisableHoverFade end,
+                        set = function(_, value)
+                            Questie.db.profile.trackerDisableHoverFade = value
+                            TrackerLinePool.OnHighlightLeave()
```
BreakBB: Why are we calling this? When changing the setting the user should not be able to hover at the same time 🤔

Localization/Translations/Options/Tracker.lua
```diff
@@ -504,6 +504,31 @@ local trackerOptionsLocales = {
         ["zhTW"] = "啟用時，滑鼠沒有指向任務清單時，背景和 (如果有啟用) 邊框會變成透明。",
     },
     ---------------------------------------------------------
+    ["Disable Quest Hover Fade"] = {
+        ["enUS"] = true,
+        ["deDE"] = false,
```
BreakBB:
```suggestion
        ["deDE"] = "Verblassen bei Quest-Hover deaktivieren",
```

Localization/Translations/Options/Tracker.lua
```diff
@@ -504,6 +504,31 @@ local trackerOptionsLocales = {
...
+        ["frFR"] = "Désactiver l’estompage au survol des quêtes",
+        ["koKR"] = false,
+        ["ptBR"] = "Desativar desvanecimento ao passar o cursor",
+        ["ruRU"] = "Отключить затухание по курсору",
+        ["zhCN"] = false,
+        ["zhTW"] = false,
+    },
+    ["When this is checked, the other quests in the tracker will stay fully opaque while hovering a quest."] = {
+        ["enUS"] = true,
+        ["deDE"] = false,
```
BreakBB:
```suggestion
        ["deDE"] = "Wenn diese Option aktiviert ist, wird der Hover-Effekt von anderen Quests deaktiviert und sie verblassen nicht.",
```

Sommos: 9de725d

Localization/Translations/Options/Tracker.lua
Sommos: 9de725d

Modules/Options/TrackerTab/QuestieOptionsTracker.lua
Sommos: You're right - It's not something we need technically, but if we were to provide a key binding for this setting as has been suggested for other options, we will need to use this to reset the pool back to an alpha value of 1 to keep visual consistency with the setting.

7409621

Modules/Options/QuestieOptionsDefaults.lua
Sommos: 7409621

Issue #7293, opened 2025-12-19
"Unique Map Icon Colors" should be on by default


Gogo1951:
### Description

Think of Zangarmar Marsh... ha. This should be on by default going into TBC. 

Seriously consider making the colors more vibrant and distinct too.

<img width="837" height="909" alt="Image" src="https://github.com/user-attachments/assets/5f00aa0c-598a-46e4-bb47-21e4ca5248a8" />

Gogo1951:
Same for the minimap option... and can we confirm these use the same colors? The current colors make it hard for me to tell what's what... but I think the minimap is using random colors and the map is using another set of random colors. (=

<img width="734" height="179" alt="Image" src="https://github.com/user-attachments/assets/9923eaf6-9406-4404-a509-bc642ff8ffc5" />

PR #7296, opened 2025-12-19
"Unique Map Icon Colors" should be on by default

Gogo1951:
Fix for https://github.com/Questie/Questie/issues/7293

This does a lot to help with zones like Ungoro and Zangarmarsh where the icons are all on top of each other.


Gogo1951:
Night and day difference. (=

You guys had the features, just have to enable them by default so more users will see them. Ha. Keep in mind ONLY devs ever touch the settings. No other users will ever go into these.

image

vs. old.

image

BreakBB:
The icon coloring makes the icon type pretty hard to actually identify, because the colors are a bit heavy. From a UX perspective clear to identify icons are better than somewhat _colorish_ icon-dots.

Therefore the map icon glow is active by default. Users still have a way to identify different icon types, but also have some colors to distinguish quests. So I am open to activate the minimap glow, also to be consistent with the world map.

That being said, to properly change options, we need a new entry to the [migrationFunctions](https://github.com/Questie/Questie/blob/af951c48d34cb6e1461414d42a0b54a0858afa86/Modules/Migration.lua#L10) table.

Gogo1951:
Hey @BreakBB thanks for breaking it down. So just add something like this to [migrationFunctions](https://github.com/Questie/Questie/blob/af951c48d34cb6e1461414d42a0b54a0858afa86/Modules/Migration.lua)? 

```lua
[20] = function()
        if Questie.db.global.alwaysGlowMinimap == nil then
            Questie.db.global.alwaysGlowMinimap = true
        end

        if Questie.db.global.questObjectiveColors == nil then
            Questie.db.global.questObjectiveColors = true
        end

        if Questie.db.global.questMinimapObjectiveColors == nil then
            Questie.db.global.questMinimapObjectiveColors = true
        end
    end,
```

BreakBB:
> Hey @BreakBB thanks for breaking it down. So just add something like this to [migrationFunctions](https://github.com/Questie/Questie/blob/af951c48d34cb6e1461414d42a0b54a0858afa86/Modules/Migration.lua)?

Actually you don't need the `nil` check. Just override the value, as we want to force it to the new default value. Only for `alwaysGlowMinimap` though. We want to keep the others disabled.

Modules/Migration.lua
```diff
@@ -126,7 +126,33 @@ local migrationFunctions = {
...
+        end
+
+        -- Preserve previous dungeon hide preference for both new flags
+        local previousHideInDungeons = Questie.db.profile.hideTrackerInDungeons
+
+        Questie.db.profile.minimizeTrackerInCombat = false
+        Questie.db.profile.minimizeTrackerInDungeons = previousHideInDungeons
+        Questie.db.profile.hideTrackerInCombat = false
+        Questie.db.profile.hideTrackerInDungeons = false
+    end,
```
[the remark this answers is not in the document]
BreakBB: You only inspected this part, because of merge conflicts.

Modules/Options/QuestieOptionsDefaults.lua
```diff
@@ -46,9 +46,9 @@ function QuestieOptionsDefaults:Load()
             nameplateTargetFrameY = 25,
             nameplateTargetFrameScale = 1.7,
             alwaysGlowMap = true,
-            alwaysGlowMinimap = false,
-            questObjectiveColors = false,
-            questMinimapObjectiveColors = false,
+            alwaysGlowMinimap = true,
+            questObjectiveColors = true,
```
BreakBB:
```suggestion
            questObjectiveColors = false,
```

Modules/Options/QuestieOptionsDefaults.lua
```diff
@@ -46,9 +46,9 @@ function QuestieOptionsDefaults:Load()
...
+            questMinimapObjectiveColors = true,
```
BreakBB:
```suggestion
            questMinimapObjectiveColors = false,
```

Modules/Migration.lua
```diff
@@ -126,7 +126,25 @@ local migrationFunctions = {
...
+    [20] = function()
+        Questie.db.global.alwaysGlowMinimap = true
+    end,
```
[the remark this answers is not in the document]
BreakBB: You are correct. This needs to be on `profile` level, no `global`.

Modules/Migration.lua
```diff
@@ -141,7 +141,10 @@ local migrationFunctions = {
         Questie.db.profile.minimizeTrackerInDungeons = previousHideInDungeons
         Questie.db.profile.hideTrackerInCombat = false
         Questie.db.profile.hideTrackerInDungeons = false
-    end
+    end,
+    [20] = function()
+        Questie.db.global.alwaysGlowMinimap = true
```
BreakBB:
```suggestion
        Questie.db.profile.alwaysGlowMinimap = true
```

PR #7300, opened 2025-12-20
Slight Icon Glow Color Refresh

### LINKED ISSUE #7299: Update Colors for Icon Glow
### Description

There are a few colors that tend to render a bit dark.

Image

Suggest a slight refresh of the color pallet.

Image

Gogo1951:
Some of the colors were too dark and too similar to one another. 

Pulled from https://tailwindcolor.com/... 18 colors now instead of 24, but less overlap -- that's by design. Odds of having all 24 colors on one map are low, and it was causing them to muddy too much.

## Issue references

Fixes #7299

## Screenshots

Updated Colors

image


Gogo1951:
image

vs

image

Gogo1951:
And I know you guys will do everything you can do to avoid taking any advice from me, but seriously... just ask your favorite AI for impartial feedback. 

You're devs, that's a good thing. You aren't designers or people who have a lot of background in UX... so just outsource that. (=

## Prompt

> You're a World of Warcraft player trying to complete quests...
> Which of these designs better helps you on your way? 

## Response...

As a WoW player just trying to get my leveling done in Teldrassil, the second design (Image 2 - the tan/brown icons) is significantly better.

Here is the breakdown from a player's perspective:
1. Visual Clarity & Contrast

    Image 1 (Dark Icons): The black/dark grey icons get lost against the darker textures of the map (like the trees and shadows). They look like visual "voids" or graphical glitches rather than interactive objects. It forces me to squint to see where the actual markers are.

    Image 2 (Light Icons): The tan/leather color mimics standard WoW UI elements (like the loot cursor or inventory icons). They "pop" against the green terrain of Teldrassil much better. I can instantly scan the map and see exactly where the density of objectives is.

2. Readability of the "Glow"

    Image 1: Because the center is pitch black, the colored outlines (pink, neon green) look a bit harsh and disconnected.

    Image 2: The colored glows (cyan, red, green) blend naturally with the lighter icon. It feels more like a rarity border or a quest objective highlight, which is intuitive to me as a player.

3. Iconography

    Image 2 clearly communicates "Lootable Object" or "Item Provision." The texture on the sack is visible.

    Image 1 silhouettes the shape, making it harder to identify what the object actually is at a quick glance.

Verdict: Go with Image 2. When I'm rushing to find quest items or specific consumables, I need the map to be readable in a split second. The lighter icons accomplish that; the dark ones just add visual noise.

BreakBB:
I don't think it is a good idea to reduce the available colors from 24 to just 8. Once you either have more than 8 quests in the same zone or more than 8 different objectives in the same zone, the colors are not unique anymore.

I do agree though, that the colors can be improved greatly and taking some UI colors (tailwind, material UI, you name it) sounds like a great approach 👍🏻 

About your screenshots and the AI feedback you posted: You are comparing apples and oranges. I am pretty sure you have the "Unique Map Icon Colors" active in the first screenshot, which you have not in the second. And since that is a non-default option, only power users (as you call them), will ever experience this horrible sight with all these black and brownish icons 😛

Gogo1951:
It's 16, not 8. And while you may have more quests, you really don't have a lot of colors that don't overlap at that point.

BreakBB:
> It's 16, not 8.

True, my sleepy brain oversaw that you changed the formatting (which make sense, now that I notice it). 16 is far better and worth a try 👍🏻

BreakBB (approved):

Issue #7301, opened 2025-12-20
Update Default Icon Grouping from 50 to 180


Gogo1951:
### Description

Defaulting to ~~120~~ **180** spacing is a massive win for map readability because it prioritizes "signals over noise". By forcing icons into groups, it lowers the cognitive load on your brain, turning a messy sea of symbols into clear, actionable clusters that you can parse in a split second. It keeps the map looking clean and modern while making sure you can actually see the terrain you’re trying to navigate.

The main problem with 30 spacing is that it creates a "clutter effect" where having more data actually makes the map less useful. When icons are packed that tightly, they overlap and blend together, making it nearly impossible to pick out specific quest types or priorities at a glance. For most players, more icons just means burying the important info under a pile of visual distractions.

## Settings

30

<img width="351" height="75" alt="Image" src="https://github.com/user-attachments/assets/97dbd712-df0c-4f80-a5ae-5f0ff815c729" />

120

<img width="356" height="78" alt="Image" src="https://github.com/user-attachments/assets/353e9a93-2eb4-4e85-99ec-efd9d7c49a15" />

## Normal Map 

30

<img width="832" height="623" alt="Image" src="https://github.com/user-attachments/assets/8362eb0f-ad5f-49b5-8c5f-8e729d457260" />

120

<img width="837" height="628" alt="Image" src="https://github.com/user-attachments/assets/33b18e1f-7b67-4c22-aed8-66e88adb180f" />

## Full Screen Map

This one is meh...

30

<img width="1453" height="1079" alt="Image" src="https://github.com/user-attachments/assets/83dd2505-b97e-486f-8931-1b9fe8d383a5" />

120

<img width="1460" height="1080" alt="Image" src="https://github.com/user-attachments/assets/2e5a7ef7-be4e-4e48-8ee5-4dd6ecf38de0" />

## World Map

30

<img width="1450" height="1080" alt="Image" src="https://github.com/user-attachments/assets/9c73ce68-f888-4ea7-8800-84f352ba84e1" />

120

<img width="1450" height="1080" alt="Image" src="https://github.com/user-attachments/assets/053c7a03-952c-45b9-9b53-402f0992c1c2" />

BreakBB:
This really depends on the quest and what one want to achieve. More **clustering** makes the data much less accurate, because of how the icons are grouped together. Especially on quests where only a few marker exist and/or a lot of people are trying to do the same quest, more clustering is not helpful.

[...]

PR #7303, opened 2025-12-22
Change icon cluster from 50 to 180

Gogo1951:
## Issue references

https://github.com/Questie/Questie/issues/7301

## Proposed changes

- Reduce "Icon Vomit" in order to improve how useful Questie is to players.

## Screenshots

Going from:

image

To this:

image

### DIFF
--- a/Modules/Migration.lua
+++ b/Modules/Migration.lua
@@ -145,6 +145,11 @@ local migrationFunctions = {
     [20] = function()
         Questie.db.profile.alwaysGlowMinimap = true
     end,
+    [21] = function()
+        if Questie.db.profile.clusterLevelHotzone == 50 then
+            Questie.db.profile.clusterLevelHotzone = 180
+        end
+    end,
 }
 
 function Migration:Migrate()


Gogo1951:
FYI the screenshots have https://github.com/Questie/Questie/pull/7300 applied as well. Slight refresh to the Color Wheel to avoid some "muddy" colors and overlapping colors.

Gogo1951:
Hey @BreakBB & @Cabro,

Regarding the subtle UI cleanup -- I think we need to look at this through the lens of the Passive User vs. the Power User.

* The Complaint: If you look at Reddit and Discord, users consistently say the screen is too cluttered to find information quickly. Clutter kills confidence. Everyone wants to play confidently, and they hate having that flow interrupted.

* The Reality: 95% of users never touch settings. They live and die by our defaults. Giving them a guide on how to change defaults isn't the answer; it's like handing a customer a cookbook when they walk into a restaurant. Settings get in the way of the game.

* The Solution: We optimize defaults for the 95%. The "Power Users" (the 5%) will customize their settings regardless, so changing the defaults won't hurt them.

Candidly, I love questing. My favorite thing in the game is to 2-3 man group and fly through zones as fast as possible. Questing is crazy fun, and the better the information flow, the faster and more fun it becomes.

I want Questie to be the reason people stop boosting and start leveling. Even if they are removing boosting, I want people to not even care because they see Questing as accessible and fun. If we make the default UX just a little cleaner, we make it just a little easier for users to have confidence.

Questie has the power to positively influence how the game is played. Consider how Warcraft Logs shaped raiding behavior by deciding that World Buffs count for parses. Imagine a world where World Buffs disqualified your parse? Imagine how different people would play! The tool shapes the behavior.

We can guide people to the fun. The game doesn't have to stick to strict #NoChanges standards; add-ons should strive to modernize and bring UX standards that didn't exist 20 years ago. I want to help the average player find joy in questing rather than feeling bewildered by too many dots on the map.

That's what these changes are about. (=

BreakBB:
We won't change these values just like that, as explained multiple times.

However we plan a representative poll about your "passive user" vs. "power user" claims, to see what most of the users expect Questie to behave like.

But that is nothing that will happen this year, as holidays are right there and the time we have is currently invested in more important topics.

Gogo1951:
You have TBC coming out soon. It's a good time to do a refresh.

Gogo1951:
Hey @BreakBB ,

I wanted to share some context on why polling is often unreliable for understanding passive users. This is my area of expertise, running product and design teams.

Polling suffers from inherent selection bias; it captures the "Vocal Minority" (power users) but misses the passive majority.

A better alternative is "Hallway Testing" -- grabbing random users (perhaps via 1:1 DMs on Discord) to get raw, unsolicited feedback. This helps you see how the average person -- or at least the average person who already took the effort to join the Questie Discord (which is already pretty heavily biased) -- interacts with the app, rather than just what the power users say they want.

Regarding the icon spacing: Even in a poll, the new spacing (Option A vs B) will win simply because it is objectively better UX/usability.

Gogo1951:
Aero and I were trying really hard to fix a lot of these UX changes back when TBC Classic rolled around the first time, as you recall. It would be amazing to see 'Routes' become a priority again!

Routes made Questie behave more like Quest Helper -- the add-on we all used back when the game was real. They give such a massive confidence booster for users. Instead of just showing them where the quests are, it gives them a suggested plan of attack. Even if the path isn't perfect, just taking that cognitive load off their plate is a huge win.

For anyone who wasn't around back then, I believe the code is still sitting in the Routes branch. It was in beta (and definitely still needs work), but it was a really solid approach to solving the 'icon vomit' complaints that we wanted to fix even 5 years back.

https://github.com/user-attachments/assets/c0e77558-04b1-4c3f-a967-45f1c4c5c0e2

https://github.com/user-attachments/assets/b5ab8a6d-935b-4856-95df-f82524148632

BreakBB:
Just to give an update on this: We haven't decided if/how we want to change the default clustering.

Also: Routes will not be implemented at any time from the current team. That will cause it to be more of a "guide addon" and remove it's simplicity.

BreakBB:
We had multiple reports of users wondering, why they see so few icons while they had clustering set to 50. We decided against fiddling with the defaults for this at this point.

Gogo1951:
"Multiple users told us what we wanted to hear."

Cool.

It's still wrong and a shitty decision that makes questing harder for users.

Wish you guys were capable of growth, but clearly you still just want to build model train sets in your basement without feedback from others.

PR #7307, opened 2025-12-25
Update Questie-BCC.toc

Gogo1951:
Just a reminder, Category fields are now supported in TBC.

Pasted_Image_12_25_25__12_28 AM

image

### DIFF
--- a/Questie-BCC.toc
+++ b/Questie-BCC.toc
@@ -1,5 +1,5 @@
 ## Interface: 20505
-## Title: Questie|cFF00FF00 v11.15.1|r
+## Title: Questie
 ## Author: Aero/Logon/Muehe/TheCrux(BreakBB)/Drejjmit/Dyaxler/Cheeq/TechnoHunter/Everyone else
 ## Notes: A standalone Classic QuestHelper
 ## Notes-esES: Un asistente de misiones


Gogo1951 (commented):
Image

By default the version is shown in the tooltip now. No need to include it along with the title.

Gogo1951:
Hey @BreakBB,

These are simple changes, but I don't know if you want them just in BCC or across the board. I think across the board would be better.

BreakBB:
[This is the most official list](https://warcraft.wiki.gg/wiki/Addon_Categories#Quests) there is. So it should be "Quests" instead of "Quests & Leveling". We can also take all the translations from that list.

> I think across the board would be better.

Yeah, for every TOC.

> By default the version is shown in the tooltip now. No need to include it along with the title.

That is just true for this new UI and [the release script](https://github.com/Questie/Questie/blob/master/release.py#L26) will override it. So let's just keep it for now.

anon1231823:
Fixed by 6ddd925

Gogo1951:
Hey @BreakBB ,

I do think you're better off going with the CurseForge categories... nobody is going to find the ones on WoW Wiki, and that hardly seems "official" but time will tell. (=

Pasted_Image_1_15_26__11_44 AM

PR #7351, opened 2026-01-19
Implement Item Drop Rate Tooltips

### LINKED ISSUE #2333: Implement full item database with loot table info
This would be useful to have both for the drop rates of quest items but also for journey search

https://github.com/Muehe/sqlua/commit/837895bd845b4dd6afa5ae6b38a72744e59011ff

@Muehe recently implemented it into his db tool, so we just need to add code to work with it in questie

Yttrium-tYcLief:
## What does this PR add?
This PR adds item droprates to unit tooltips for item objectives and requiredSourceItems.

image

These can be toggled with a new option in Questie Options > General tab > Tooltip Options > "Show item drop rates", default is on.

## How does it work?
There's a new module called DropDB, mostly independent of QuestieDB. It is not compiled, simply static. Drop rate data is stored in static tables.

Rates do not dictate anything about quest eligibility or where icons show up. They also exist wholly outside of the main item DB. We only show them on tooltips if we're already planning to show the item on an NPC tooltip, in which case we check if there's drop data for that item+NPC combo and display it if so.

These only show on direct unit tooltips, not on map icon tooltips, since those often combine multiple NPCs into one tooltip so we wouldn't know which rate to show.

## Where does the data come from?
There are three 'levels' of data - corrections, private server, and Wowhead.

We show data based on which table has a higher priority, if it contains an item:npc match. Corrections is the highest priority, followed by private server data, followed last by Wowhead.

Corrections are manual; we shouldn't waste too much time on these, but if we notice something egregious, the corrections table allows us to correct it manually. Corrections propagate to later expansions (unless themselves overwritten in later expansion corrections), just like normal QuestieDB corrections.

Private server data is exported by querying those databases for all quest-only item drop rates; essentially, if the private server database believes an item should only drop while on a quest, then we care about that item's drop data and it's included in our data. This helps ensure 100% drop rates for guaranteed items (boss heads, etc), while also giving mostly accurate rates for most other drops.

For Era, TBC, and Wrath, we use cmangos data. For Cata, we use mangos3. MoP subscribes to the Cata mangos3 export because MoP databases are a bit of a mess; our own MoP baseDB is stitched together from three separate sources as-is.

Last but not least is Wowhead data. Wowhead data is scraped directly from their website for each expansion (reflecting drop rate changes across expansions) and is shown whenever we both don't have a correction for that item:npc, and whenever the private server data believes that item doesn't require being on the quest to drop. A good example of this is Westfall Stew - the ingredients for that quest's objectives drop regardless of being on the quest, so Wowhead's crowdsourced drop data is probably fairly accurate compared to whatever might be in a private server database.

Having Wowhead support is also important for future releases (a la Season of Discovery) where there's simply no equivalent private server data source to reference.

In order to obtain a list of IDs to scrape Wowhead for, a new QuestieSlash command was added: `/questie itemdrop`. This command queries the currently loaded and compiled database and prints out every single item ID found in item objectives and requireSourceItem tables for every quest in the DB.

This PR also includes an overhaul to our scrapers to add expansion-level support to them; now you can specify `-ex tbc` etc in order to query a specific expansion's database. Defaults to classic era if unspecified. This is necessary as part of this PR because we need to scrape and maintain separate item drop DBs for each expansion, so the scraper needs to be able to differentiate.

## How is the data presented?
Drop rates are rounded with the following rules:

- If above 10%, show integers only.
- If below 10% but above 2%, show 1 decimal place.
- If below 2%, show 2 decimal places.
- If below 0.01%, show 3 decimal places. (This should almost never happen but, better than showing 0.00%)

Drop rates are only shown next to your own objectives, for now.

If Questie Debug is enabled, then an icon is also shown to inform developers of the data source being used.

## Misc
Fixes #2333


ExternalScripts(DONOTINCLUDEINRELEASE)/scraper/item_drop/item_drop_spider.py
```diff
@@ -0,0 +1,49 @@
...
+    def __init__(self) -> None:
+        super().__init__()
+        self.start_urls = [self.base_url_classic.format(item_id) for item_id in ITEM_IDS]
+
+    def parse(self, response):
+        result = {}
+        for script in response.xpath('//script/text()').extract():
+            result["itemId"] = response.url.split("/")[-2][5:]
+            if script.startswith('\nWH.Gatherer.addData'):
+                result["name"] = re.search(r'"name":"((?:[^"\\]|\\.)*)"', script).group(1)
```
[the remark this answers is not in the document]
Yttrium-tYcLief: While I agree that we should probably be using JSON rather than Regex for all of this, the new item drop spider is simply following the existing conventions from the other spiders. We'll notice if something goes drastically wrong. Redoing it in JSON is a good idea for the future but outside the scope of this PR.

ExternalScripts(DONOTINCLUDEINRELEASE)/scraper/item_drop/item_drop_spider.py
```diff
@@ -0,0 +1,49 @@
...
+            if script.startswith('\n    var tabsRelated ='):
+                list_views_pattern = re.compile(r'new Listview\((.*?)}\)', re.DOTALL)
+
+                for match in list_views_pattern.findall(script):
+                    list_view_name = re.search(r'id: \'(.*?)\',', match).group(1)
+                    if list_view_name == "dropped-by":
+                        #print(match)
+                        complex_pattern = re.compile(r'"id":(\d+).*?"modes":{.*?"0":{.*?"count":(\d+).*?"outof":(\d+)}}')
+                        for item in complex_pattern.findall(match):
+                            result[str(item[0])] = [int(item[1]),int(item[2])]
```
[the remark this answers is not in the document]
Yttrium-tYcLief: We'll notice if something goes drastically wrong. Redoing it in JSON is a good idea for the future but outside the scope of this PR.

Database/DropTables/dropDB.lua
```diff
@@ -0,0 +1,135 @@
...
+    -- 3. Copy that list of IDs into the file Questie/ExternalScripts(DONOTINCLUDEINRELEASE)/scraper/item_drop/item_ids.py
+    -- 4. Run:   python Questie/ExternalScripts(DONOTINCLUDEINRELEASE)/scraper/runner.py --item-drop -ex #      where # is the expansion level; 0 for classic, 1 for tbc, etc. also accepts strings
+    -- 5. Once it's done scraping, the data will be in Questie/ExternalScripts(DONOTINCLUDEINRELEASE)/scraper/item_drop/item_drop_data.lua
+
+    -- To extract cmangos/mangos3 data:
+    -- 1. Host the relevant MySQL database locally on your machine
+    -- 2. Tweak Questie/ExternalScripts(DONOTINCLUDEINRELEASE)/scraper/item_drop/cmangos_itemdrops.py for your host IP, user, password, and database ID, if necessary
+    -- 3. Run:   python cmangos_itemdrops.py
+    -- 4. Once it's done extracting, the data will be in Questie/ExternalScripts(DONOTINCLUDEINRELEASE)/scraper/item_drop/item_drop_data.lua
+    -- (we don't need a list of IDs like for wowhead because we rely on cmangos' own data for what items are quest-only drops)
```
BreakBB: Might be better to move this to a scraper README and leave a note here where to find it.

Database/QuestieDB.lua
```diff
@@ -364,6 +366,17 @@ function QuestieDB:GetItem(itemId)
     return item
 end
 
+---@param itemId ItemId
+---@param npcId NpcId
+---@return number?
+function QuestieDB.GetItemDroprate(itemId, npcId)
+     if not DropDB or not DropDB.GetItemDroprate then
+         Questie:Debug("ItemDrops: DropDB not available")
+         return nil
```
BreakBB, line 375: Would it be better to return `0` as default?

ExternalScripts(DONOTINCLUDEINRELEASE)/scraper/item_drop/item_drop_data.lua
BreakBB, line 1: Do we need this data to be checked in?

ExternalScripts(DONOTINCLUDEINRELEASE)/scraper/item_drop/item_drop_spider.py
```diff
@@ -0,0 +1,65 @@
...
+            case 2:
+                base_url = "https://www.wowhead.com/tbc/item={}"
+            case 3:
+                base_url = "https://www.wowhead.com/wotlk/item={}"
+            case 4:
+                base_url = "https://www.wowhead.com/cata/item={}"
+            case 5:
+                base_url = "https://www.wowhead.com/mop-classic/item={}"
+            case _: # If number is unknown, treat it as classic
+                base_url = "https://www.wowhead.com/classic/item={}"
```
BreakBB, line 30: I think it's worth extracting this into a function only returning the first part and then simply add `/item={}`, `/object={}` etc. Not really important though and could be done after merging the PR.

Modules/Options/QuestieOptionsDefaults.lua
```diff
@@ -61,6 +61,7 @@ function QuestieOptionsDefaults:Load()
             enableTooltipsQuestLevel = true,
             showQuestXpAtMaxLevel = true,
             enableTooltipsNextInChain = true,
+            enableTooltipDroprates = true,
```
BreakBB, line 64: We need a migration step for this default to actually be applied.

Modules/Tooltips/Tooltip.lua
```diff
@@ -133,6 +133,7 @@ local function _FetchTooltipsForGroupMembers(key, tooltipData)
     if QuestieComms and QuestieComms.data:KeyExists(key) then
         ---@tooltipData @tooltipData[questId][playerName][objectiveIndex].text
         local tooltipDataExternal = QuestieComms.data:GetTooltip(key);
+        --DevTools_Dump(QuestieComms.data)
```
BreakBB: Remember to remove these debug logs

Database/QuestieDB.lua
```diff
@@ -364,6 +366,17 @@ function QuestieDB:GetItem(itemId)
     return item
 end
 
+---@param itemId ItemId
+---@param npcId NpcId
+---@return number?
```
BreakBB:
```suggestion
---@return table<number, string>?
```

Modules/Tooltips/Tooltip.lua
```diff
@@ -295,17 +300,45 @@ function QuestieTooltips.GetTooltip(key, playerZone)
...
+                                dropRateText = string.format("%.0f", dropRateData[1])
+                            elseif dropRateData[1] >= 2 then
+                                dropRateText = string.format("%.1f", dropRateData[1])
+                            elseif dropRateData[1] >= 0.01 then
+                                dropRateText = string.format("%.2f", dropRateData[1])
+                            else
+                                dropRateText = string.format("%.3f", dropRateData[1])
+                            end
+                            dropRateText = "  |cFF999999" .. dropIcon .. "[" .. dropRateText .. "%]|r";
+                        end
```
BreakBB: I'd prefer to have this extracted to a new function that simply returns the `dropRateText`.

Modules/Tooltips/Tooltip.lua
```diff
@@ -322,7 +355,7 @@ function QuestieTooltips.GetTooltip(key, playerZone)
     for questId, questData in pairs(tooltipData) do
         local hasObjective = false
         local tempObjectives = {}
-        for _, playerList in pairs(questData.objectivesText or {}) do
+        for x, playerList in pairs(questData.objectivesText or {}) do
```
BreakBB:
```suggestion
        for _, playerList in pairs(questData.objectivesText or {}) do
```

Modules/QuestieSlash.lua
```diff
@@ -263,6 +263,74 @@ function QuestieSlash.HandleCommands(input)
         return
     end
 
+    if mainCommand == "itemdrop" then
```
BreakBB: Maybe add a check to only make it available when debug is active?

Modules/QuestieSlash.lua
```diff
@@ -263,6 +263,74 @@ function QuestieSlash.HandleCommands(input)
...
+                if table[i] == val then
+                    return true
+                end
+            end
+            return false
+        end
+
+        local data = {}
+        for i=1,35000 do -- Manually search every quest ID from 1-35000 (roughly the upper bound for MoP;
+            -- if you're reading this and know if there's a way to just 'check every quest' cleanly feel free to rewrite)
```
BreakBB, line 273: As I said on Discord, the [validators](https://github.com/Questie/Questie/blob/master/cli/validate-mop.lua#L226) compile the DBs in the CI / CLI and you can then run all kind of operations on the data 😅 I mean this works and shouldn't be needed very often, so it shouldn't matter 🤷🏻

Questie-Mists.toc
```diff
@@ -137,6 +137,12 @@ Database\Corrections\mopNPCFixes.lua
 Database\Corrections\mopItemFixes.lua
 Database\Corrections\mopObjectFixes.lua
 
+# Item Drops
+Database\DropTables\dropDB.lua
+Database\DropTables\data\itemDropCorrections.lua
+Database\DropTables\data\cataItemDrops.lua
```
BreakBB: This is not required anymore, is it?

Database/DropTables/data/cataItemDrops.lua
BreakBB, line 1: I think it would be good to wrap the item drop tables in a multi line string and use `loadstring` to transform it to a table. Big tables need to be fully processed while the user is still in the loading screen, strings are much cheaper.

Database/QuestieDB.lua
Yttrium-tYcLief: The problem is that 0 would imply a drop rate of 0% (suggesting low but still present), rather than a non-existent droprate. In our code, the tooltip would then choose to display this number, rather than no drop data at all.

Database/DropTables/dropDB.lua
Yttrium-tYcLief: Sure, can do.

ExternalScripts(DONOTINCLUDEINRELEASE)/scraper/item_drop/item_drop_data.lua
Yttrium-tYcLief: You mean adding it to .gitignore? That'd be fine with me if so. It's basically just a temp file anyway, just like the .json files.

ExternalScripts(DONOTINCLUDEINRELEASE)/scraper/item_drop/item_drop_spider.py
Yttrium-tYcLief: While that would work currently with their pattern, we've seen Wowhead change their URL structure in the past and I wanted something that, while a little verbose, might require less reworking to react to a change like that. If we want to go that route that's fine with me - but like you said, not necessary for this PR.

Modules/Options/QuestieOptionsDefaults.lua
Yttrium-tYcLief: Oops! Can't believe I forgot about that.

Modules/Tooltips/Tooltip.lua
Yttrium-tYcLief: Ack, that's embarrassing! Thought I had caught every one I left.

Database/QuestieDB.lua
Yttrium-tYcLief: Oops, that was refactored and I forgot to update it.

Database/QuestieDB.lua
BreakBB: Sound reasonable. Just wanted to ask 👍🏻

ExternalScripts(DONOTINCLUDEINRELEASE)/scraper/item_drop/item_drop_data.lua
BreakBB: Yeah, `.gitignore` these "intermission files" and remove them from the PR.

ExternalScripts(DONOTINCLUDEINRELEASE)/scraper/item_drop/item_drop_spider.py
BreakBB: If they change their URL structure that would _most likely_ affect all areas, so it would be even easier if we can react in a single point, instead of multiple. But yeah, leave it like it is for now :)

Modules/Tooltips/Tooltip.lua
Yttrium-tYcLief: oops, checked that variable when testing and forgot to revert, good catch

Modules/QuestieSlash.lua
Yttrium-tYcLief: sure thing

Modules/QuestieSlash.lua
Yttrium-tYcLief: This was also implemented before that discussion, and even then I still don't have a properly working lua environment on my end. The other side benefit is it's also a bit independent of the current DB implementation, thinking back to that period in SoD where we hotloaded corrections on top of the live DB (though I think you've since written in validation for that stuff).

Honestly this whole method was very hacky and the only reason I figured I should leave it in is because we'll need to run it again in the future. It's one of those moments where I wonder if writing a better, cleaner method is really worth the effort when we'll literally only run the function maybe once a year.

Questie-Mists.toc
Yttrium-tYcLief: Why would this not be required?

Modules/QuestieSlash.lua
BreakBB: Yeah, I agree. Time is spent better elsewhere 👍🏻

Questie-Mists.toc
BreakBB: We are loading `cata` for MoP and also `mop`? This is an leftover from your testing isn't it?

Yttrium-tYcLief: Ah, nope, Cheeq noticed this one too. This is intentional - because right now we don't have a MoP pserver item drop DB, we load the Cata one instead when in MoP. This leaves most MoP-specific quest items with only Wowhead data, which isn't ideal, but it's better than nothing.

Tackling a MoP pserver DB source is going to be a massive headache, as we've already seen with the full base DB struggle.

BreakBB: aaaaah! Add a comment to that line then, to make it clear. It's just out of the pattern of the rest 😅

Yttrium-tYcLief: Can add one to the TOC, already have one in the DropDB.lua where it's loaded:

https://github.com/Questie/Questie/blob/90a0ced624aeaee5132372da345a43b996f48874/Database/DropTables/dropDB.lua#L47-L51

Issue #7222, opened 2025-10-19
Add TBC & WotLK moonwell locations


Sommos:
### Description

Follow up to #7213 

- TBC & WotLK moonwell locations should be added
- Prune moonwell locations to have a single ID 
- Prune moonwell locations in Vanilla zones for Cataclysm onwards

Alessandro-Barbieri:
Moonwell at 68 60.2 in Moonglade is active in classic too, I can provide a screenshot.

PR #7358, opened 2026-01-20
[feature] Add TBC moonwells & cleanup Classic moonwell locations

Sommos:
## Issue references
Linked to #7222

## Proposed changes
- Add TBC moonwell locations & expansion check
- Clean up Classic moonwell locations

## Screenshots
image


Modules/Map/Moonwell.lua
```diff
@@ -5,38 +5,58 @@ local Moonwell = QuestieLoader:CreateModule("Moonwell")
...
+    [177278] = true,
+    [177279] = true,
+    [177280] = true,
+    [177281] = true,
 }
 
+Moonwell.dataTBC = {
+    [400021] = true, -- Cenarion Thicket moonwell
+    [400022] = true, -- Evergrove moonwell
+}
```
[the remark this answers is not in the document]
Sommos: 2910665

anon1231823:
Any reason these have to have a specific name? As opposed to just using the generic moonwell like everything else is using.

Yttrium-tYcLief:
IMO we should be following the same structure as Mailboxes.lua when it comes to handling expansions, which is to say, a static list of objects for each expansion rather than an additive list.

I also do not think there's any need to have the moonwell locations localized. The only purpose for cataloging moonwells in the firstplace is to show icons on the map for tailors, so simply localizing "Moonwell" should be enough. Especially as some locations use shared IDs.

anon1231823:
> IMO we should be following the same structure as Mailboxes.lua when it comes to handling expansions, which is to say, a static list of objects for each expansion rather than an additive list.
> 
> 
> 
> I also do not think there's any need to have the moonwell locations localized. The only purpose for cataloging moonwells in the firstplace is to show icons on the map for tailors, so simply localizing "Moonwell" should be enough. Especially as some locations use shared IDs.


Yes, moonwell is already translated. I was trying to avoid additional translations.

Database/Corrections/classicObjectFixes.lua
```diff
@@ -359,6 +359,13 @@ function QuestieObjectFixes:Load()
             [objectKeys.spawns] = {[zoneIDs.STRANGLETHORN_VALE]={{13.5,15.7},{19.5,20.9},{22.2,19.8},{24.4,18.8},{24.5,65.7},{24.6,18.9},{25,58.3},{25.3,45.2},{25.4,49.3},{25.5,19.9},{25.7,66.9},{26.2,48.5},{26.3,67.8},{26.9,20.4},{27,39.7},{27.1,40.5},{27.3,68},{27.4,68.9},{27.7,21},{28.7,21.8},{29.2,23.2},{29.4,23.7},{29.4,24.6},{29.4,27.4},{29.4,27.7},{29.5,24.3},{29.5,25.1},{29.5,27.4},{29.5,27.7},{29.6,25.7},{31.9,32.1},{33,31.3},{33,32.1},{33.1,32.5},{34.2,32.4},{34.2,32.5},{34.2,34.7},{34.7,34.5},{34.8,33.7},{34.9,32.4},{35.1,35.6}}},
             [objectKeys.zoneID] = zoneIDs.STRANGLETHORN_VALE
         },
+        [400010] = {
+            [objectKeys.name] = "Moonglade Moonwell",
+            [objectKeys.questStarts] = {},
+            [objectKeys.questEnds] = {},
+            [objectKeys.spawns] = {[zoneIDs.MOONGLADE]={{68, 60.2}}},
```
BreakBB, line 366: So there are two moonwells in Moonglade? Because `177273` is also located in Moonglade 🤔

Sommos: Yes - There are actually 4 moonwells in Moonglade.

PR #7398, opened 2026-02-05
QoL for TBC Anniversary Dark Portal Launch

Yttrium-tYcLief:
This branch was intended to contain more stuff, but for now only has a few things, as of the time of this PR.

## Icons in Questgiver Tooltips
imageimageimageimageimage

These are colored based on the quest properties; in essence, the icon will match whatever color the text is. This is mainly to help differentiate quest starts/finishes from quest objectives.

## Questie Icon in Chat Messages
Before:
image
After:
image

This is achieved purely with a chat filter. We still send the original message in chat, but folks who actually have Questie installed will see that section replaced with just the Questie logo.

### DIFF
--- a/Modules/Tooltips/Tooltip.lua
+++ b/Modules/Tooltips/Tooltip.lua
@@ -303,7 +303,7 @@ function QuestieTooltips.GetTooltip(key, playerZone)
                             colorText = ":33:204:231"
                         else -- normal quest, use leveled colors
                             local r, g, b = QuestieLib:GetDifficultyColorPercent(level)
-                            colorText = ":" .. tostring(r * 255) .. ":" .. tostring(g * 255) .. ":" .. tostring(b * 255)
+                            colorText = ":" .. tostring(math.floor(r * 255)) .. ":" .. tostring(math.floor(g * 255)) .. ":" .. tostring(math.floor(b * 255))
                         end
                         if tooltip.type == "NPC" then
                             questString = "|TInterface\\Addons\\Questie\\Icons\\tooltip_available.png:14:14:0:0:32:32:0:32:0:32" .. colorText .. "|t" .. questString


Modules/Tooltips/Tooltip.lua
```diff
@@ -273,6 +277,32 @@ function QuestieTooltips.GetTooltip(key, playerZone)
             if tooltip.name then
                 if Questie.db.profile.showQuestsInNpcTooltip then
                     local questString = QuestieLib:GetColoredQuestName(questId, Questie.db.profile.enableTooltipsQuestLevel, true)
+                    if tooltip.type then
+                        local level, _ = QuestieLib.GetTbcLevel(questId)
+                        local repeatable = QuestieDB.IsRepeatable(questId)
+                        local event = QuestieEvent.IsEventQuest(questId)
+                        local pvp = QuestieDB.IsPvPQuest(questId)
```
BreakBB: We should inline these into the if-conditions, so they are not all run every time.

Modules/Tooltips/Tooltip.lua
```diff
@@ -71,6 +74,7 @@ function QuestieTooltips:RegisterQuestStartTooltip(questId, name, starterId, key
         questId = questId,
         name = name,
         starterId = starterId,
+        type = type or "NPC",
```
BreakBB: Can `type` actually be `nil`? The type does not reflect it. There is already `starterType or "NPC"` as default function call in `Modules/Quest/AvailableQuests/AvailableQuests.lua`.

Modules/QuestieAnnounce.lua
```diff
@@ -14,6 +14,48 @@ local GetItemInfo = C_Item.GetItemInfo or GetItemInfo
...
+            C_Timer.After(0.1, function()
+                SafeAddMessageEventFilter(event, filter)
+            end)
+        else
+            Questie:Error("Failed to register chat filter for", event, ":", err)
+        end
+    end
+end
+
+local pattern = (l10n:GetUILocale() == "ruRU" and "{звезда}" or "{rt1}").." Questie%s?:"
```
BreakBB: This does not work for `ruRU` because `l10n:GetUILocale()` will always return `enUS` on top level. Questie needs to be initialized for it to return the correct locale.

Why not setting the pattern in `QuestieAnnounce:InitializeLogoFilter()` as that is somewhat of the "init" for this code part?

And if you do that, we also don't need the `GetUILocale` mocks in the tests which failed earlier.

Modules/QuestieAnnounce.lua
BreakBB, line 1: Are we replacing "Questie: " with the icon or do we have icon + "Questie: " now? I am overlooking that somehow.

Modules/Tooltips/Tooltip.lua
Yttrium-tYcLief: god dude I was so tired when I wrote all this stuff I can't believe I left it as-is

Modules/Tooltips/Tooltip.lua
Yttrium-tYcLief: Not anymore, that was a holdover from an earlier revision where nil could be passed. It's defaulted earlier in the pipeline now.

Modules/QuestieAnnounce.lua
Yttrium-tYcLief: Is it bad that I just felt trauma when reading the string `l10n:GetUILocale()`?

Anyway... good idea, implementing.

Modules/QuestieAnnounce.lua
Yttrium-tYcLief: We are replacing the entire Questie string. This one is up for debate on what we should do, but I recall us discussing that and liking how it looked. It's much cleaner.

See screenshots in the PR description.

BreakBB: Yep, looks good as it is 👍🏻

PR #7412, opened 2026-02-11
Disallow Aldor/Scryer allegiance auto accept

erunehtar:
Disallow Aldor/Scryer allegiance auto accept. Even thought these quests were already in the disallowed turn-in section, it seems that wasn't enough to prevent them from being automated.

### DIFF
--- a/Modules/Auto/DisallowedIDs.lua
+++ b/Modules/Auto/DisallowedIDs.lua
@@ -183,6 +183,9 @@ _AutoQuesting.disallowedQuests = {
         [29205] = true, -- The Forlorn Spire
         [29206] = true, -- Into the Fire
         [31152] = true, -- Peering Into the Past
+        -- Aldor/Scryer quests
+        [10551] = true, -- Allegiance to the Aldor
+        [10552] = true, -- Allegiance to the Scryers
     },
     turnIn = {
         [2199] = true, -- Lore for a Price (crafting material objectives)


BreakBB:
Hey @erunehtar since they are already in the `turnIn` section, it is _highly_ unlikely Questie will turn them in.

Are you sure you don't have the auto turn in/accept feature active from any other addon? Like any questing guide addon or LeatrixPlus?

erunehtar:
> Are you sure you don't have the auto turn in/accept feature active from any other addon? Like any questing guide addon or LeatrixPlus?

No I do not use any other auto quest completion addons. I do not have any questing guide addons like RestedXP or any other, and I don't use LeatrixPlus.

Could it be these quests are already auto turn-in from the game itself? I mean, when you accept them, there's no second quest window where we click "Complete" ? Because if that was the case, then disallowing auto turn-in wouldn't do anything. In which case, I think its safer to assume we also need to disallow auto accepting them.

BreakBB:
Thanks for verifying that. Still strange to me, that Questie behaves differently on these quests than it does on others. Yeah, these quests are "turn-in only", so you don't accept them in the first place aka they are never in your quest log.

I'll talk with the team if we can reproduce this misbehavior on other quests, to be sure this is not a general problem.

Thanks a lot for bringing this to our attention and sorry for auto completing that quest in the first place, if that's on Questie.

erunehtar:
I guess its kind of expected that these quests get auto accepted, since the code for auto accept only check if its disallowed in the auto accept list:

https://github.com/Questie/Questie/blob/master/Modules/Auto/AutoQuesting.lua#L13-L15

imho, letting these two specific quests auto accept and rely on the disallow auto turn-in is too dangerous; the consequences are irreversible. Its not like an escort quest that can be restarted.

BreakBB:
Oh wait, they are not "turn-in only". You DO accept them before turning them in. Yeah, I agree we should also disallow auto acceptance.

BreakBB:
Would you mind still sharing a screenshot of your addon list ingame @erunehtar ?

erunehtar:
image
image
image
image
image

Issue #7405, opened 2026-02-07
[Refactoring] Standardize Wording & Introduce a Simple Glossary


NORPG:
### Description

**Background**

While working on translations, we noticed inconsistent wording in option-related strings (e.g. enabled / activated / checked).
These differences were not intentional, but came from multiple developers contributing without a shared ruleset.

Although these terms are mostly synonymous in English UI context, inconsistent usage:

- creates ambiguity for translators
- increases translation effort
- may imply semantic differences that don’t actually exist

**Proposal**

1. Standardize English wording
For options that represent the same toggle behavior, use one consistent phrasing (e.g. “When activated, …” or “If activated, …”).
2. Add a simple terminology / glossary reference
A small table documenting preferred wording for common UI concepts would help both developers and translators maintain consistency.
3. Encourage consistency within the same document/addon
Different wording should only be used when there is a meaningful difference in behavior.

**Non-Goals**

- This glossary applies _only_ to text displayed in the UI.
It does not affect developers’ writing style, terminology, or wording choices elsewhere (e.g. code comments, documentation, commit messages).
- This is _not_ an attempt to enforce strict wording or remove developer flexibility.
- This does _not_ require retroactively rewriting all existing strings.
- This does _not_ dictate how each language must translate the text; translators may still adapt wording naturally for their locale.
- This is _not_ about perfect literal translation, but about clarity and intent.

**Benefits**

- Clearer intent for translators
- More consistent localization across languages
- Less confusion and rework
- Cleaner UI text overall

NORPG:
Previous similar work.
Uniform term of Item/Object/Objective in zhCN/zhTW.  #6875

anon1231823:
I started some work on this as well

d338129 and 2fe628c and ad7bbfc and 69fa141 and 085ed39 and 78f8059

anon1231823:
Here is some wording that we should use moving foward to align

1. We should have `If checked,`, not `When this is checked` or `When this is enabled` or `When enabled` or `When active` or any other variant that you can think of.

Here are the translated versions
```
deDE: nothing has been standardized yet
esES/esMX: Si está seleccionado,
frFR: Si cette option est cochée,
koKR: 이 옵션을 선택하면
ptBR: Se marcado,
ruRU: Когда включено,
zhCN: 启用时，
zhTW: 啟用時，
```

PR #7414, opened 2026-02-12
Remove periods

Loukky:
Remove periods to shorten text length


anon1231823:
Since this also affects zhTW, @NORPG do you agree with this?

NORPG:
This is actually somewhat similar to #7405: when English uses a period, should Chinese also use a period?

Of course, this doesn't affect readability, but a consistent guideline is needed.

My suggestion is that when neither English nor Chinese uses a comma as a separator, it likely means it's not a sentence but a phrase, and a period isn't necessary. If not, I suggest maintaining the consistency with English.

Loukky:
These translated texts are not like full articles that require strict punctuation for paragraph structure.

They are short sentences or brief phrases. In this context, a period at the end is not necessary.

@anon1231823 @NORPG 
Therefore, if we want to standardize the style, I suggest removing the trailing period consistently — this could apply to non-Chinese languages as well.

anon1231823:
I agree, the descriptions for options read more as phrases rather than sentences. I just haven't really removed them because it's a lot of work. I noticed the Russian localization always omitted the periods.

PR #7462, opened 2026-03-04
Fix custom localization initialization

Jakanis:
## Issue references
I haven't created an issue for that, so I'll describe it briefly here.

Some time ago there was a change to quest lookup structure with description removal, which lead to this commit - https://github.com/Questie/Questie/commit/dd335d357ae4fd1c4d29de99e645c1e2422daacc
I guess instead of returning slice of first and third column it returns table of two rows with ids 1 and 3? I'm not sure about this, but my localization addon doesn't translate quests with old questLookup structure.

Also there was an improvement to startup performance in this commit - https://github.com/Questie/Questie/commit/4271c26df3837287dc1c8e947c9b2b8766350c43
Which at some point (didn't find a commit) had related change in lookup calls - from plain table `l10n.itemLookup[locale]` to function call `l10n.itemLookup[locale]()`, which also changed assignment in _InitializeLocaleOverride to `l10n.questLookup[overridingLocale] = function() return QUESTIE_LOCALES_OVERRIDE.questLookup end` instead of plain table assignment

## Proposed changes

Roll these changes back:
- Remove compatibility code from Questie instead of fixing and supporting it
- Force override addons to use the same structure as in Questie lookup files - no description for quests and return function instead of table
- That also allows external addons to use the same lazy evaluation approach

## Alternative
We could do two things instead:
- Check if `QUESTIE_LOCALES_OVERRIDE` lookup is a table or a function and process it accordingly
- Fix questLookup assignment compatibility for tables - build a slice if it's an old three-column table. I guess that's only possible with iteration through table with rebuilding it to a new one. Something like that:
```
l10n.questLookup[overridingLocale] = function()
    local result = {}
    for id, data in pairs(QUESTIE_LOCALES_OVERRIDE.questLookup) do
        result[id] = {data[1], data[3] or data[2]}
    end
    return result
end

```

### DIFF
--- a/Localization/l10n.lua
+++ b/Localization/l10n.lua
@@ -107,13 +107,10 @@ end
 _InitializeLocaleOverride = function()
     local overridingLocale = QUESTIE_LOCALES_OVERRIDE.locale
     supportedLocals[overridingLocale] = true
-    l10n.itemLookup[overridingLocale] = function() return QUESTIE_LOCALES_OVERRIDE.itemLookup end
-    l10n.questLookup[overridingLocale] = function()
-        -- We don't want to break compatibility with the override addons, so we simply ignore [2] which was the unused quest description
-        return {QUESTIE_LOCALES_OVERRIDE.questLookup[1], QUESTIE_LOCALES_OVERRIDE.questLookup[3]}
-    end
-    l10n.npcNameLookup[overridingLocale] = function() return QUESTIE_LOCALES_OVERRIDE.npcNameLookup end
-    l10n.objectLookup[overridingLocale] = function() return QUESTIE_LOCALES_OVERRIDE.objectLookup end
+    l10n.itemLookup[overridingLocale] = QUESTIE_LOCALES_OVERRIDE.itemLookup
+    l10n.questLookup[overridingLocale] = QUESTIE_LOCALES_OVERRIDE.questLookup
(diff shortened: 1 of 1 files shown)


Jakanis:
Well, at a first glance, it looks like a bad idea - breaking compatibility with external addons. But I currently can see only one such addon, and it's mine (and it is already a bit broken and outdated) :D
Updating it to match this change was straightforward - just removing quest description from questLookup generation and changing first and last lines of lookup files. I guess if there are some other translations in development, it won't be a problem to update them.

BreakBB:
Hey @Jakanis 

Thanks a lot for reaching out on this! Yeah, I kept the compatibility, because I didn't want to generate any work on your (and maybe other) addon(s). Since you are open to update your addon to make it follow the new structure, I am totally open to remove the compatibility.

Let me know when your addon is ready for this, then I'll merge this and ship a new release.

Jakanis:
I think external addons should be updated periodically anyway, I already see some new strings in Localization/Translations and there will be more in the future.
Also, I didn’t want to create any extra compatibility work for Questie team with this custom localization feature. And this is easy to follow, since we can just check Questie's lookup/translation files and essentially copy the structure with translated strings.

Please don't wait for me, I'll update localization addon as soon as this change is released.

Thank you for allowing us to make WoW accessible to more people. It turned out even more important than I expected. ❤️

PR #7471, opened 2026-03-11
Add Chinese translations for two quests

Loukky:


Loukky:
@anon1231823 please help to merge,I have added all the missing translations I could find.

PR #7504, opened 2026-04-12
fix: defer SetMapID after WorldMapFrame:Show() to prevent zone reset

AdamsGH:
## Problem

On some server builds (e.g. TBC Anniversary), `WorldMapFrame:OnShow` always resets the map to the player's current zone by calling `MapUtil.GetDisplayableMapForPlayer()`. This means clicking "Show on Map" in the tracker always opens the player's current zone instead of the quest objective zone.

## Fix

Call `WorldMapFrame:Show()` first, then apply `SetMapID` via `C_Timer.After(0)` so it runs after `OnShow` completes.

When the frame is already visible, `Show()` is skipped and `SetMapID` is called directly.

### DIFF
--- a/Modules/Tracker/TrackerUtils.lua
+++ b/Modules/Tracker/TrackerUtils.lua
@@ -117,13 +117,24 @@ function TrackerUtils:SetTomTomTarget(title, zone, x, y)
     end
 end
 
+-- On some server builds (e.g. TBC Anniversary), WorldMapFrame:OnShow always resets
+-- the map to the player's current zone via MapUtil.GetDisplayableMapForPlayer().
+-- C_Timer.After(0) defers SetMapID to the next frame, after OnShow completes.
+-- When the frame is already visible, Show() is skipped and SetMapID runs directly.
+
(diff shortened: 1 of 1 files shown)


BreakBB:
Hey @AdamsGH 

Thanks for taking the time to look into this and provide this fix. We appreciate your help!

BreakBB:
I can't reproduce the problem you are trying to solve though @AdamsGH  🤔 Can you explain when exactly this happens and can you verify it is a problem with only Questie active?

Things I tested:

1. WorldMap closed + "Show on Map" of an objective on a different map than the current player
2. WorldMap open + "Show on Map" of an objective on a different map than the current player
3. WorldMap open on a different map than the current player + "Show on Map" of an objective on a different map than the current player
4. WorldMap in full screen mode
5. WorldMap in window mode
6. Only Questie active
7. Questie + regular addons (including LeatrixMaps) active

BreakBB:
I am closing this as there was no further response about what the problem is this PR tries to fix.

PR #7519, opened 2026-04-29
[quest] add some quest for titan server only

Loukky:


Database/Corrections/QuestieQuestBlacklist.lua
```diff
@@ -1681,6 +1681,10 @@ function QuestieQuestBlacklist:Load()
         [95706] = not Questie.IsTitanReforged, -- Gobb's Weekly Greed Deal -- only present on Titan servers
```
Cabro, line 1682: we also need 94576 here

Loukky: add

PR #7537, opened 2026-05-19
feat: party-shared quest map overlay and log visibility

gnummers:
## Summary
This PR adds PartyQuests functionality on top of Questie to improve party/group quest visibility.

### What is included
- New Modules/PartyQuests.lua module
  - Draws synced party quest objective pins on map/minimap
  - Supports all-party or single-player focus
  - Handles missing event-coordinate payload safely (no error spam)
- Slash command support in Modules/QuestieSlash.lua
  - /questie partyquests (help)
  - /questie partyquests map [on|off|all|playerName]
  - /questie partyquests list [playerName]
  - /questie partyquests showall [on|off]
- Modules/Network/QuestieComms.lua
  - Emits QC_ID_REMOTE_QUESTLOG_UPDATED message on remote quest log updates
- Modules/QuestieInit.lua
  - Initializes PartyQuests module
- TOC updates for Classic/BCC/WotLK/Cata/Mists to load PartyQuests
- README updates for new quest communication commands

## Notes
- This relies on existing Questie comms sync; party members need Questie for shared data.
- Includes ASCII-safe chat help text for WoW chat rendering compatibility.


gnummers:
Updated this PR with the latest fixes from gnummers/master.

New in this update:
- Fixed PartyQuests objective pin generation to safely skip partial comms payloads (prevents repeated corrupted-objective error spam).
- Added missing payload guards for event/killcredit/objective-id requirements.
- Kept prior PartyQuests features intact (map overlay, list/help/showall commands).
- Bumped TOC versions to 11.27.5.

Latest head commit: 955608967.

gnummers:
Addressed the latest review suggestions in a7618ec2:

- PartyQuests manual icon IDs are now player-scoped (quest:objective:player) to avoid cross-player collisions.
- SetShowCompleted() now has behavioral effect by syncing with showOnlyObjectives; rendering logic now respects both flags.
- Fixed /questie partyquests showall toggle message to report the actual resulting mode.

Questie-BCC.toc
```diff
@@ -1,5 +1,5 @@
 ## Interface: 20505
-## Title: Questie|cFF00FF00 v11.27.2|r
+## Title: Questie|cFF00FF00 v11.27.6|r
```
BreakBB: Please never do any version bumps, that is not the concern of PRs.

BreakBB (commented):
Hey @gnummers 

Thanks A LOT for tackling this long requested feature. Our time is rather limited atm, but we'll try to get through this as soon as possible.

For settings we usually use the UI instead of slash commands, but personally I'd prefer to keep things as they are for now and change that later on. I'd even vote to activate this feature by default.

Questie-BCC.toc
gnummers: Understood!

gnummers:
> Hey @gnummers
> 
> Thanks A LOT for tackling this long requested feature. Our time is rather limited atm, but we'll try to get through this as soon as possible.
> 
> For settings we usually use the UI instead of slash commands, but personally I'd prefer to keep things as they are for now and change that later on. I'd even vote to activate this feature by default.

You're very welcome, sir!

gnummers:
Addressed the QuestieMap crash in PartyQuests: manual icon ids are now kept numeric for QuestieMap's contract, and PartyQuests uses a separate player-scoped manualKey to prevent cross-player pin collisions. Pushed in commit 7a481f1bb.

Cabro:
Hey, lovely initiative!
I took the PR to test it and have a feel for it, and this is my feedback:
1 - by far the most obvious one: this is only slash command, before we can ship this we need to have actual options selections in the options pane
2 - the descriptions are lacking
2.a - they are missing localization strings for basically everything
2.b - they don't tell you which one is default for each slash command
2.c - they don't tell you which one is currently active
2.d - `*` can be removed, it's just clutter
2.e - the commands can be compressed to be in the form of `/questie partyquests map [blank/on/off/Thrall]` with the description, so it's less spammy in chat (we are actually trying to minimize the addon's impact on users)
2.f - what's the difference between these 2?
image
2.g - it needs an escape char for this one 
image


So far what i notice with it, i have to group up and will update as i go

gnummers:
> Hey, lovely initiative! I took the PR to test it and have a feel for it, and this is my feedback: 1 - by far the most obvious one: this is only slash command, before we can ship this we need to have actual options selections in the options pane 2 - the descriptions are lacking 2.a - they are missing localization strings for basically everything 2.b - they don't tell you which one is default for each slash command 2.c - they don't tell you which one is currently active 2.d - `*` can be removed, it's just clutter 2.e - the commands can be compressed to be in the form of `/questie partyquests map [blank/on/off/Thrall]` with the description, so it's less spammy in chat (we are actually trying to minimize the addon's impact on users) 2.f - what's the difference between these 2? image
> 
> So far what i notice with it, i have to group up and will update as i go

Thank you for the feedback! Will address this.

gnummers:
Thanks again for the detailed feedback! Here's what was addressed in the latest commits:

**Point 1 — Options pane UI**: Added two toggles to the Social Options group in the General tab:
- *Show party quest map pins* (enable/disable the feature)
- *Show completed party quest objectives* (show all vs. incomplete only, disabled when pins are off)

Settings persist via `Questie.db.profile` and are restored on login.

**Point 2.a — Localization**: New option strings go through `l10n()` and are registered in `Localization/Translations/Options/Social.lua` with `enUS = true` and other locales marked `false` for translators.

**Point 2.b/2.c — Defaults / current state**: The compressed help line for each command now shows the current state inline, e.g. `map pins (now: OFF, focus: all)`.

**Point 2.d — Remove `*` bullets**: Done.

**Point 2.e — Compressed help**: Help is now 3 single lines (one per command) rather than an expanded multi-line block.

**Point 2.f — `map` vs `map on` ambiguity**: Bare `/questie partyquests map` now prints the current status rather than silently toggling, so `map on/off` is the only way to change state — no more ambiguity.

**Point 2.g — Escape char**: Replaced `[on|off]` separators with `[on/off]` throughout help text — the `|r` in `|off]` was being interpreted as a WoW chat color-reset escape sequence.

gnummers:
image

gnummers:
image

gnummers:
@coderabbitai resume

Modules/PartyQuests.lua
```diff
@@ -0,0 +1,376 @@
...
+function PartyQuests:GetShowOnlyObjectives()
+    return _PartyQuests.showOnlyObjectives
+end
+
+---@param showOnlyObjectives boolean
+function PartyQuests:SetShowOnlyObjectives(showOnlyObjectives)
+    _PartyQuests.showOnlyObjectives = showOnlyObjectives and true or false
+    _PartyQuests.showCompleted = not _PartyQuests.showOnlyObjectives
+    PartyQuests:RefreshMapPins()
+end
```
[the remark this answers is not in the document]
gnummers: Fixed, thank you.

anon1231823:
> image
> 
> 

Have the descriptions show when you hover over the option not embedded in the GUI frame, like the other options.

gnummers:
> Have the descriptions show when you hover over the option not embedded in the GUI frame, like the other options.

Done, thank you!

gnummers:
@coderabbitai review

gnummers:
> Caution
> 
> Some comments are outside the diff and can’t be posted inline due to platform limitations.
> 
> ⚠️ Outside diff range comments (1)
> > Modules/QuestieSlash.lua (1)> `206-210`: _⚠️ Potential issue_ | _🟡 Minor_ | _⚡ Quick win_
> > > **Inverted mode text in toggle case.**
> > > When toggling without an explicit `on`/`off` argument, the printed mode message is inverted relative to the new state. After toggling from incomplete-only to all, it prints "only incomplete" and vice versa.
> > > The issue: after `SetShowOnlyObjectives(not current)`, the new value of `showOnlyObjectives` is `not current`. When `not current` is false (showing all), the ternary should print "all (including completed)", but currently prints "only incomplete".
> > > 
> > > 🐛 Proposed fix to swap the ternary strings
> > > ```diff
> > >                  local current = PartyQuests:GetShowOnlyObjectives()
> > >                  PartyQuests:SetShowOnlyObjectives(not current)
> > > -                local mode = (not current) and "all (including completed)" or "only incomplete"
> > > +                local mode = (not current) and "only incomplete" or "all (including completed)"
> > >                  Questie:Print("|cFFFF6F22[PartyQuests]|r", "Now showing " .. mode .. " quest objectives.")
> > > ```

Modules/QuestieSlash.lua already has the correct mapping:

 `local mode = (not current) and "all (including completed)" or "only incomplete"`

After SetShowOnlyObjectives(not current), that message matches the new state correctly, so no code change was needed.

Modules/PartyQuests.lua
```diff
@@ -0,0 +1,425 @@
...
+---@param playerName string
+---@param objective table
+---@return table
+local function _GetTooltipBody(questName, objectiveText, playerName, objective)
+    local progressText = tostring(objective.fulfilled or 0) .. "/" .. tostring(objective.required or 0)
+    return {
+        {"Quest: ", questName},
+        {"Player: ", playerName},
+        {"Objective: ", objectiveText},
+        {"Progress: ", progressText},
```
BreakBB: What's the reason for this tooltip format? I'd prefer it to be very similar to the regular one that also shows the progress of your group.

Cabro:
In addition to the tooltips not being in the format we have, this is extra.
- icons are all bag. it should be based on objective type
- cannot create quest links from it
- when the party quest is removed, your icons are also removed if you are on the same quest.

gnummers:
> In addition to the tooltips not being in the format we have, this is extra.
> 
> * icons are all bag. it should be based on objective type
> * cannot create quest links from it
> * when the party quest is removed, your icons are also removed if you are on the same quest.

Thank you!

● Implemented — PartyQuests now addresses the regressions you called out.

   1. Icons not all bag: Party icon texture selection is now resolved from objective type (slay/object/loot/event), and no longer forced by stale objectiveData.Icon values.  
   2. Quest links + same-quest icon interference workaround: Party pins now carry QuestData for link insertion, and click handling uses QuestData.Id for correct quest links. Party pins also use isolated synthetic IDs/type isolation so removing remote party objectives no longer collides with your own quest icon bookkeeping.

Modules/PartyQuests.lua
gnummers: I don't really have a good reason. I also like the regular Questie format.

Tooltip format + group stacking + colorization: Party pins now feed CustomTooltipData in Questie-style objective lines (progress + objective + colored player name) and stack by quest/objective/player instead of the old 4-row manual format.

Addressed this in https://github.com/Questie/Questie/pull/7537/commits/010149f6ebf16edbc270815d4541e5e0b4bb39c1 and https://github.com/Questie/Questie/pull/7537/commits/a2a0852aa8af9291e1f4266a854f3898a966bb44

BreakBB:
@Cabro @Sommos this can be closed right?

gnummers:
> @Cabro @Sommos this can be closed right?

I'm still here if you need anything further.

Cabro:
> @Cabro @Sommos this can be closed right?

Yes.

BreakBB:
Hey @gnummers we took your idea and re-implemented it in #7574

Thanks a lot again for taking the time to provide this PR!

gnummers:
> Hey @gnummers we took your idea and re-implemented it in #7574
> 
> Thanks a lot again for taking the time to provide this PR!

You're welcome!

PR #7550, opened 2026-05-30
[fix] DBM HUD Overlay Fix

### LINKED ISSUE #481: DBM Overlay not hidden by ALT+Z
I'm on WoW Classic Era, but is suppose it is possible this issue affects all version of the addon.

I've noticed that the overlay is not hidden by the ALT+Z hide UI shortcut which is a bit annoying when taking screenshots.

The particular instances where this have been an issue is from Questie using your overlay to display quest related icons, but I'm guessing it would be better to go right to the source than having them create a workaround.

Thank you for your consideration on this matter and your continued work on this amazing addon. :D

anon1231823:
This PR Hides the DBM HUD Overlay icons when the UI is hidden. This is related to an issue that was made on the DBM GitHub. 

Fixes https://github.com/DeadlyBossMods/DBM-Vanilla/issues/481

### DIFF
--- a/Modules/QuestieDBMIntegration.lua
+++ b/Modules/QuestieDBMIntegration.lua
@@ -75,8 +75,7 @@ end
 --   Event/Enable/Disable Handlers  --
 -------------------------------------
 do
-    -- PARENTED TO WORLDFRAME: This ticker stays alive and clears markers even when UIParent is hidden
-    local eventFrame = CreateFrame("frame", nil, WorldFrame)
+    local eventFrame = CreateFrame("frame", "QuestieDBMIntegration", UIParent)
     local GetInstanceInfo, IsInInstance = GetInstanceInfo, IsInInstance
     local warningShown = false
 
--- a/Modules/QuestieDBMIntegration.lua
+++ b/Modules/QuestieDBMIntegration.lua
@@ -107,7 +107,7 @@ end
 --   Event/Enable/Disable Handlers  --
 -------------------------------------
 do
-    -- PARENTED TO WORLDFRAME: This ticker stays alive and clears markers even when UIParent is hidden (Alt+Z)
+    -- PARENTED TO WORLDFRAME: This ticker stays alive and clears markers even when UIParent is hidden
     local eventFrame = CreateFrame("frame", nil, WorldFrame)
     local GetInstanceInfo, IsInInstance = GetInstanceInfo, IsInInstance
     local warningShown = false
--- a/Modules/QuestieDBMIntegration.lua
+++ b/Modules/QuestieDBMIntegration.lua
@@ -85,6 +85,7 @@ local function UpdateUIVisibility()
 
     if isHidden then
         HudSuspended = true
+        DBM.HudMap:ClearAllEdges()
         for tableString in pairs(AddedHudIds) do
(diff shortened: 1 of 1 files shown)


Modules/QuestieDBMIntegration.lua
```diff
@@ -201,10 +235,22 @@ do
             DBM:Schedule(1, DelayedMapCheck)--Well, since DBM is loaded, might as well use DBM scheduler instead of creating a ticker
         end
     end)
+
+    -- Reliable UI visibility ticker monitoring WorldFrame
+    local totalElapsed = 0
+    eventFrame:SetScript("OnUpdate", function(_, elapsed)
```
BreakBB: I don't like this. There are so few people actually using this feature, but this will be run for every user on every frame no matter if they have it active or not.

Modules/QuestieDBMIntegration.lua
```diff
@@ -71,11 +74,41 @@ local function RemoveHudQuestIcon(tableString)
...
+end
+
 -------------------------------------
---  Event/Enable/Disable Handlers  --
+--   Event/Enable/Disable Handlers  --
 -------------------------------------
 do
-    local eventFrame = CreateFrame("frame", "QuestieDBMIntegration", UIParent)
+    -- PARENTED TO WORLDFRAME: This ticker stays alive and clears markers even when UIParent is hidden
+    local eventFrame = CreateFrame("frame", nil, WorldFrame)
```
BreakBB: Why was the frame name removed?

BreakBB (changes requested):

Modules/QuestieDBMIntegration.lua
BreakBB: The change is not what I meant.

I took a look into this and this is all we need to hide and show the icons:

```lua
    -- Hide nodes when the user hides the game UI
    UIParent:HookScript("OnHide", function()
        if QuestieHUDEnabled then
            CleanupPoints(9999)
        end
    end)

    -- Show nodes when the user shows the game UI
    UIParent:HookScript("OnShow", function()
        if QuestieHUDEnabled then
            ReAddHudIcons()
        end
    end)
```

I am open to keep a boolean flag variable which deactivates some things and the bug fix about `NorthrendPoints`

BreakBB (changes requested):

Modules/QuestieDBMIntegration.lua
```diff
@@ -72,10 +72,11 @@ local function RemoveHudQuestIcon(tableString)
 end
 
 -------------------------------------
---  Event/Enable/Disable Handlers  --
+--   Event/Enable/Disable Handlers  --
 -------------------------------------
 do
-    local eventFrame = CreateFrame("frame", "QuestieDBMIntegration", UIParent)
+    -- PARENTED TO WORLDFRAME: This ticker stays alive and clears markers even when UIParent is hidden
+    local eventFrame = CreateFrame("frame", nil, WorldFrame)
```
BreakBB:
```suggestion
    local eventFrame = CreateFrame("frame", "QuestieDBMIntegration", UIParent)
```

BreakBB (approved):

PR #7553, opened 2026-05-31
[Feature] Improve Journey quest filtering and hidden quest management

### LINKED ISSUE #7391: All Zones dropdown suggestion: Don't show Hidden quests
I really like the new All Zones dropdown in the Quests by Zone window. However it would be nice to have Hidden quests not show up. It might also be nice to have it respect the same filters you have to not show up on the minimap such as hiding Event quests, PVP quests, etc.

anon1231823:
This PR addresses several of the requests made in #7391. The primary features requested
~~1. Hiding quests in the journey would remove them from their respective section (available, repeatable, etc.)~~ **Overruled as an unwanted feature by an internal discussion by the Questie dev team**
2. The option that hides the icons in the GUI option also hides the respective category in the journey
3. A hide quest button available in the Quest by Zone section

## Issues
Fixes #7391 

## Screenshots
Before:
image

After:
image
image
image
image
image

### DIFF
--- a/Database/Corrections/Holidays/QuestieEvent.lua
+++ b/Database/Corrections/Holidays/QuestieEvent.lua
@@ -89,6 +89,17 @@ local DMF_CALENDER_ICON_TEXTURES = {
 }
 
 function QuestieEvent.Initialize()
+    -- Always populate event names regardless of showEventQuests setting,
+    -- so that IsEventQuest() works correctly for the Journey hidden section
+    for _, questData in pairs(QuestieEvent.eventQuests) do
+        local questId = questData[2]
+        local hideQuest = questData[5]
+        if (not hideQuest) then
+            _QuestieEvent.eventNamesForQuests[questId] = questData[1]
+            QuestieCorrections.hiddenQuests[questId] = nil
+        end
+    end
+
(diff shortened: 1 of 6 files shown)


BreakBB:
@anon1231823 as discussed on Discord, we don't want any setting that affects the Journey. I guess this PR is obsolete then, right?

anon1231823:
> @anon1231823 as discussed on Discord, we don't want any setting that affects the Journey. I guess this PR is obsolete then, right?

No, #3 is still relevant

anon1231823:
> @anon1231823 as discussed on Discord, we don't want any setting that affects the Journey. I guess this PR is obsolete then, right?

Oh, also I don't think we want to segregate this option from the journey if we are going to be consistent with what shift-click on a quest does. So I think this should function in this way.

BreakBB:
> No, 3 is still relevant

True, so let's rebuild this PR to only include that new button to hide quests. Maybe open a new one 🤔

anon1231823:
> > No, 3 is still relevant
> 
> 
> 
> True, so let's rebuild this PR to only include that new button to hide quests. Maybe open a new one 🤔 

2 is still relevant and I already removed 1

PR #7555, opened 2026-05-31
[fix]fix chopiness in updating objectives

Loukky:


Modules/Quest/QuestieQuest.lua
BreakBB: I don't like this approach tbh. There is a specific setting linked to the changes you did recently.

When that setting is not active (which won't be for 99,5% of all users), no extra code should not be run. Especially not such expensive ones as regex operations.

So `FullDescription` should default to `objective.text` or simply be `nil` when the setting of "remove slain" is not active.

Modules/Quest/QuestieQuest.lua
```diff
@@ -1493,9 +1494,12 @@ function _QuestieQuest.ObjectiveUpdate(self)
 
             self.Type = obj.type;
             self.Description = obj.text
-            self.FullDescription = string.match(obj.raw_text, "^(.*):%s*%d+/%d+$")
-                or string.match(obj.raw_text, "^(.*)：%s*%d+/%d+$")
-                or obj.text
+            if obj.raw_text ~= self._rawText then
+                self._rawText = obj.raw_text
+                self.FullDescription = string.match(obj.raw_text, "^(.*):%s*%d+/%d+$")
```
BreakBB: same as above

BreakBB (changes requested):

Loukky:
@coderabbitai resume

Modules/Libs/QuestieLib.lua
```diff
@@ -132,7 +132,19 @@ function QuestieLib:GetObjectiveDescription(objective)
...
+    if Questie.db.profile.trimObjectiveText ~= false then
+        desc = objective.Description
+    else
+        desc = objective.FullDescription
+        if not desc or desc == objective.Description then
+            local raw = objective._rawText or objective.Description
+            desc = string.match(raw, "^(.*):%s*%d+/%d+$")
+                or string.match(raw, "^(.*)：%s*%d+/%d+$")
+                or objective.Description
+            objective.FullDescription = desc
```
BreakBB: Oh no, this will change the reference while the function should only return `desc`.

This is a lot of implicit code, while it can clearly be explicit.

```
if Questie.db.profile.trimObjectiveText then
   <do string.match stuff to remove "slain">
else
    desc = objective.Description
end
```

Modules/Quest/QuestieQuest.lua
```diff
@@ -1411,17 +1411,14 @@ function QuestieQuest:PopulateQuestLogInfo(quest)
                 Questie:Error(l10n("Missing objective data for quest "), quest.Id, " ", objective.text)
             else
                 if not quest.Objectives[objectiveIndex] then
-                    local fullDesc = string.match(objective.raw_text, "^(.*):%s*%d+/%d+$")
```
BreakBB, line 1414: Why is this not using `QuestieLib:GetObjectiveDescription`?

Modules/Options/GeneralTab/QuestieOptionsGeneral.lua
```diff
@@ -504,7 +508,14 @@ function QuestieOptions.tabs.general:Initialize()
                         get = function(info) return QuestieOptions:GetProfileValue(info); end,
                         set = function(info, value)
                             QuestieOptions:SetProfileValue(info, value)
-                            QuestieTracker:Update()
+                            TrackerLinePool.UpdateObjectiveLines(function(line)
```
BreakBB: I don't like this, as it won't update the line width for example. Why is `QuestieTracker:Update()` not enough?

BreakBB (changes requested):

Modules/Quest/QuestieQuest.lua
Loukky: fixed

Modules/Quest/QuestieQuest.lua
Loukky: fixed

Modules/Options/GeneralTab/QuestieOptionsGeneral.lua
Loukky: I'm worried about causing lag in the tracker.

Loukky:
@coderabbitai resume

Modules/Quest/QuestieQuest.lua
```diff
@@ -1411,17 +1411,13 @@ function QuestieQuest:PopulateQuestLogInfo(quest)
...
-                        or objective.text
-
                     quest.Objectives[objectiveIndex] = {
                         Id = quest.ObjectiveData[objectiveIndex].Id,
                         Index = objectiveIndex,
                         questId = quest.Id,
                         _lastUpdate = 0,
                         Description = objective.text,
-                        FullDescription = fullDesc,
+                        _rawText = objective.raw_text,
```
BreakBB: I think `FullDescription` is a better name to be used here than `_rawText`.

And since this is the place where `Description` is created from the QuestLogCache objectives, I would have expected this is also the place where `FullDescription` gets set for the first time. No other place should touch `raw_text`, if at all...

So at this place I would have expected something like:

```lua
local fullDesc
if Questie.db.profile.trimObjectiveText then
    fullDesc = <string match stuff>
end

...
FullDescription = fullDesc,
...
```

And basically the same should also happen below.

Modules/Libs/QuestieLib.lua
```diff
@@ -132,7 +132,17 @@ function QuestieLib:GetObjectiveDescription(objective)
     if not objective then
         return ""
     end
-    local desc = Questie.db.profile.trimObjectiveText ~= false and objective.Description or objective.FullDescription or objective.Description
+    if Questie.db.profile.trimObjectiveText ~= false then
```
BreakBB: If you checkout the other comment first, then this place can simply become:

```lua
local desc = objective.FullDescription or objective.Description

if (not desc) then
   return ""
end

return desc:gsub("%.$", "")
```

BreakBB (changes requested):

Loukky:
@BreakBB I don't know how to optimize it further.If possible, could you optimize this PR according to your ideas?

PR #7556, opened 2026-05-31
[fix] Auto-size some buttons and fix formatting

anon1231823:
## Proposed changes

Autosizes several buttons in My Journey and Options. Also reloads UI when changing GUI locale in Advanced Options so that buttons are refitted for new locale and centered in the icons tab.

## Screenshots

image
image


Modules/Journey/tabs/MyJourney/MyJourneyTab.lua
```diff
@@ -80,7 +80,7 @@ function _QuestieJourney.myJourney:DrawTab(container)
 
     local noteButton = AceGUI:Create("Button");
     noteButton:SetText(l10n('Add New Adventure Note'));
-    noteButton:SetPoint("RIGHT");
+    noteButton:SetAutoWidth(true);
```
BreakBB, line 83: Why was `SetPoint` removed?

Modules/Options/AdvancedTab/QuestieOptionsAdvanced.lua
```diff
@@ -225,6 +225,7 @@ function QuestieOptions.tabs.advanced:Initialize()
                         Questie.db.global.questieLocale = lang;
                         Questie.db.global.questieLocaleDiff = true;
                     end
+                    ReloadUI()
```
BreakBB: If we do that, then we need to adjust the description. Afaik we let users know if changing a setting also triggers a reload.

Modules/Options/QuestieOptions.lua
```diff
@@ -100,6 +113,71 @@ function QuestieOptions:SetProfileValue(info, value)
     Questie.db.profile[info[#info]] = value
 end
 
+function QuestieOptions:RefreshIconsTabButtons(widget)
```
BreakBB, line 116: This is a lot of code for what, auto sizing buttons? Do they actually look bad atm?

anon1231823: Well for esMX they don't fit the whole word it gets truncated for profession trainers

Modules/Options/AdvancedTab/QuestieOptionsAdvanced.lua
anon1231823: Fixed by [8e85ede](https://github.com/Questie/Questie/pull/7556/commits/8e85ede4411168ead4089afcd6edf272e4369605)

Modules/Options/QuestieOptions.lua
BreakBB: I'd prefer to simply increase the width a bit. This is a lot of complexity for little benefit

anon1231823:
This isn't even a full solution. The reset nameplate button, reset target frame, reset tracker, etc. all used some janky way of placing the buttons in the frame that doesn't take into account it's placement. So simply increasing the width of any of these buttons doesn't change their relative position and you can see they immediately become off-center.

image

BreakBB:
@anon1231823 any progress on this?

If increasing the size of some buttons move them off center, then feel free to correct that. But I don't want any complex logic to find buttons and make them auto resize. That is bound to break.

Feel free to also extract the "Townsfolk group" change into a new PR and the "reload on language change" as well.

anon1231823:
> @anon1231823 any progress on this?
> 
> 
> 
> If increasing the size of some buttons move them off center, then feel free to correct that. But I don't want any complex logic to find buttons and make them auto resize. That is bound to break.
> 
> 
> 
> Feel free to also extract the "Townsfolk group" change into a new PR and the "reload on language change" as well.

The code works. I haven't heard any rebuttal other than "it's more complicated than I like the code to be". But the existing code doesnt recenter the buttons. Can you propose something that achieves the same goal?

BreakBB:
Yeah, resize the buttons manually so they fit all languages and manually fix the positioning if the size change moves them off center.

I also left [this comment](https://github.com/Questie/Questie/pull/7556#discussion_r3332587817)

PR #7561, opened 2026-06-03
[feature] Refactor Townsfolk and menu

anon1231823:
## Proposed changes

Remove translations from Questie and use global strings. In this case, using the global strings for ~~professions and~~ townsfolks

Also create a submenu for Townsfolk
image

Also just made it so the submenu opens to the right and just moves the main menu to the left if it goes off screen
image


BreakBB:
As discussed on Discord:

1. We don't want a "Townsfolk" sub-menu. Users should have these things accessible right away, without hiding them in a sub-menu.
2. Removing the Tailoring sub-menu seems like a good idea.
3. We can also think about adding anvils, forges, etc. which are required for other professions.
4. I am unsure about changing the behavior of the dropdown lib so it moves the main menu when otherwise going off screen. From an UX perspective, jumping items are a no-no 🤔 IF we want to keep this, then it needs to be made clear in the code, that we did a manual change to a third party lib (as we did for HBD for example).
5. Replacing translations with global strings is not something we want.

anon1231823:
> As discussed on Discord:
> 
>     1. We don't want a "Townsfolk" sub-menu. Users should have these things accessible right away, without hiding them in a sub-menu.
> 
>     2. Removing the Tailoring sub-menu seems like a good idea.
> 
>     3. We can also think about adding anvils, forges, etc. which are required for other professions.
> 
>     4. I am unsure about changing the behavior of the dropdown lib so it moves the main menu when otherwise going off screen. From an UX perspective, jumping items are a no-no 🤔 IF we want to keep this, then it needs to be made clear in the code, that we did a manual change to a third party lib (as we did for HBD for example).
> 
>     5. Replacing translations with global strings is not something we want.

image
The menu should open further to the left. Right now it doesn't have enough space and we can just avoid the issue of repositioning it entirely.

PR #7565, opened 2026-06-03
[feature] Show active quests in Quest by Zone

### LINKED ISSUE #2341: Show active quests in Quest by Zone.
## Description
Indication in Quest by Zone if you're currently on that quest.
Now you have to cross check with your quest log.

image

anon1231823:
## Issue references

Fixes #2341

## Proposed changes
Adds a quest log icon next to a quest in QBZ that is already in your log to signify that that quest is in your log

## Screenshots
image


Logonz:
image

New variant.

New Gutter Functionality
Switched icons
Disabled Tooltip
Added tests to guard against overwrites of our modified AceTreeGroup

Libs/AceGUI-3.0/widgets/AceGUIContainer-TreeGroup.lua
BreakBB, line 1: Is it guaranteed, the client will always serve our own, modified version of this lib, if we don't do any renaming to it, to make it unique (like we did for other libs)?

Icons/QuestBook.png
BreakBB, line 1: Even though it's not used in this PR, I am fine to add this icon nevertheless for future use.

Libs/AceGUI-3.0/widgets/AceGUIContainer-TreeGroup.test.lua
BreakBB, line 1: This is some really weird test, but it at least verifies the Questie check stays. We'll see how happy we are with this once we need to update Ace. 😅

Libs/AceGUI-3.0/widgets/AceGUIContainer-TreeGroup.test.lua
```diff
@@ -0,0 +1,71 @@
...
+            AssertContains(source, "EnableButtonTooltips(false)", path .. " should disable default TreeGroup button tooltips")
+            AssertContains(source, "temp.iconSize = 14", path .. " should set Journey quest row icon size")
+            AssertContains(source, "temp.useIconGutter = true", path .. " should opt Journey quest rows into the icon gutter")
+            AssertContains(source, "temp.iconGutterOffset = -3", path .. " should keep Journey quest row gutter tuning")
+            AssertContains(source, "Questie.icons[\"complete\"]", path .. " should use Questie's complete icon for completed current-log quests")
+            AssertContains(source, "Questie.icons[\"incomplete\"]", path .. " should use Questie's incomplete icon for incomplete current-log quests")
+            AssertNotContains(source, "Questbook.png:14:14", path .. " should not use the old inline QuestBook texture prefix")
+            AssertNotContains(source, "QuestBook.png:14:14", path .. " should not use the old inline QuestBook texture prefix")
+        end
+    end)
```
BreakBB: I am not sure we need this and/or should even do this. QBZ and QBF are Questie modules and should be tested separately, not as part of the Ace lib. Also not as "snapshot test" imo, but some actual unit test (if that is possible).

Libs/AceGUI-3.0/widgets/AceGUIContainer-TreeGroup.lua
Logonz: Honestly I was playing with just doing our own TreeGroupQuestie or something but that would have basically made a whole new file because we would still keep the old one.
The diff would suck - I can do that as a last step if we want so we can be sure.

I did do my due diligence though and checked the latest release of AceGUI and they where still on v49 and had 0 roes diff compared to the code I pushed 8 years ago.
So while no we cannot be sure,  but I increased the version so we should not race condition against another addon unless they do something similar.
The worst thing that would happen is that the icon would be inline on the same row pushing the text to the right. The other parts would be ignored as it is fully data driven.

Would love your thoughts because I was unsure

Icons/QuestBook.png
Logonz: I agree

Libs/AceGUI-3.0/widgets/AceGUIContainer-TreeGroup.test.lua
Logonz: Haha yeah I actually asked the AI to make these exact tests because I just wanted a test that made sure no one replaced this code without knowing.

Didn't want to build a whole thing for something we haven't changed in 8 years and we are talking about maybe ditching AceGUI as well.
So mostly just a cost effective way not to get a regression.
But as I said in a previous comment, the latest AceGUI treegroup has 0 difference from our 8 year old version :P

Libs/AceGUI-3.0/widgets/AceGUIContainer-TreeGroup.test.lua
Logonz: Yeah I agree, I saw then and planned to remove it actually but forgot - so good that you saw it.

Libs/AceGUI-3.0/widgets/AceGUIContainer-TreeGroup.lua
BreakBB: Since no changes have been applied for a long time to Ace TreeGroup, I am fine with just bumping the version. Might be safer to increase it more than just one version? 🤔 

I agree though that we shouldn't invest too much time into this.

PR #7567, opened 2026-06-03
[feature] TomTom to nearest quest

### LINKED ISSUE #3208: Auto set TomTom waypoint arrow to closest quest
_I brought up this feature in Discord and was asked to put in an Issue on GitHub, so here goes:_

## Description
It would be great if Questie could automatically set the TomTom waypoint arrow to the quest that is closest in proximity (so long as it is being tracked).

Examples:
- I currently have 0 quests, just picked up a new quest, TomTom would show the new quest as soon as it is tracked (if auto tracking is on, then it should show up as soon as tracking is registered)
- I have 1 or more quests, I just fully completed a quest, TomTom should point to next closest quest or to the turn-in for the quest, whatever is closer
- I have a quest that has multiple objectives, I just completed objective 1/3, TomTom should point to next closest objective
- Going with the point above, TomTom should try to stick with the current quest objective instead of pointing to a new quest, unless there is a considerable gap between the next objective and the closest next quest - This would probably have to be experimented with

I used to use a quest called QuestHelper way back in the day.  I will see if there is a repo for that and if so, try to find what they were doing as far as API's and what not.

Thanks for considering this idea, it would be a great addition to an already great addon =)

anon1231823:
## Issue references

Fixes #3208

## Proposed changes

Option to set TomTom target to nearest quest

## Screenshots

image
image

### DIFF
--- a/Localization/Translations/Options/Tracker.lua
+++ b/Localization/Translations/Options/Tracker.lua
@@ -1037,7 +1037,7 @@ local trackerOptionsLocales = {
         ["frFR"] = "Définir automatiquement |cFF54e33bTomTom|r sur la quête la plus proche",
         ["koKR"] = "가장 가까운 퀘스트에 |cFF54e33bTomTom|r 자동 설정",
         ["ptBR"] = "Definir |cFF54e33bTomTom|r automaticamente para a missão mais próxima",
-        ["ruRU"] = "Автоматически устанавливать точку |cFF54e33bTomTom|r на ближайшее задание",
+        ["ruRU"] = "Автоустановка точки |cFF54e33bTomTom|r на ближайшее задание",
         ["zhCN"] = "自动设置|cFF54e33bTomTom|r箭头指向最近的任务",
         ["zhTW"] = "自動設定 |cFF54e33bTomTom|r 到最近的任務",
     },
--- a/Localization/Translations/Options/Tracker.lua
+++ b/Localization/Translations/Options/Tracker.lua
@@ -1028,6 +1028,32 @@ local trackerOptionsLocales = {
         ["zhTW"] = "此捷徑將會設定 TomTom 箭頭指向 NPC 或第一個未完成的任務目標（如果位置資料可用）\n\n注意：此功能不適用於成就",
     },
     ---------------------------------------------------------
+    ---------------------------------------------------------
+    ["Auto-Set |cFF54e33bTomTom|r to Closest Quest"] = {
+        ["enUS"] = true,
+        ["deDE"] = "|cFF54e33bTomTom|r automatisch auf die nächste Quest setzen",
+        ["esES"] = "Establecer |cFF54e33bTomTom|r automáticamente en la misión más cercana",
+        ["esMX"] = "Establecer |cFF54e33bTomTom|r automáticamente en la misión más cercana",
+        ["frFR"] = "Définir automatiquement |cFF54e33bTomTom|r sur la quête la plus proche",
+        ["koKR"] = "가장 가까운 퀘스트에 |cFF54e33bTomTom|r 자동 설정",
+        ["ptBR"] = "Definir |cFF54e33bTomTom|r automaticamente para a missão mais próxima",
+        ["ruRU"] = "Автоматически устанавливать точку |cFF54e33bTomTom|r на ближайшее задание",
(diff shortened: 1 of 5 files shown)


Modules/Tracker/TrackerUtils.lua
```diff
@@ -119,6 +119,31 @@ function TrackerUtils:SetTomTomTarget(title, zone, x, y)
...
+                bestSpawn = spawn
+                bestZone = zone
+                bestName = name
+            end
+        end
+    end
+    if bestSpawn then
+        TrackerUtils:SetTomTomTarget(bestName, bestZone, bestSpawn[1], bestSpawn[2])
+    elseif Questie.db.char._tom_waypoint and TomTom.RemoveWaypoint then
+        TomTom:RemoveWaypoint(Questie.db.char._tom_waypoint)
```
BreakBB, line 142: How does TomTom handle `nil` as parameter here?

Modules/Options/QuestieOptionsDefaults.lua
```diff
@@ -101,6 +101,7 @@ function QuestieOptionsDefaults:Load()
             alwaysShowTracker = false,
             trackerColorObjectives = "minimal",
             trackerSortObjectives = "byZone",
+            autoSetTomTom = false,
```
BreakBB, line 104: This again needs a migration

Modules/Tracker/QuestieTracker.lua
```diff
@@ -1581,6 +1582,11 @@ function QuestieTracker:Update()
             end)
         end)
     end
+
+    if Questie.db.profile.autoSetTomTom and IsAddOnLoaded("TomTom") and (GetTime() - lastTomTomUpdate) >= 1.0 then
+        lastTomTomUpdate = GetTime()
+        TrackerUtils:AutoSetTomTomClosestQuest()
```
BreakBB, line 1588: My first thought was: This is not the correct place to update the TomTom target, because the player moves. Now I am asking myself what the correct place is 🤔 Because what if the player takes a detour and then the target automatically changes to a different quest?

Localization/Translations/Options/Tracker.lua
```diff
@@ -1028,6 +1028,32 @@ local trackerOptionsLocales = {
...
+        ["frFR"] = "Définir automatiquement |cFF54e33bTomTom|r sur la quête la plus proche",
+        ["koKR"] = "가장 가까운 퀘스트에 |cFF54e33bTomTom|r 자동 설정",
+        ["ptBR"] = "Definir |cFF54e33bTomTom|r automaticamente para a missão mais próxima",
+        ["ruRU"] = "Автоустановка точки |cFF54e33bTomTom|r на ближайшее задание",
+        ["zhCN"] = "自动设置|cFF54e33bTomTom|r箭头指向最近的任务",
+        ["zhTW"] = "自動設定 |cFF54e33bTomTom|r 到最近的任務",
+    },
+    ["Automatically sets the TomTom waypoint arrow to the closest quest objective or turn-in."] = {
+        ["enUS"] = true,
+        ["deDE"] = "Setzt den TomTom-Wegpunktpfeil automatisch auf das nächste Questziel oder den nächsten Abgabeort.",
```
BreakBB, line 1046:
```suggestion
        ["deDE"] = "Setzt den TomTom-Pfeil automatisch auf das nächste Questziel oder den nächsten Abgabeort.",
```

BreakBB (changes requested):
We need to think about manually setting a TomTom target through the Tracker. In the current state the manual target is overwriting on each Tracker update.

anon1231823:
> We need to think about manually setting a TomTom target through the Tracker. In the current state the manual target is overwriting on each Tracker update.

This PR is about setting TomTom targets automatically, so what exactly do we need to think about manual targets for

BreakBB:
> > We need to think about manually setting a TomTom target through the Tracker. In the current state the manual target is overwriting on each Tracker update.
> 
> This PR is about setting TomTom targets automatically, so what exactly do we need to think about manual targets for

If the user manually sets the TomTom target, then it get overridden whenever the next Tracker:Update happens (with the current implementation). Sounds very frustrating to me, because there might be situations where I want to focus on a different quest first and don't want to disable the setting.

My thinking about this is, that a new target is only set if the current target is cleared. That way the automatic waypoint is not randomly removed and a manual override also won't be just removed.

PR #7572, opened 2026-06-04
[feature] breadcrumb tooltip arrow

### LINKED ISSUE #7238: Add breadcrumb quests to tooltip
### Description

So when you press shift on a quest that is part of a chain, you can see "Next in chain". I would like the same thing for quests that have a breadcrumb. So if you held shift, it would show you any breadcrumb quests for the quest you are looking at.

anon1231823:
## Issue references

Fixes #7238

## Proposed changes

Arrow pointing up when you hold shift on a quest that shows breadcrumb quests.

## Screenshots

image


Modules/Options/QuestieOptionsDefaults.lua
```diff
@@ -56,6 +56,7 @@ function QuestieOptionsDefaults:Load()
             enableTooltips = true,
             enableTooltipsItemID = false,
             enableTooltipsNPCID = false,
+            enableTooltipsBreadcrumbQuests = true,
```
BreakBB: This again needs a migration

Modules/Tooltips/MapIconTooltip.lua
```diff
@@ -661,7 +682,7 @@ end
 ---@param questId QuestId
 ---@param questLevel number
 ---@return string, string
-function _MapIconTooltip.GetNextQuestInChainLines(questId, questLevel)
+function _MapIconTooltip.GetNextQuestInChainLines(questId, questLevel, indent)
```
BreakBB, line 685: LuaType is missing

Modules/Tooltips/MapIconTooltip.lua
```diff
@@ -233,6 +233,27 @@ function MapIconTooltip:Show()
 
                 if questData.title ~= nil then
                     local quest = QuestieDB.GetQuest(questData.questId)
+
+                    if Questie.db.profile.enableTooltipsNextInChain then
+                        local breadcrumbs = QuestieDB.QueryQuestSingle(questData.questId, "breadcrumbs")
+                        if Questie.db.profile.enableTooltipsBreadcrumbQuests and shift and breadcrumbs then
```
BreakBB: Check for the new setting and `shift` in the line about the breadcrumb query, so the DB is only called when all required checks pass.

BreakBB (changes requested):
@Cabro what do you think about the arrow?

Modules/Options/QuestieOptionsDefaults.lua
anon1231823: Well you need to tell me which number. I have no idea if it should be 29 or 30 or something else

BreakBB: We need a new migration number whenever the last number was already shipped in a release.

anon1231823:
@BreakBB all your concerns have been addressed. Can this be merged?

Modules/Tooltips/MapIconTooltip.lua
BreakBB: Looks like the type was added. We don't use `---@type fun` for function declarations. The previous was correct as it seems 🤔

Modules/Migration.lua
```diff
@@ -202,6 +202,9 @@ local migrationFunctions = {
             Questie.db.profile.trackerFontOutline = ""
         end
     end,
+    [34] = function()
+        Questie.db.profile.enableTooltipsBreadcrumbQuests = true
```
BreakBB: I think we decided in the team to leave breadcrumb related stuff off by default as most users don't give a damn about them.

Modules/Tooltips/MapIconTooltip.lua
```diff
@@ -239,6 +239,27 @@ function MapIconTooltip:Show()
 
                 if questData.title ~= nil then
                     local quest = QuestieDB.GetQuest(questData.questId)
+
+                    if Questie.db.profile.enableTooltipsNextInChain and Questie.db.profile.enableTooltipsBreadcrumbQuests and shift then
```
BreakBB: Why do we need the `enableTooltipsNextInChain` check here as well? They shouldn't be connected, are they?

anon1231823:
@BreakBB ok resolved

BreakBB (approved):

PR #7574, opened 2026-06-05
[feature] Show party members quests on map

### LINKED ISSUE #7477: Add toggle to show party member's quests on map
### Description

Requesting an option in the **Classic TBC version of Questie** to show **quest objectives from party members’ quests**, even if the player **does not have the quest** or **already completed it**.

Right now Questie only shows objectives for quests the player currently has. When leveling in a group this becomes annoying because party members are often on different quests in the same area, or someone already completed a quest but is helping others finish it.

Because of that, the map and minimap do not show the relevant mobs/objects/areas for the rest of the party.

**Requested behavior**

Add a **toggle in Questie settings**, something like:

`Show Party Members' Quest Objectives`

When enabled:

- Quest objectives from **party members’ active quests** appear on the **world map and minimap**
- These objectives should show even if:
  - the player **doesn't have the quest**
  - the player **already completed the quest**

Objectives should include normal Questie markers:

- mobs to kill  
- items to collect  
- objects to interact with  
- area objectives

**Example use case**

Player A and Player B are questing together.

- Player A already completed a quest.
- Player B still needs to kill mobs for it.

Player A currently **cannot see the mob markers**, even though they are helping Player B complete the quest.

With this feature enabled, Player A would still see the quest markers so they can help efficiently.

**UI suggestion**

To avoid confusion, party objectives could be visually different, for example:

- slightly different color
- small party icon overlay
- tooltip showing which party member has the quest

Example tooltip:

Quest Objective  
Party Member: PlayerName

**Summary**

Add a toggle that allows Questie to display **party members' quest objectives on the map and minimap**, regardless of whether the player has the quest themselves. This would make group questing much easier.

Sommos:
## Issue references

Fixes #7477

## Proposed changes

- ADD: ability to see other party members quests on the world map
- ADD: ability to see other party members extraObjectives on the world map
- ADD: frame chunking to process quests
- ADD: transparency to other party members quest objective icons

## Screenshots

image
image
image


Sommos:
@coderabbitai full review

Modules/Network/QuestiePartyObjectives.lua
```diff
@@ -0,0 +1,367 @@
...
+    local stop = math.min(index + CHUNK_SIZE - 1, #queue)
+    for i = index, stop do
+        local questId = queue[i]
+        _ClearQuest(questId)
+        _DrawQuest(questId)
+    end
+
+    if stop < #queue then
+        C_Timer.After(0, function() _ProcessQueue(queue, stop + 1) end)
+    else
```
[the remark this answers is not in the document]
Sommos: fe6ce9e

Localization/Translations/Options/Social.lua
```diff
@@ -303,6 +303,30 @@ local socialOptionsLocales = {
         ["zhCN"] = "不在队伍时会用系统白色显示你的任务进度信息，其他玩家无法看到",
         ["zhTW"] = "不在隊伍中時會用系統白字顯示你的任務進度訊息，其他玩家無法看到",
     },
+    ["Show party members' quest objectives"] = {
```
BreakBB: We need translations

Modules/EventHandler/EventHandler.lua
```diff
@@ -347,6 +347,12 @@ function EventHandler:RegisterLateEvents()
     Questie:RegisterEvent("GROUP_JOINED", GroupEventHandler.GroupJoined)
     Questie:RegisterEvent("GROUP_LEFT", GroupEventHandler.GroupLeft)
 
+    -- On a /reload (or login) while already in a group, GROUP_JOINED does not fire, so request party quest logs now;
```
BreakBB, line 350: Oh wow, we missed this for quite some time. Great finding and great comment 👍🏻

Modules/Network/QuestieComms.lua
```diff
@@ -20,6 +20,10 @@ local ZoneDB = QuestieLoader:ImportModule("ZoneDB")
 local l10n = QuestieLoader:ImportModule("l10n")
 ---@type QuestLogCache
 local QuestLogCache = QuestieLoader:ImportModule("QuestLogCache")
+---@type QuestiePartyObjectives
```
BreakBB, line 23: God I really hate the "old" comms... Moving all of this to `Modules/Network/Comms.lua` would be quite the refactoring though...

Modules/Network/QuestiePartyObjectives.lua
```diff
@@ -0,0 +1,396 @@
...
+
+    if #queue == 0 then
+        return
+    end
+    processing = true
+    _ProcessQueue(queue, 1)
+end
+
+-- Debounce: coalesce bursts of incoming packets into a single processing pass.
+_Kick = function()
```
BreakBB: Even though I like this name, I don't think it's easy to understand what it actually does 😂

Modules/Network/QuestiePartyObjectives.lua
```diff
@@ -0,0 +1,396 @@
+---@class QuestiePartyObjectives : QuestieModule
+local QuestiePartyObjectives = QuestieLoader:CreateModule("QuestiePartyObjectives")
+---@class QuestiePartyObjectivesPrivate
+QuestiePartyObjectives.private = QuestiePartyObjectives.private or {}
+---@type QuestiePartyObjectivesPrivate
+local _QuestiePartyObjectives = QuestiePartyObjectives.private
```
BreakBB: This is unused
```suggestion
```

Modules/Network/QuestiePartyObjectives.lua
```diff
@@ -0,0 +1,396 @@
...
+
+    for objectiveIndex, remoteObjective in pairs(neededIndices) do
+        if drawnIconCount + iconCount >= MAX_PARTY_ICONS then
+            break
+        end
+
+        -- Prefer the database ObjectiveData (canonical Type/Id); fall back to comms data.
+        local objData = quest.ObjectiveData[objectiveIndex]
+        local objType = objData and objData.Type or typeCharToFull[remoteObjective.type]
+        local objId = objData and objData.Id or remoteObjective.id
```
BreakBB: In which cases do we need the fallbacks?

Modules/Network/QuestiePartyObjectives.lua
```diff
@@ -0,0 +1,396 @@
...
+
+        if objType and objId then
```
BreakBB, line 273: Can it actually happen that both are `nil`?

Modules/Network/QuestiePartyObjectives.lua
```diff
@@ -0,0 +1,396 @@
...
+            local cachedSpawnList = spawnListCache[questId] and spawnListCache[questId][objectiveIndex]
+            local objective = {
+                Id = objId,
+                Type = objType,
+                Index = objectiveIndex,
+                questId = questId,
+                Description = (objData and objData.Text) or _GetObjectiveName(objType, objId) or "",
```
BreakBB: We should also add the recently added `FullDescription` which does not strip the "slain" from the objective text. That makes sense for some locales.

Modules/Network/QuestiePartyObjectives.lua
```diff
@@ -0,0 +1,396 @@
...
+                IsPartyObjective = true,
+                -- Skip in-world unit/item tooltip registration; the map icon tooltip already
+                -- shows party members via QuestieComms, and this avoids leaking tooltip entries.
+                hasRegisteredTooltips = true,
+                registeredItemTooltips = true,
+            }
+
+            -- must be p-called (matches how QuestieQuest calls it internally)
+            local ok = pcall(QuestieQuest.PopulateObjective, QuestieQuest, quest, objectiveIndex, objective, true)
+            if ok then
```
BreakBB, line 308: Do we want logs/error when this is not ok?

Modules/Network/QuestiePartyObjectives.lua
```diff
@@ -0,0 +1,396 @@
...
+        return
+    end
+
+    local players = QuestieComms.remoteQuestLogs[questId]
+    if not players then
+        return
+    end
+
+    -- An objective index is drawn if at least one online party member still needs it. Offline
+    -- members are ignored so their icons disappear until they reconnect.
```
BreakBB, line 229: Are we already listening to an event which fires when a user goes offline and comes back online? I guess `GroupRosterUpdate`?

Modules/Options/QuestieOptionsDefaults.lua
```diff
@@ -141,6 +141,7 @@ function QuestieOptionsDefaults:Load()
             questAnnounceCompleted = false,
             questAnnounceIncompleteBreadcrumb = true,
             questAnnounceLocally = false,
+            showPartyQuestObjectives = true,
```
BreakBB, line 145: We need a migration for this.

Modules/QuestiePlayer.lua
```diff
@@ -121,27 +121,21 @@ function QuestiePlayer:GetCurrentContinentId()
...
-                player.colorHex = argbHex;
-                return player;
-            end
-            if(index > 6 and not UnitInRaid("player")) then
-                break;
-            end
+    -- Resolve the name through the unit API instead of string-comparing against names built
+    -- from UnitName: comms sender names are normalized full names (e.g. "Bob-OldBlanchy"),
+    -- which don't reliably match a manual "name-realm" concatenation for cross-realm members.
+    if(UnitInParty(playerName) or UnitInRaid(playerName)) then
```
BreakBB, line 127: Nicely done and thanks for also adjusting the tests right away 👍🏻

Modules/Network/QuestieComms.lua
Sommos: Yeah, it's not in a great position right now. 

If you have the time, maybe you could write up an issue with the steps of what would be involved in refactoring it to make it better and I could take a swing at it after this? :)

Modules/Network/QuestiePartyObjectives.lua
Sommos: Good catch, crazy that CodeRabbit was unable to pick that up.

8e5c3e3

Modules/Network/QuestiePartyObjectives.lua
Sommos: 8e5c3e3

Modules/Options/QuestieOptionsDefaults.lua
Sommos: Slipped my mind, good catch!

8e5c3e3

Modules/Network/QuestiePartyObjectives.lua
Sommos: e098ccf

Modules/Network/QuestiePartyObjectives.lua
Sommos: Not that both are nil, but each can be nil by itself.

Modules/Network/QuestiePartyObjectives.lua
Sommos: Should only need it for `objData`, `objType` and `objId` do not need fallbacks.

a2997ca

Modules/Network/QuestiePartyObjectives.lua
Sommos: Good shout!

9ebcd50

Modules/Network/QuestiePartyObjectives.lua
Sommos: Yes, it's wired up to `GroupRosterUpdate` which means it's giving us ~ 2.5 seconds of latency. 1 second for `ScheduleUpdate` and 1.5 seconds for the `_ScheduleProcessing` debouncing.

Modules/Network/QuestiePartyObjectives.lua
```diff
@@ -0,0 +1,433 @@
...
+    for _, objective in pairs(entry.objectives) do
+        _UnloadObjective(objective)
+    end
+    drawnIconCount = drawnIconCount - entry.iconCount
+    drawnByQuest[questId] = nil
+end
+
+-- Draw a single party quest's objectives (assumes it has already been cleared).
+---@param questId number
+local function _DrawQuest(questId)
```
Logonz, line 217: I wonder if we should try and split the file apart a bit.

Have one be a generator that is basically Get data, what should be drawn and so on.
And on that is a consumer that actually does a lot of the calls against the rest of the addon.

Create a nice data format from this file which is consumed - transformation would then happen outside of this file.
This could really help with actually creating real tests around this code.

Do the Modules/Network/QuestiePartyObjectives really need to know about Unload and PopulateObjectives and so on.
Split it up by the networking layer and map layer - move the map code closer to the actual map handlers.

BreakBB: I also think this is a good idea. But honestly we should do that in a separate PR at this point.

I suggest you add a comment @Sommos and/or open up an issue for it, so we don't forget about it entirely.

Modules/Network/QuestiePartyObjectives.lua
```diff
@@ -0,0 +1,488 @@
...
+    end
+    local objectives = C_QuestLog.GetQuestObjectives(questId)
+    local objective = objectives and objectives[objectiveIndex]
+    local text = objective and objective.text
+    if (not text) or text == "" then
+        return nil
+    end
+    -- Strip the trailing progress counter ("...: n/m"); the tooltip prepends fulfilled/required
+    -- separately. Same regex used by _GetFullDescription and QuestieQuest:PopulateQuestLogInfo.
+    return string.match(text, "^(.*):%s*%d+/%d+$") or string.match(text, "^(.*)：%s*%d+/%d+$") or text
```
BreakBB, line 127: We should extract a function in near future so this regex monster not duplicated over and over again 🙈

Modules/Options/GeneralTab/QuestieOptionsGeneral.lua
```diff
@@ -96,6 +98,18 @@ function QuestieOptions.tabs.general:Initialize()
                             Questie:Debug(Questie.DEBUG_DEVELOP, "Quest announce locally changed to:", value)
                         end,
                     },
+                    showPartyQuestObjectives = {
```
BreakBB, line 101: Do we have and/or need an option to prevent Questie from sharing the progress of the local user with the party?

BreakBB (commented):
Looks good to me for a first great version. Yes, there is room for improvement, but let's ship a first version 🥳

Modules/Options/GeneralTab/QuestieOptionsGeneral.lua
Cabro: I personally do not find a need for such an option. It shouldn't affect the player's experience if Questie is sharing the info. If anything it prevents some less than honest intents?

BreakBB: Fine to me 👍🏻

BreakBB (approved):

Issue #5267, opened 2023-12-09
Darkmoon Faire not handled in WotLK


BreakBB:
## Bug description

As follow up for #5264 we also need to fix Darkmoon Faire for WotLK. There are TODOs in the code:

- Fix the date detection, as with WotLK Blizzard changed the starting date to the first Sunday of the month (check ingame calendar for end date)
- Also with TBC Terrokar Forest was added as DMF location. The location rotates: Mulgore -> Elwynn -> Terrokar -> Mulgore ...

Cabro:
Small correction, this change started with TBC, not wotlk

anon1231823:
Was this ever resolved?

BreakBB:
Nope, still an issue

anon1231823:
<img width="2538" height="1439" alt="Image" src="https://github.com/user-attachments/assets/4f64e879-4a4d-411b-8d3b-b56358178f97" />

For what it's worth, here on Era it's Monday at 00:23 in the screenshot. Local time is 21:23 but server is 3 hours ahead so 00:23.

Quests are showing but the Faire is not open yet. I have been told by my guild that it starts at 03:00 ST.

PR #7579, opened 2026-06-08
[Fix] DMF Timer for Era

anon1231823:
## Issue references

Tangentially related to #5267

## Proposed changes

- Refactor Code
- Start at 3am ST for DMF for Era

### DIFF
--- a/Database/Corrections/Holidays/QuestieEvent.lua
+++ b/Database/Corrections/Holidays/QuestieEvent.lua
@@ -239,41 +239,33 @@ _GetDarkmoonFaireLocationEra = function(currentDate)
     local eventLocation = (currentDate.month % 2) == 0 and DMF_LOCATIONS.MULGORE or DMF_LOCATIONS.ELWYNN_FOREST
 
     local dayOfMonth = currentDate.monthDay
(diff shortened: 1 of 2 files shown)


Database/Corrections/Holidays/QuestieEvent.lua
```diff
@@ -239,41 +239,33 @@ _GetDarkmoonFaireLocationEra = function(currentDate)
...
+        [4] = 6,  -- 1st is a Wednesday -> Monday is 6th
+        [5] = 5,  -- 1st is a Thursday -> Monday is 5th
+        [6] = 4,  -- 1st is a Friday -> Monday is 4th
+        [7] = 10, -- 1st is a Saturday -> Monday is 10th
+    }
+
+    local startDay = startDayByFirstWeekday[firstWeekday]
+    if not startDay then
+        return DMF_LOCATIONS.NONE
+    end
```
BreakBB: This should never happen. If it does it's okay to crash, so we notice. That's why I removed it.


```suggestion
```

BreakBB:
@anon1231823 I slightly adjusted the PR. Would be nice if you could validate the changes. It looks good to me.

Issue #3177, opened 2021-09-24
Line wrapping on quest objective tooltips


silverhawke249:
Discord user Fizz#1777 pointed out that long objective texts causes the tooltip to stretch really wide:
![image](https://user-images.githubusercontent.com/50395934/134629955-8b1ce27b-2ed7-4067-8187-4f55858238f2.png)

It would be nice to implement a word wrap feature so that long objective texts like these don't overly stretch the tooltip box.

Issue #5138, opened 2023-09-27
Word wrap quest lines in tracker.


lordporkbone:
basically i play wow in windowed mode because i have a ultrawide monitor. sometimes when i pick up a quest, the tracker displays long quest descriptions in on one line and pushes the tracker into the middle of the screen.

MyTechnoHunter:
by default the tracker is set to grow for both height and width.. you can change this by using the sizer point of the tracker.. once you manually size it it stays manual till reset. 
some quests have objective text that is all one line (thus pushing the tracker like you described)
setting the tracker manual will disable the autogrow and will split the description to multi lines

i suggest placing the top of the tracker where you want it, then using the sizer to drag the tracker lower bounds to the FULL height of what you want your tracker to grow to, then let it go to have it snap back to the height it was for the number of quests currently in your tracker list

248bdh:
In fact, after manually changing the size of the task tracker, the task text still appears to be unable to wrap.
At this point, text that exceeds the tracker box will not be displayed.
Also, newly added tasks will not be displayed.
This seems to be because the size of the task tracker is fixed and cannot be further extended downwards.

Cabro:
This is something that Questie really needs, tbh

Logonz:
Related work in #7593 fixed UTF-8 wrapping for tooltip/Journey text by introducing `WrappedText` and `TooltipLayout`, but this issue is still not fixed by that PR.

The tracker uses a separate rendering path and does not currently use `WrappedText`/`TooltipLayout`.

Relevant tracker code:
- `Modules/Tracker/QuestieTracker.lua` `_UpdateLineWidth(...)`
- objective text set in `Modules/Tracker/QuestieTracker.lua` around the quest objective `line.label:SetText(...)` calls
- tracker labels created in `Modules/Tracker/LinePool/TrackerLine.lua`
- wrapped line sizing adjusted in `Modules/Tracker/LinePool/TrackerLinePool.lua`

Likely follow-up:
- apply tracker-specific wrapping to objective/completion/achievement/scenario objective lines, or adapt `WrappedText` for tracker labels
- preserve manual tracker sizing and height calculation behavior

PR #7593, opened 2026-06-13
UTF8 support and improved TextWrap

### LINKED ISSUE #7521: Rewrite the TextWrap function to not duplicate chinese chars and to work with spaceless strings

Logonz:
Fixes #7521 #3177 

# Main changes:
- Extracted wrapping from QuestieLib:TextWrap into Modules/Libs/WrappedText.lua.
- Added Modules/Tooltips/TooltipLayout.lua for two-pass tooltip row layout.
- Updated MapIconTooltip to build rows first, measure fixed rows, then wrap descriptions.
- Added UTF-8-safe wrapping support for Lua 5.1.
- Added numeric-token protection for localized/CJK text.
- Added tooltip measurement using actual GameTooltip render fonts.
- Added tests for wrapped text and tooltip layout behavior.

# Problem

Old QuestieLib:TextWrap had several issues:

- Lua 5.1 string functions work on bytes, not UTF-8 characters.
- CJK text often has no spaces, causing bad wrap points and duplicated/missing characters.
- Tooltip descriptions were wrapped while GameTooltip width was still changing.
- Measurement used the wrong font/width in some tooltip paths.
- Numeric expressions like 1,050, 50%, and full-width variants could be split.

# What changed

### Wrapped text module

Added Modules/Libs/WrappedText.lua.

WrappedText:TextWrap(...) now owns generic wrapping behavior:
- UTF-8-safe character boundaries.
- Space-preferred wrapping for Latin text.
- Width fallback when CalculateScreenAreaFromCharacterSpan() is unavailable or unreliable.
- Numeric token protection:
    - 1050
    - 1,050
    - 3.14
    - 50%
    - full-width digit/separator/percent variants
- Optional orphan trailing word/glyph combining.
- Optional font source so wrapping can match actual render font.

QuestieLib:TextWrap was removed. Callers now use WrappedText directly.

### Tooltip layout module

Added Modules/Tooltips/TooltipLayout.lua.

Tooltip rendering now uses a row model:
- AddLine
- AddDoubleLine
- AddDescription

Render flow:
1. Build tooltip rows.
2. Measure non-description rows only.
3. Derive stable tooltip content width.
4. Subtract description prefix width.
5. Wrap description rows.
6. Render final rows once.

This prevents description text from changing the width used to wrap itself.

Also:
- Measures AddDoubleLine with left + right widths + Blizzard gap.
- Uses actual GameTooltipTextLeft/Right font sources.
- Skips all measurement/expansion when no description rows exist.

### Map icon tooltip changes

Updated Modules/Tooltips/MapIconTooltip.lua to use TooltipLayout.

Notable behavior:
- Quest descriptions use AddDescription.
- Reputation and next-chain icons remain semantic to MapIconTooltip.
- Indentation uses deterministic transparent texture prefixes via TooltipLayout.CreateIndentUI(width).
- Reputation block stays before next-chain block.
- “Shift-click to hide” uses translated plain text plus Questie:Colorize.

### Journey tooltip changes

Updated Modules/Journey/QuestieJourneyUtils.lua to call WrappedText:TextWrap(...) directly.

### TOC updates

Registered new modules in all expansion TOCs:
- Modules\Libs\utf8.lua
- Modules\Libs\WrappedText.lua
- Modules\Tooltips\TooltipLayout.lua


Modules/Libs/utf8.lua
```diff
@@ -0,0 +1,59 @@
...
+---actual UTF-8 characters for proper length calculation.
+---@param s string UTF-8 encoded string
+---@return number Count of UTF-8 codepoints/characters
+function utf8.strlen(s)
+  local count = 0
+  for _ in s:gmatch(_CHARPAT) do
+    count = count + 1
+  end
+  return count
+end
```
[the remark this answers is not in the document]
Logonz: We use our own system for import modules, so we do not always need to do a return of the main module.

anon1231823:
Can you add the issues this fixes so they auto close when this merges.

Loukky:
image

Per Chinese national standard GB/T 15835-2011 (Rules for Writing Numerals in Publications), section 5.1.7.2 on line breaks: any number written with Arabic digits must stay entirely on one line and cannot be split between two lines.
Other than that, everything else is great.

Modules/Libs/WrappedText.lua
```diff
@@ -0,0 +1,334 @@
...
+    end
+
+    return lineEndIndex, lineEndIndex + 1, false
+end
+
+---@param character string Single UTF-8 character.
+---@param index number UTF-8 character index of `character`.
+---@param lastSpaceIndex number? Previous space boundary.
+---@return number? lastSpaceIndex Updated space boundary.
+local function _UpdateLastBreakIndex(character, index, lastSpaceIndex)
```
BreakBB: This is not an _update_ function 🤔 I guess `_GetLastBreakIndex` is the better name.

Modules/Libs/WrappedText.test.lua
```diff
@@ -0,0 +1,208 @@
...
+                end,
+            }
+        end
+
+        before_each(function()
+            originalUIParent = _G.UIParent
+            originalQuestLogObjectivesText = _G["QuestLogObjectivesText"]
+
+            require("Modules.Libs.utf8")
+            utf8 = QuestieLoader:ImportModule("utf8")
```
BreakBB: if you return `utf8` from the utf8 module, then you can just use


```suggestion
            utf8 = require("Modules.Libs.utf8")
```

No need to add the QuestieLoader to unit tests.

Modules/Libs/WrappedText.test.lua
```diff
@@ -0,0 +1,208 @@
...
+                        areas[#areas + 1] = {left = 0, bottom = 0, width = width, height = 1}
+                    end
+
+                    return areas
+                end,
+            }
+        end
+
+        before_each(function()
+            originalUIParent = _G.UIParent
```
BreakBB, line 64: `originalUIParent` is a big name for the mock in `setupTests.lua` 😆 

I am not sure if we actually need this, because each unit test elsewhere simply has `dofile("setupTests.lua")`

Modules/Libs/WrappedText.test.lua
```diff
@@ -0,0 +1,208 @@
...
+            _G["QuestLogObjectivesText"] = {
+                GetWidth = function() return 275 end,
+                GetFont = function() return "Font", 12, "" end,
+            }
+            _G.UIParent = {
+                CreateFontString = function() return CreateTextWrapFontStringMock() end,
+            }
+
+            package.loaded["Modules.Libs.WrappedText"] = nil
+            WrappedText = require("Modules.Libs.WrappedText")
```
BreakBB, line 77: I think this has the same effect as using `dofile`, doesn't it?

Modules/Libs/WrappedText.test.lua
```diff
@@ -0,0 +1,208 @@
+dofile("setupTests.lua")
+
+describe("WrappedText", function()
+    ---@type WrappedText
+    local WrappedText
+
+    describe("TextWrap", function()
```
BreakBB, line 7: There are no tests for the `prefix` parameter and also none for `fontSource`. I know that `prefix` is used, therefore a simple test would be great.

Modules/Tooltips/TooltipLayout.lua
```diff
@@ -0,0 +1,328 @@
...
+--Import modules.
+-------------------------
+---@type WrappedText
+local WrappedText = QuestieLoader:ImportModule("WrappedText")
+
+local tinsert = table.insert
+
+local MIN_TOOLTIP_TEXT_WIDTH = 375
+local DEFAULT_DOUBLE_LINE_GAP = 38.4
+local INDENT_TEXTURE_PATH = "Interface\\Minimap\\UI-bonusobjectiveblob-inside.blp"
```
BreakBB, line 14: Is this one present in all client version? Could it be safer to use an icon we have in the Icons dir?

Modules/Tooltips/TooltipLayout.lua
```diff
@@ -0,0 +1,328 @@
...
+    rightText = _G[tooltipName .. "TextRight2"]
+    if (leftText and rightText and leftText:GetRight() and rightText:GetLeft()) then
+        gap2 = rightText:GetLeft() - leftText:GetRight()
+    end
+
+    tooltipDoubleLineGapTooltip:Hide()
+    tooltipDoubleLineGapTooltip:ClearLines()
+
+    tooltipDoubleLineGapAvg = (gap1 + gap2) / 2
+    return tooltipDoubleLineGapAvg or DEFAULT_DOUBLE_LINE_GAP
```
BreakBB: `or DEFAULT_DOUBLE_LINE_GAP` can never happen, can it?

Modules/Tooltips/TooltipLayout.test.lua
```diff
@@ -0,0 +1,199 @@
...
+        _G.UIParent = originalUIParent
+        _G.CreateFrame = originalCreateFrame
+        _G.TestTooltipTextLeft1 = nil
+        _G.TestTooltipTextLeft2 = nil
+        _G.TestTooltipTextRight1 = nil
+        _G.QuestieTooltipLayoutGapMeasureTooltipTextLeft1 = nil
+        _G.QuestieTooltipLayoutGapMeasureTooltipTextRight1 = nil
+        _G.QuestieTooltipLayoutGapMeasureTooltipTextLeft2 = nil
+        _G.QuestieTooltipLayoutGapMeasureTooltipTextRight2 = nil
+        if WrappedText then
```
BreakBB: This should always be true, shouldn't it?

BreakBB (commented):
There are also merge conflicts

Modules/Libs/WrappedText.lua
Logonz: Honestly, maybe - but ill do you one better and just remove this function fully - it is a bit much in terms of syntactic sugar just for 
if character == " " then
 lastSpaceIndex = endIndex
end

Modules/Tooltips/TooltipLayout.lua
Logonz: No you are right, it is a leftover from before the avg check

Modules/Tooltips/TooltipLayout.test.lua
Logonz: Yes, correct!

Modules/Tooltips/TooltipLayout.lua
Logonz: It should be fine because otherwise it would not work on current master - because i reuse a texture that has been used for the same purpose today.
https://github.com/Questie/Questie/blob/fe7cf26e4717f2df4f430ae703b879d5df6d24c6/Modules/Tooltips/MapIconTooltip.lua#L36

---

A case can of course be made to use our own transparent texture.

Modules/Libs/WrappedText.test.lua
Logonz: Keeping this because the test overwrites _G.UIParent. Restoring prevents leakage into later tests.
We modify it each run so for safety surrounding other tests this is preferable imo?

Modules/Libs/WrappedText.test.lua
Logonz: WrappedText.lua captures globals at load time (QuestLogObjectivesText, UIParent behavior). Test sets mocks first, then forces fresh require. Better than dofile here because it keeps normal require path behavior.

BreakBB: I am fine either way. I'll streamline other tests after we merge this, so they also don't use `dofile`. But not important for this PR.

Modules/Libs/WrappedText.test.lua
Logonz: There should now be tests for prefix and fontSource - at least some.
No real test for GetFont() returning partial or nil values but fontSource path is covered.

Modules/Libs/WrappedText.test.lua
Logonz: I am definitely not a testing expert - so if you want it a certain way i can follow your lead.

BreakBB: I honestly didn't know about `package.loaded` before 🤷🏻 And I also don't like that `dofile` and `require` use a different path format. So I am happy to start using the **explicit** `package.loaded` and always use `require` instead of `dofile`.

anon1231823:
I guess add #5138 and #3177 to this for fixing

Logonz:
While #5138 is related to wrapping it is not fixed here as wrapping in the tracker (doesn't)works the same.
I think the solutions provided her could fix it, but i think it is outside the scope of this task.

PR #7606, opened 2026-06-17
Use pvp/event/repeatable color codes too in certain cases

Cabro:
Fixes color code for certain quests


Modules/Tooltips/MapIconTooltip.lua
```diff
@@ -332,7 +331,8 @@ function MapIconTooltip:Show()
                 end
             else
                 if (quest and shift and xpReward > 0) then
-                    tooltipRows:AddDoubleLine(questTitle, l10n("(") .. FormatLargeNumber(xpReward) .. xpString .. l10n(")"), 0.2, 1, 0.2, r, g, b);
+                    local rewardString = QuestieLib:PrintDifficultyColor(quest.level, l10n("(") .. FormatLargeNumber(xpReward) .. xpString .. l10n(")") .. " ", QuestieDB.IsRepeatable(questId), QuestieEvent.IsEventQuest(questId), QuestieDB.IsPvPQuest(questId))
```
Logonz: I would say lets try not to duplicate the code, it will be less "performant" but increase the readability if you move `local rewardsString = [...]` to where `r, g, b = QuestieLib:GetDifficultyColorPercent(quest.level);` was before and use the variable in both places.

Issue #5046, opened 2023-07-18
The icons for a quest are being reconstructed each time after any progress instead of just updating tooltips to reflect real values.


Cabro:
This has started to happen between https://discord.com/channels/263036731165638656/810171186000298044/1109893558317613069 and just recently.

This is very distracting. Especially if you have a quest that has icons all over the map.
Not only is this distracting, but it's causing micro-stutters. It is 100% from Questie, as the microstutters stop happening after quest is complete and there's no more icons on the map.

https://github.com/Questie/Questie/assets/1155017/516196f0-d141-44c6-a804-879ab7408b00

8.8.2 wotlk

Dyaxler:
@Cabro is there another quest that this repro's on? All the ones I've tried don't do this... the only character I'm attuned with for BWL has already completed this quest you're on. I really don't feel like soloing the BWL attunement on my level 72 Priest much less soloing BWL just to get to that step you're on. :D

Cabro:
https://www.wowhead.com/wotlk/quest=9641/irradiated-crystal-shards

Dyaxler:
I'm not experiencing the same "redraw" issue you are with Irradiated Crystal Shards. Before looting the quest item, I did have my map open.

There are checks in the code to prevent this from happening each time there is a Quest Update. _DetermineIconsToDraw() is the primary check function that handles this. There is a key in the "quest" table `AlreadySpawned` and it's built upon accepting a quest. Once built, it stays there during a play session and is rebuilt each time a player login or reloadsUI. If the `AlreadySpawned` key is missing for an objective, then it's added during a quest update. 

The only way icons would be redrawn like you experienced in your video is if the `AlreadySpawned` key is somehow getting wiped out or not created at all. It might be quest specific.

After all that I did find the bug.

```
[11:07:13] DevTools: value=QuestieDB:GetQuest(8301).sourceItemId
[11:07:13] [1]=20402

[11:07:16] DevTools: value=QuestieDB:GetQuest(9641).sourceItemId
[11:07:16] [1]=0
```

In QuestieQuest:UpdateQuest() I made a change to resolve one of your other bugs for zero objectives:
https://github.com/Questie/Questie/blob/8e56d0184bf8903909d531da09938b3d55dc2e36/Modules/Quest/QuestieQuest.lua#L569C1-L592C65

[...]

PR #7625, opened 2026-06-21
[fix] Fix a frame leak when untracked quests are re-populated to the map

BreakBB:
## Problem

When progressing an "untracked quest", Questie re-creates the objective icons, instead of just showing the existing ones. That results in a frame leak in `QuestieFramePool`.

## Issue references

Most likely related to #5046

## Proposed changes

The problem was that removing tooltips for untracked quests also marked the objective icons as "not present" without actually removing them. So Questie re-created the icons. Now `objective.AlreadySpawned` is not cleared when removing tooltips, so Questie can reuse these icons.

## Screenshots

Stacked frames for the same quest objectives (before):
image

### DIFF
--- a/Modules/Map/QuestieMap.lua
+++ b/Modules/Map/QuestieMap.lua
@@ -86,8 +86,16 @@ function QuestieMap:UnloadQuestFrames(questId, iconType)
     if QuestieMap.questIdFrames[questId] then
         if not iconType then
             for _, frame in pairs(QuestieMap:GetFramesForQuest(questId)) do
+                -- Capture this before Unload() because it clears frame.data.
+                local objective = frame.data and frame.data.ObjectiveData
+
                 frame:Unload();
+
+                if objective then
+                    objective.AlreadySpawned = {}
+                end
             end
+
             QuestieMap.questIdFrames[questId] = nil;
         else
             for name, frame in pairs(QuestieMap:GetFramesForQuest(questId)) do
--- a/Modules/Map/QuestieMap.test.lua
+++ b/Modules/Map/QuestieMap.test.lua
@@ -13,6 +13,37 @@ describe("QuestieMap", function()
     before_each(function()
         QuestieLib = require("Modules.Libs.QuestieLib")
         QuestieMap = require("Modules.Map.QuestieMap")
+        QuestieMap.questIdFrames = {}
+    end)
+
+    describe("UnloadQuestFrames", function()
+        it("should clear AlreadySpawned for objective frames on full unload", function()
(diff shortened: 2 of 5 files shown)


Logonz:
Yeah, I think your change is correct, but it moves where the responsibility needs to live.

Before this PR we had this flow:

```text
AvailableQuests.RemoveQuest
  -> QuestieMap:UnloadQuestFrames()
       -> actual frames are unloaded
  -> QuestieTooltips:RemoveQuest()
       -> AlreadySpawned was cleared
```

That was wrong for the untracked/FakeHide path, because those frames are still alive. Clearing `AlreadySpawned` there makes Questie forget about live frames, so repopulating can draw duplicates.

But it also meant real unload paths were accidentally getting their `AlreadySpawned` cleanup from tooltip removal.

After this PR:

```text
AvailableQuests.RemoveQuest
  -> QuestieMap:UnloadQuestFrames()
       -> actual frames are unloaded
  -> QuestieTooltips:RemoveQuest()
       -> AlreadySpawned is no longer cleared
```

So now the FakeHide/untracked case is fixed, but the true unload case lost its cleanup.

The problem is that `_DetermineIconsToDraw` uses `AlreadySpawned` as the redraw guard:

```lua
if not objective.AlreadySpawned[id] then
    -- draw icons
end
```

So if a real unload removes the frames but leaves `AlreadySpawned[id]`, Questie can later think the icons still exist and skip redrawing them.

That is especially relevant for `SpecialObjectives`, because some quest lifecycle paths reset `quest.Objectives`, but keep/reuse `quest.SpecialObjectives`. Example flow:

```text
accept quest
  -> draw special objective icons
  -> specialObjective.AlreadySpawned gets populated

abandon/complete/remove quest
  -> QuestieMap:UnloadQuestFrames()
       -> frames are gone
  -> tooltip cleanup no longer clears AlreadySpawned

retake/repopulate
  -> specialObjective is reused
  -> AlreadySpawned still has old entries
  -> _DetermineIconsToDraw skips drawing
```

So the fix I pushed is basically:

```text
Tooltip removal:
  -> only tooltip state

FakeHide/untracked:
  -> preserve AlreadySpawned

Real frame unload:
  -> clear AlreadySpawned for objective-backed frames
```

Concretely, `QuestieMap:UnloadQuestFrames(questId)` now captures `frame.data.ObjectiveData` before `frame:Unload()` clears `frame.data`, and then clears `objective.AlreadySpawned = {}` for those objective frames.

That keeps your original leak fix intact, but preserves the old cleanup behavior for actual unloads in the place that owns frame unloading instead of in the tooltip module.

PR #7628, opened 2026-06-22
[feature] Add party icon slider options

### LINKED ISSUE #7614: Party quest icons PR needs slider options for icons
### Description

We need sliders for icon limit (do we want them like the existing ones - distance between the icons AND limit on number of icons, or just follow your existing icons rule?)

### Questie version

11.29.5

### Game flavor

Era/HC

Sommos:
## Issue references

Fixes #7614

## Proposed changes

- ADD: party icon limit option
- ADD: party objective icon filter distance

## Screenshots

image


Modules/Network/QuestiePartyObjectives.lua
```diff
@@ -214,8 +213,9 @@ end
 -- Draw a single party quest's objectives (assumes it has already been cleared).
 ---@param questId number
 local function _DrawQuest(questId)
+    local maxPartyIcons = Questie.db.profile.partyIconLimit or 500
```
BreakBB: We don't want fallbacks for things, that should never fallback.

```suggestion
    local maxPartyIcons = Questie.db.profile.partyIconLimit
```

Modules/Options/AdvancedTab/QuestieOptionsAdvanced.lua
```diff
@@ -143,7 +145,49 @@ function QuestieOptions.tabs.advanced:Initialize()
...
+                desc = function() return l10n("Minimum distance between two party members' objective icons in the same zone.\n\nSet to 0 to show all icons. Higher values reduce icon clutter."); end,
+                width = 1.5,
+                disabled = function() return (not Questie.db.profile.showPartyQuestObjectives); end,
+                min = 0,
+                max = 5,
+                step = 1,
+                get = function(info) return QuestieOptions:GetProfileValue(info); end,
+                set = function(info, value)
+                    QuestieOptions:SetProfileValue(info, value)
+                    QuestieOptionsUtils:Delay(0.5, function() QuestiePartyObjectives:Update() end, l10n("Setting party objective filter distance to %s : Redrawing!", value))
```
BreakBB: I wonder if we should just re-word all these settings to something like

`Setting filter distance to %s : Redrawing!`

I don't think there is a real benefit of having a 100% matching chat message for each of these setting.

Modules/Options/QuestieOptionsDefaults.lua
```diff
@@ -5,6 +5,8 @@ function QuestieOptionsDefaults:Load()
     return {
         profile = {
             objectiveFilterDistance = 2,
+            partyObjectiveFilterDistance = 2,
+            partyIconLimit = 500,
```
BreakBB, line 9: These need an entry in the Migration file.

Modules/Options/AdvancedTab/QuestieOptionsAdvanced.lua
```diff
@@ -184,7 +184,7 @@ function QuestieOptions.tabs.advanced:Initialize()
                 get = function(info) return QuestieOptions:GetProfileValue(info); end,
                 set = function(info, value)
                     QuestieOptions:SetProfileValue(info, value)
-                    QuestieOptionsUtils:Delay(0.5, function() QuestiePartyObjectives:Update() end, l10n("Setting party objective filter distance to %s : Redrawing!", value))
+                    QuestieOptionsUtils:Delay(0.5, function() QuestiePartyObjectives:Update() end, l10n("Setting filter distance to %s : Redrawing!", value))
```
BreakBB, line 187: We don't have this translation yet. But I'd say we should introduce it and use it for the other settings in here as well. There are three different "filter distance" and "icon limit" sliders.

Sommos: [8e11969](https://github.com/Questie/Questie/pull/7628/commits/8e119693bbb4206a0e5422d5ae339c6057c23186)

PR #7639, opened 2026-06-23
Rename instanceIdToUiMapId to instanceIdToAreaId

BreakBB:
It always was areaId not uiMapId

## Issue references

Fixes #

## Proposed changes

-
-


Logonz (approved):

Modules/QuestiePlayer.lua
BreakBB, line 96:
```suggestion
---@return AreaId
function QuestiePlayer:GetCurrentZoneId()
```

PR #7640, opened 2026-06-24
[locale] Validate localization lookup loadstrings

Logonz:
## Summary
- Replace the old German Classic-only localization validator with exhaustive lookup loadstring validation across all lookup expansions and locales.
- Expand `Localization/l10n.test.lua` so behavior smoke tests cover all lookup expansions/locales.
- Leave the current malformed TBC French lookup unfixed intentionally, so CI can demonstrate the new validation failure before the data fix.

## Validation
- `busted cli/validators/localizationLookupValidator.test.lua`
- Full localization validation is expected to fail until the TBC French lookup is fixed.


cli/validate-localization.lua
```diff
@@ -1,95 +1,4 @@
...
-    Questie.db.global.debugEnabled = true
-    QuestieDBCompiler:Compile(function() end)
-
-    QuestieDB:Initialize()
-
-    assert(QuestieDB.GetQuest(2).name == "Klaue von Scharfkralle")
-    assert(QuestieDB:GetNPC(3).name == "Fleischfresser")
-    assert(QuestieDB:GetObject(31).name == "Alte Löwenstatue")
-    assert(QuestieDB:GetItem(159).name == "Erfrischendes Quellwasser")
-end
```
BreakBB, line 93: Why was this part changed so heavily and replaced with stubs?

Localization/l10n.test.lua
```diff
@@ -12,302 +12,196 @@ describe("l10n", function()
...
+            npc = expected.npc,
+            object = expected.object,
+            quest = expected.quest,
+        }
+    end
 
-        l10n.InitializeUILocale()
-        l10n:Initialize()
+    local function _ApplyOverrides(expected, overrides)
+        if overrides then
```
BreakBB: I hate "logic" in tests, because it basically means, you also need tests for the tests to make sure that logic is correct.

I do like the "parameterized tests" approach below, but I vote to remove at least one layer so we can get rid of these `if`-conditions.

Also tests code is "free code" as it's not shipped to the user and I prefer explicit tests, instead of implicit, hard to read/understand ones. That's why all tests where explicitly written, even with duplication.

cli/validators/localizationLookupValidator.lua
```diff
@@ -0,0 +1,299 @@
...
+local function _GetSortedKeys(entries)
+    local keys = {}
+    for key in pairs(entries) do
+        table.insert(keys, key)
+    end
+    table.sort(keys)
+    return keys
+end
+
+local function _ExtractLuaChunkLine(reason)
```
BreakBB: It seems the output of the chunk line is not entirely correct:

Image

The commit happened in line 4438, not 4432.
https://github.com/Questie/Questie/pull/7640/changes/03387cdba59f293474c21897882c40e69fb36340#diff-bbffc76c79b675f126f3997f80cc6de666e4c07db3d328e2c1729e29b62a148aR4438

Might be that line only references the content of `loadstring`?

cli/validators/localizationLookupValidator.lua
```diff
@@ -0,0 +1,299 @@
...
+    if os.getenv("GITHUB_ACTIONS") ~= "true" then
+        return
+    end
+
+    print(
+        "::error file=" .. _EscapeGitHubAnnotationValue(filePath) ..
+        ",title=" .. _EscapeGitHubAnnotationValue(title) .. "::" ..
+        _EscapeGitHubAnnotationValue(reason)
+    )
+end
```
BreakBB: Might be worth to move these functions to a re-usable file, but that can also be done later on, once we need it at a different place 🤷🏻

cli/validators/localizationLookupValidator.lua
```diff
@@ -0,0 +1,299 @@
...
+
+        if type(l10n.questLookupOverrides) ~= "function" then
+            error(filePath .. " (" .. locale .. "): did not register l10n.questLookupOverrides as a function", 0)
+        end
+
+        _AssertLookupFunctionReturnsTable(filePath .. " (" .. locale .. ")", "l10n.questLookupOverrides", l10n.questLookupOverrides)
+    end)
+end
+
+---@param options? LocalizationLookupValidatorOptions
```
BreakBB: The `options` seem only relevant for the tests, I am not sure if we want to keep them in that case? I guess we need them to make this validator testable, but I'd give it a better name then tbh.

cli/validators/localizationLookupValidator.test.lua
```diff
@@ -0,0 +1,228 @@
...
+end
+
+local function _CreateLookupFixture(content, overridesContent)
+    _Mkdir("cli/output")
+    _Mkdir(fixtureRoot)
+    _Mkdir(fixtureRoot .. "/Classic")
+    _Mkdir(fixtureRoot .. "/Classic/lookupItems")
+
+    _WriteFile(fixtureRoot .. "/Classic/lookupItems/deDE.lua", content)
+    _WriteFile(fixtureRoot .. "/lookupOverrides.lua", overridesContent or [=[
```
BreakBB: Do we really need this `or` (easy to miss "logic")?

cli/validators/localizationLookupValidator.test.lua
```diff
@@ -0,0 +1,228 @@
...
+    end)
+
+    it("should validate generated lookup files and lookup overrides", function()
+        _CreateLookupFixture([=[
+if GetLocale() ~= "deDE" then return end
+local l10n = QuestieLoader:ImportModule("l10n")
+l10n.itemLookup["deDE"] = loadstring([[return {
+    [1] = "Test Item",
+}]])
+]=])
```
BreakBB: We should be able to use proper indentation for these shouldn't we?

cli/validators/localizationLookupValidator.test.lua
```diff
@@ -0,0 +1,228 @@
...
+    lfs.mkdir(path)
+end
+
+local function _WriteFile(path, content)
+    local file = assert(io.open(path, "w"))
+    file:write(content)
+    file:close()
+end
+
+local function _CreateLookupFixture(content, overridesContent)
```
BreakBB: The parameters of this function really scare me 🙈

I think that is a sign of a bad interface of `LocalizationLookupValidator.Validate`.

Might be better to move the "file loading" out of this validator, so it only receives already loaded data and the data loading stays as part of `validate-localization.lua`. That would also follow the structure of the other validators, which receive the DB data + DB keys.

BreakBB: It might even make no sense to keep `validate-localization.lua` but instead move the calls to the expansion specific `validate-` files.

anon1231823:
Spanish is also affected by this tracker issue that TBC french is having

cli/validators/localizationLookupValidator.lua
Logonz: Yeah it is within the loadstring - i think its "fine" it at least gives an indication of what is wrong and 99% of the time a person knows what they changed.

BreakBB: Yeah, my thinking as well. I was also thinking if it's save to do some `- 6` math 🤔

Localization/l10n.test.lua
Logonz: The problem with non-logic tests is that they did not cover every expansion and every locale.
We could expand the leaf into all of them but it must be remembered to add those things to the test.

Which i am fine with it only happens every expansion / locale add (which wont happen again  probably)

Logonz:
> Spanish is also affected by this tracker issue that TBC french is having

What do you mean?
From what i can tell there is no "tracker issue"
The problem comes from not having Automatic as language and switching between them.
We do not force a rebuild for example if you are on a English client with english and then switch over to a french client.
If you then switch to french (without the fix in the localization swap PR) you do not rebuild the database.
Thus you get a mix of titles being as the compiled database - but things like objectives and zones being localized from the client (and by extension the api)

Basically - all strings we internally keep and compile will stay as the previous language unless recompiled.
GameObject tooltips etc should also be broken in this state for example.

Localization/l10n.test.lua
BreakBB: Yeah, I am 100% for it with adding a test for every `expansion x language`. We really need to make sure translation problems are found in the CI. In the lookup files is way too much room for human error.

Localization/l10n.test.lua
```diff
@@ -12,302 +12,197 @@ describe("l10n", function()
...
+        assert.are.same({"Bringt die Lieferung zu Zinge."}, QuestieDB.questData[1359][QuestieDB.questKeys.objectivesText])
     end)
 
-    it("should load zhCN lookups", function()
-        _G.GetLocale = function() return "zhCN" end
-        dofile("Localization/lookups/Classic/lookupItems/zhCN.lua")
-        dofile("Localization/lookups/Classic/lookupNpcs/zhCN.lua")
-        dofile("Localization/lookups/Classic/lookupObjects/zhCN.lua")
-        dofile("Localization/lookups/Classic/lookupQuests/zhCN.lua")
+    it("should store quest objective strings as one-element tables", function()
```
BreakBB: From what I understand there shouldn't be any `questObjective` as string anymore, so do we want to have tests for it suggesting this still is a valid case?

Localization/lookups/Cata/lookupLoadstrings.test.lua
```diff
@@ -0,0 +1,488 @@
...
+    local originalQuestieLoader
+    local currentLookupFile
+    local originalLoadstring
+
+    local pairs = pairs
+    local type = type
+    local tostring = tostring
+    local error = error
+
+    local function _LoadRegisteredLookup(lookupTable, locale, filePath)
```
BreakBB, line 14: We duplicated these load and assertion across all expansion test files. Do you think we should extract them for reuse?

Localization/lookups/Cata/lookupLoadstrings.test.lua
```diff
@@ -0,0 +1,488 @@
...
+            --     end
+            --     for index, line in pairs(data[3]) do
+            --         if type(index) ~= "number" then
+            --             error(filePath .. " has a non-number quest objective line key for ID " .. tostring(id) .. ": " .. tostring(index), 0)
+            --         end
+            --         if type(line) ~= "string" then
+            --             error(filePath .. " has invalid quest objective line type for ID " .. tostring(id) .. ": " .. type(line), 0)
+            --         end
+            --     end
+            -- end
```
BreakBB: Since we removed the Description, we don't need this

Localization/lookups/Cata/lookupLoadstrings.test.lua
```diff
@@ -0,0 +1,488 @@
...
+            itemLookup = {},
+            npcNameLookup = {},
+            objectLookup = {},
+            questLookup = {},
+        }
+        originalGetLocale = _G.GetLocale
+        originalQuestieLoader = _G.QuestieLoader
+        originalLoadstring = _G.loadstring
+
+        _G.loadstring = function(source, chunkName)
```
BreakBB, line 128: I guess this might also be worth extracting.

BreakBB (commented):
I LOVE this new layout of these tests! This is exactly what I thought we would need 😍

Localization/lookups/Cata/lookupLoadstrings.test.lua
Logonz: The duplicated functions I agree with you I just don't think there is a nice place to place them!
All our tests are self contained and moving them out would have us create a test folder or something for a single helper test file.
I want to move them out but I honestly think it is both uglier and more confusing so I settled at just duplicating because honestly how often do we actually change this. This is one of those senior engineer moments I think, do not complicate things for the gain of nothing in practice.

BreakBB (changes requested):
We would want collect all the errors and print them at the end of each test. Which is also what the CLI validators are doing. Otherwise multiple errors can only be found in multiple runs.

BreakBB (approved):

PR #7643, opened 2026-06-24
Quest fixes and ZG uimapid fix

Cabro:


Database/Zones/data/uiMapIdToAreaId.lua
```diff
@@ -172,6 +172,7 @@ ZoneDB.private.uiMapIdToAreaId = [[return {
     [1456] = 1638, -- Thunder Bluff
     [1457] = 1657, -- Darnassus
     [233]  = 1977, -- Zul'Gurub - Raid - old ZG
+    [337]  = 1977, -- Zul'Gurub - Raid - new ZG
```
Logonz: Put in something like "new ZG after <expansion>" if it is In Mop or Cata - you will thank yourself in the future.

Modules/Auto/DisallowedIDs.lua
```diff
@@ -40,7 +40,8 @@ _AutoQuesting.disallowedNPCs = {
     [12944] = true, -- Lokhtos Darkbargainer (Thorium Brotherhood, Blackrock Depths)
     [14567] = true, -- Derotain Mudsipper (Blacksmith recipes, Tanaris)
     [14828] = true, -- Gelvas Grimegate (Darkmoon Faire Ticket Redemption)
-    [14921] = true, -- Rin'wosho the Trader (Zul'Gurub Isle, Stranglethorn Vale)
+    [14921] = true, -- Rin'wosho the Trader (Yojamba Isle, Stranglethorn Vale)
```
Logonz, line 43: Is this correct? Zul Gurub to Yojamba, is that true for all expansions so it is not MoP specific - if it is different between exp-s be clear about that.

Cabro: There is no Zul'Gurub Isle.

PR #7644, opened 2026-06-24
[fix] Prevent auto-accept filters from re-allowing quests

### LINKED ISSUE #1: Quest Not Tracked: Garments of The Light
The level 5 priest quest, Garments of the Light, is not tracked properly. It shows the quest objective as the quest giver as opposed to the guard at the lake to the east.

Quest Giver: Priestess Josetta

Objective NPC: Guard Roberts

Logonz:
### Part 1: `OnQuestDetail()` filter accumulation
`AutoQuesting.OnQuestDetail()` now preserves earlier auto-accept filter rejections when later filters are evaluated.

Previously, the repeatable/PvP checks reassigned `doAcceptQuest`, so a quest rejected by an earlier filter could be re-allowed by a later passing check. For example, a trivial quest could still be accepted when trivial auto-accept was disabled if subsequent repeatable/PvP checks passed.

The repeatable and PvP checks now accumulate with the existing decision using `and`, so once a quest is rejected, later filters cannot turn the decision back to accepted.

### Part 2: `OnGossipShow()` PvP filtering
`AutoQuesting.OnGossipShow()` now still honors the PvP auto-accept filter when trivial and repeatable quests are allowed.

Previously, gossip auto-accept short-circuited to the first available quest whenever trivial and repeatable quests were allowed. That bypassed the PvP filter, so with PvP auto-accept disabled, a PvP quest could still be selected.

The shortcut now only applies when trivial, repeatable, and PvP quests are all allowed. Otherwise, the existing filtering loop runs and can skip PvP quests as intended.

### Intent comments
Added concise comments near both filter blocks to document that all Auto Accept variants must be considered. This should make it clearer for future changes that new variants must not accidentally re-allow rejected quests or bypass filtering shortcuts.

## Tests

Added regression tests covering:
- `OnQuestDetail()`: trivial quests are not accepted even when repeatable/PvP checks would pass
- `OnQuestDetail()`: repeatable quests are not accepted even when the PvP check would pass
- `OnGossipShow()`: PvP quests are skipped even when trivial and repeatable settings are enabled

Validation run:
- `busted Modules/Auto/AutoQuesting.test.lua --filter "OnGossipShow"`
- `busted Modules/Auto/AutoQuesting.test.lua --filter "OnQuestDetail"`
- `busted Modules/Auto/AutoQuesting.test.lua`

Fixes #1

### DIFF
--- a/Modules/Auto/AutoQuesting.lua
+++ b/Modules/Auto/AutoQuesting.lua
@@ -102,7 +102,7 @@ function AutoQuesting.OnGossipShow()
         if #availableQuests > 0 then
             local indexToAccept = 0
 
-            if Questie.db.profile.autoAccept.trivial and Questie.db.profile.autoAccept.repeatable then
+            if Questie.db.profile.autoAccept.trivial and Questie.db.profile.autoAccept.repeatable and Questie.db.profile.autoAccept.pvp then
                 indexToAccept = 1
             else
                 for i = 1, #availableQuests do
--- a/Modules/Auto/AutoQuesting.lua
+++ b/Modules/Auto/AutoQuesting.lua
@@ -29,6 +29,7 @@ function AutoQuesting.OnQuestDetail()
         end
     end
 
+    -- Validate every disabled Auto Accept variant without letting later checks re-allow a rejected quest.
     local doAcceptQuest = true
     if (not Questie.db.profile.autoAccept.trivial) then
         local questLevel = QuestieDB.QueryQuestSingle(questId, "questLevel")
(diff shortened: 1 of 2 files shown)


BreakBB (approved):
Great finding!

Modules/Auto/AutoQuesting.lua
BreakBB, line 1: Not sure if a mistake or AI slot, but this change is not a `[fix]` as this is not relevant for the changelog 😅

Logonz: Ill see if we can rewrite the history on this one :D

BreakBB: I am sure you can :P

Issue #7642, opened 2026-06-24
Questie needs automatic locale select


Cabro:
### Description

Basically a dynamic default value which equals GetLocale()

### Questie version

11.30

### Game flavor

Era/HC

Logonz:
I checked this path, and Questie already has an automatic locale mode, but it is represented indirectly.

Current defaults are:

```lua
questieLocale = "enUS"
questieLocaleDiff = false
```

The important value is `questieLocaleDiff = false`. That means “Automatic”.

On startup:

- if `questieLocaleDiff == true`, Questie uses the saved `questieLocale`
- otherwise Questie uses `QUESTIE_LOCALES_OVERRIDE.locale` if present, else `GetLocale()`

So for a fresh/default setup:

- frFR client → effective locale is frFR
- enUS client → effective locale is enUS

Also, if the client locale changes before login while Questie is still in Automatic mode, the existing startup logic should detect that `dbCompiledLang` no longer matches the effective locale and trigger a recompile.

The confusing/risky part is mostly naming/state clarity:

- `questieLocale = "enUS"` is saved even when Auto is active
- the real mode is hidden behind `questieLocaleDiff`
- the dropdown says Automatic, but does not show the effective locale
- switching locale modes currently applies live, which ties into #7641

Long-term improvement proposal:

1. Keep Automatic as the default for fresh installs/new SavedVariables.
2. Make the UI clearer, e.g. show `Automatic (frFR)` when Auto resolves to frFR.
3. Separate selected mode from effective locale in code, even if only with helper functions at first:
   - selected mode: Auto vs explicit
   - effective locale: actual locale Questie will use now
4. Do not reset global explicit locale choices per character.
   - If the user explicitly picked enUS, a new character should not silently override that.
   - They can switch back to Automatic.
5. Pair this with #7641:
   - switching to/from Automatic should trigger confirm + reload/recompile only if the effective locale changes.

So I think #7642 is less about adding new per-character behavior and more about making the existing Auto behavior explicit, stable, and easier to understand.

Issue #7641, opened 2026-06-24
Questie locale selection


Cabro:
### Description

When switching locale in Questie settings we need:
1 - it needs to be followed by a recompile so the DBs are also changed to the new locale
2 - it needs a confirmation popup since it triggers a forced recompile
3 - translations for the confirmation popup

### Questie version

11.30

### Game flavor

Era/HC

Logonz:
Agreed. Locale changes should be treated as a DB-affecting setting, not as a live UI-only toggle.

I looked through the flow:

- `l10n:Initialize()` applies localized item/quest/NPC/object names to the raw DB.
- The DB compiler then stores that localized data in the compiled DB.
- Map icons/tracker data can also hold copied names/text from the compiled DB.
- The current locale dropdown only calls `l10n:SetUILocale(...)` and updates saved vars.
- It does not invalidate the compiled DB, reload UI, clear caches, or redraw existing icons.

So changing locale in settings can leave Questie in a mixed/stale state.

Proposed fix:

1. Change the locale dropdown so selecting a different **effective locale** opens a confirmation popup.
2. Do not apply the setting until the user confirms.
3. On confirm:
   - save the selected locale mode/value
   - call `l10n:SetUILocale(...)`
   - mark the active compiled DB as uncompiled
     - SoD: `Questie.db.global.sod.dbIsCompiled = false`
     - non-SoD: `Questie.db.global.dbIsCompiled = false`
   - call `ReloadUI()`
4. On next startup, the existing init path will recompile because the DB is uncompiled / compiled language differs.
5. Add the popup string to `Localization/Translations/Options/Advanced.lua`.

Implementation area:
- `Modules/Options/AdvancedTab/QuestieOptionsAdvanced.lua`
  - locale dropdown setter
  - add a `StaticPopupDialogs["QUESTIE_LOCALE_CHANGE_CONFIRM"]`
  - reuse the existing recompile/reset popup pattern
- `Localization/Translations/Options/Advanced.lua`
  - add confirmation text

One nuance:
- `Automatic` on a frFR client and explicit `frFR` both resolve to effective locale `frFR`.
- Switching between those does not require DB recompile, though we may still want to save the mode.
- Recompile/reload is only required when the effective locale actually changes.

PR #7645, opened 2026-06-24
[fix] Confirm locale changes before reload

Logonz:
## Summary
- Treat the locale dropdown selection as a selected mode/value (`Automatic` or explicit locale) separately from the effective locale Questie will use.
- Show `Automatic (<locale>)` so Automatic mode clearly reflects the current client/override locale.
- When a dropdown change would alter the effective locale, prompt before applying it; on confirmation, save the selection, invalidate the active compiled database, and reload the UI.
- When switching between Automatic and explicit mode without changing the effective locale, save the mode without a recompile/reload.
- Add the new confirmation popup string to Advanced options translations with non-English locales marked for human translation.

Addresses #7641 by making runtime locale changes use the safe recompile/reload flow.
Addresses #7642 by keeping fresh/default settings in Automatic mode and clarifying which locale Automatic resolves to.

## Validation
- `luac -p Modules/Options/AdvancedTab/QuestieOptionsAdvanced.lua Localization/Translations/Options/Advanced.lua`: passed
- `git diff --check`: passed
- targeted Python check for the new translation entry and supported locale keys: passed
- `lua cli/validate-localization.lua`: could not run in this environment because Lua module `lfs` is missing
- `busted`: not available in this environment
- `luacheck`: not available in this environment


Localization/Translations/Options/Advanced.lua
```diff
@@ -422,6 +422,18 @@ local advancedOptionsLocales = {
         ["zhCN"] = "自动选择",
         ["zhTW"] = "自動",
     },
+    ["Changing Questie's locale requires the database to be recompiled. The UI will reload and Questie may take a moment to process data after reload. Continue?"] = {
+        ["enUS"] = true,
+        ["deDE"] = "Das Ändern der Questie-Sprache erfordert eine Neukompilierung der Datenbank. Die Benutzeroberfläche wird neu geladen und Questie benötigt nach dem Neuladen möglicherweise einen Moment, um die Daten zu verarbeiten. Fortfahren?", -- GPT-5.5 machine translated using other translations as context.
```
BreakBB: I am not a fan of using machine translations, since we have a strong translation team which is closer to 100% correctness than every machine. Adding these will make it harder for our translators to find them (even though you at least added the comments, stating AI translations).

I do admit though, that the German translation is fine.

Modules/Options/AdvancedTab/QuestieOptionsAdvanced.lua
```diff
@@ -22,7 +22,29 @@ local QuestieJourney = QuestieLoader:ImportModule("QuestieJourney")
...
+        return l10n:GetFallbackLocale(QUESTIE_LOCALES_OVERRIDE.locale)
+    end
+
+    return l10n:GetFallbackLocale(GetLocale())
+end
+
+---@return nil
+local function _InvalidateCompiledDatabase()
+    if Questie.IsSoD then
+        Questie.db.global.sod = Questie.db.global.sod or {}
```
BreakBB: I don't think we need this, as didn't need it before.

Localization/l10n.lua
BreakBB: I just noticed we don't need this anymore, as we added `esMX` to `supportedLocals` at some point.


```suggestion
```

Localization/l10n.lua
```diff
@@ -104,6 +114,7 @@ function l10n.InitializeUILocale()
 end
 
 ---Load the locale override from QUESTIE_LOCALES_OVERRIDE (provided by a different addon)
+---@return nil
```
BreakBB, line 117: Do we really want to start adding these?

PR #7648, opened 2026-06-25
[fix] Split data prep and drawing party icons

### LINKED ISSUE #7617: Split data preparation and drawing for party icons
> I wonder if we should try and split the file apart a bit.
>
> Have one be a generator that is basically Get data, what should be drawn and so on.
And on that is a consumer that actually does a lot of the calls against the rest of the addon.
>
> Create a nice data format from this file which is consumed - transformation would then happen outside of this file.
> This could really help with actually creating real tests around this code.
>
> Do the Modules/Network/QuestiePartyObjectives really need to know about Unload and PopulateObjectives and so on.
Split it up by the networking layer and map layer - move the map code closer to the actual map handlers.

Related to https://github.com/Questie/Questie/pull/7574#discussion_r3428031702

Sommos:
## Issue references

Fixes #7617

## Proposed changes

- FIX: split data prep and drawing party icons
- ADD: additional test cases for party icons


Questie-Mists.toc
```diff
@@ -222,6 +222,7 @@ Modules\Map\QuestieMapUtils.lua
 Modules\Map\HBDHooks.lua
 Modules\Map\WeaponMasterSkills.lua
 Modules\Map\Moonwell.lua
+Modules\Map\PartyObjectiveDrawer.lua
```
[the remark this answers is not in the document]
Sommos: This is wrong - If a module is not loaded yet, `ImportModule` creates a placeholder table, when `QuestieQuest` runs `CreateModule("QuestieQuest") it returns the same placeholder table and populates it in place.

Modules/Network/PartyObjectivesProvider.lua
```diff
@@ -0,0 +1,243 @@
...
+local function _PlayerHasQuest(questId)
+    return QuestLogCache.questLog_DO_NOT_MODIFY[questId] ~= nil
+end
+
+---@param name string
+---@return boolean @true if the named party member is online
+local function _IsPlayerOnline(name)
+    -- UnitIsConnected accepts a group member's name (same pattern as QuestieComms:CheckInGroup,
+    -- which calls UnitInParty/UnitInRaid by name). Returns falsy for offline or non-group names.
+    return UnitIsConnected(name)
```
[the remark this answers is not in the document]
Sommos: This is wrong - It follows existing patterns, and calls the correct API.

Modules/Map/PartyObjectiveDrawer.lua
```diff
@@ -0,0 +1,236 @@
...
+
+        local objective = _BuildObjective(descriptor)
+        -- Reuse the DB-built spawn list read-only; PopulateObjective builds it from
+        -- Type/Id when absent (the required-source-item case).
+        objective.spawnList = descriptor.spawnList or {}
+
+        local ok = pcall(QuestieQuest.PopulateObjective, QuestieQuest, quest, descriptor.Index, objective, true)
+        if ok then
+            drawn[#drawn + 1] = objective
+            iconCount = iconCount + _CountIcons(objective)
```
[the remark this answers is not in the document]
Sommos: [4f15a4f](https://github.com/Questie/Questie/pull/7648/commits/4f15a4f03b28e88802beed655becaba751776b97)

Modules/Map/PartyObjectiveDrawer.lua
```diff
@@ -0,0 +1,236 @@
...
+            state = { attempts = 0, pending = false }
+            prefetchedQuests[questId] = state
+        end
+        if (not state.pending) and state.attempts < MAX_PREFETCH_RETRIES then
+            state.pending = true
+            C_Timer.After(1.5, function()
+                state.pending = false
+                state.attempts = state.attempts + 1
+                QuestiePartyObjectives:ScheduleUpdate(questId)
+            end)
```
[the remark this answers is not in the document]
Sommos: [4f15a4f](https://github.com/Questie/Questie/pull/7648/commits/4f15a4f03b28e88802beed655becaba751776b97)

Modules/Network/PartyObjectivesProvider.lua
BreakBB, line 1: For me personally `Generator` sounds like creating data out of thin air. Maybe something like `PartyObjectivesProvider` is a better name 🤔

Modules/EventHandler/GroupEventHandler.test.lua
```diff
@@ -0,0 +1,73 @@
...
+    -- seeding call itself is not counted.
+    local function seed(members, remoteQuestLogs)
+        _G.GetNumGroupMembers = function() return members end
+        QuestieComms.remoteQuestLogs = remoteQuestLogs or {}
+        GroupEventHandler.GroupRosterUpdate()
+        QuestiePartyObjectives.ScheduleUpdate = spy.new(function() end)
+    end
+
+    before_each(function()
+        QuestiePlayer = QuestieLoader:CreateModule("QuestiePlayer")
```
BreakBB: We usually don't reference `QuestieLoader` in tests, but instead use `require` to load modules.

That way you don't get empty modules from `QuestieLoader`, _can_ still mock certain things, but also have the option to not mock certain things in adjacent modules.

Modules/Map/PartyObjectiveDrawer.lua
```diff
@@ -0,0 +1,243 @@
+---@class PartyObjectiveDrawer : QuestieModule
+local PartyObjectiveDrawer = QuestieLoader:CreateModule("PartyObjectiveDrawer")
+-------------------------
+--Import modules.
+-------------------------
+---@type QuestiePartyObjectives
+local QuestiePartyObjectives = QuestieLoader:ImportModule("QuestiePartyObjectives")
```
BreakBB: This is a cyclic import. So something is breaking the chain of command imo.

I'd say `QuestiePartyObjectives` is is allowed to call `PartyObjectiveDrawer` and `PartyObjectivesGenerator`, but both modules are not allowed to call `QuestiePartyObjectives` and also not each other.

So either a callback needs to be handed over or the data needs to be **completely** be provided by `PartyObjectivesGenerator`. I'd prefer the latter as fetching information is not something a drawer should take care of.

Modules/Network/PartyObjectivesGenerator.test.lua
```diff
@@ -0,0 +1,199 @@
...
+        QuestLogCache.questLog_DO_NOT_MODIFY = {}
+
+        _G.GetNumGroupMembers = function() return 3 end
+        _G.UnitIsConnected = function() return true end
+
+        PartyObjectivesGenerator = require("Modules.Network.PartyObjectivesGenerator")
+    end)
+
+    describe("ShouldDraw", function()
+        it("returns true when the setting is on and in a small group", function()
```
BreakBB: Absolute nitpick: I am a fan of letting test names start with `should` 🤷🏻

Modules/Network/PartyObjectivesGenerator.test.lua
```diff
@@ -0,0 +1,199 @@
...
+            assert.is_true(PartyObjectivesGenerator.ShouldDraw())
+        end)
+
+        it("returns false when the setting is off", function()
+            Questie.db.profile.showPartyQuestObjectives = false
+            assert.is_falsy(PartyObjectivesGenerator.ShouldDraw())
```
BreakBB: The test name states `false`, but the assertion tests `falsy`, meaning `nil` would also be fine. I think we should use `is_false` to make sure it actually is a boolean.

BreakBB (commented):
Great refactoring, I fully support this architecture switch and I think that is something other places can profit from as well!

I noted some minor findings.

Modules/Network/PartyObjectivesProvider.lua
Sommos: [4e87e80](https://github.com/Questie/Questie/pull/7648/commits/4e87e807e140d1721defd5649aa15b44789f0837)

Modules/EventHandler/GroupEventHandler.test.lua
Sommos: Yeah, this is a common pattern that I just missed. 

[4e87e80](https://github.com/Questie/Questie/pull/7648/commits/4e87e807e140d1721defd5649aa15b44789f0837)

Modules/Map/PartyObjectiveDrawer.lua
Sommos: [4e87e80](https://github.com/Questie/Questie/pull/7648/commits/4e87e807e140d1721defd5649aa15b44789f0837) fixes the cyclic import, good catch!

Modules/Network/PartyObjectivesGenerator.test.lua
Sommos: No worries, should have seen that pattern in our other tests, but just missed it.

[4e87e80](https://github.com/Questie/Questie/pull/7648/commits/4e87e807e140d1721defd5649aa15b44789f0837)

Modules/Network/PartyObjectivesGenerator.test.lua
Sommos: [4e87e80](https://github.com/Questie/Questie/pull/7648/commits/4e87e807e140d1721defd5649aa15b44789f0837)

PR #7652, opened 2026-06-25
Fix comms tooltip registration after skipped item objectives

Logonz:
## Summary
Previously, `QuestieCommsData:RegisterTooltip` treated missing or hidden item DB data as control flow for the whole objective list. That meant one item objective could stop registration for its direct item tooltip and for later valid objectives. Now source/provider expansion is gated by usable item DB data, while direct item tooltip registration remains keyed by the received item objective and the loop continues registering later objectives.

## Before
```text
received quest objective list
  -> RegisterTooltip loops objectives
  -> missing/hidden item DB data
  -> return from whole RegisterTooltip call
  -> no direct i_<itemId> tooltip
  -> later valid objectives never register tooltip keys
```

## After
```text
received quest objective list
  -> RegisterTooltip loops objectives
  -> missing/hidden item DB data skips source/provider expansion
  -> direct i_<itemId> tooltip still registers for item hover
  -> later valid objectives continue registering tooltip keys
```

## Example
If a party member sends quest progress where objective 1 is a hidden or locally missing item and objective 2 is “kill this monster,” the old code stopped at objective 1. Hovering the item or the monster would not show that party member’s progress. The new code avoids expanding the item onto unknown or hidden source entities, but still registers the direct item tooltip and continues to register the monster tooltip.

## Responsibility / ownership
Usable item DB data owns source/provider expansion because those tooltip keys come from `item.Sources`. The received item objective owns the direct `i_<itemId>` tooltip because that key comes from the objective itself. A missing or hidden item should not abort the rest of the objective list.

## Validation
- `busted -p ".test.lua" --filter "RegisterTooltip" Modules/Network/QuestieCommsData.test.lua`: `2 successes / 0 failures / 0 errors`
- `busted -p ".test.lua" Modules/Network`: `19 successes / 0 failures / 0 errors`

### DIFF
--- a/Modules/Network/QuestieCommsData.test.lua
+++ b/Modules/Network/QuestieCommsData.test.lua
@@ -7,13 +7,22 @@ describe("QuestieCommsData", function()
     ---@type QuestieDB
     local QuestieDB
 
-    before_each(function()
-        QuestieComms = require("Modules.Network.QuestieComms")
+    local originalGetItem
 
-        QuestieDB = require("Database.QuestieDB")
+    before_each(function()
+        QuestieComms = QuestieLoader:ImportModule("QuestieComms")
+        QuestieComms.data = {}
+
+        QuestieDB = QuestieLoader:ImportModule("QuestieDB")
+        originalGetItem = QuestieDB.GetItem
         QuestieDB.GetItem = function() return nil end
 
-        require("Modules.Network.QuestieCommsData")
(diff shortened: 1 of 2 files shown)


Modules/Network/QuestieCommsData.test.lua
BreakBB, line 1: Your agent really loves logic and conditions in unit tests 😅 

I'll clean this one up.

BreakBB (approved):

PR #7661, opened 2026-06-27
Merge journey quest details

### LINKED ISSUE #7391: All Zones dropdown suggestion: Don't show Hidden quests

BreakBB:
## Issue references

FIxes #7391
Closes #7553

## Proposed changes

The quest details in QBZ/QBF and Advanced Search are pretty similar atm, but not exactly the same. This PR changes that and merges the information of both into a single one, choosing be best of both worlds.

### Before

image
image
image

### After
image
image
image


anon1231823:
Tag that this fixes #7391

PR #7584, opened 2026-06-10
[feature] Add profession requirements to advanced search

Sommos:
## Issue references

Fixes #7580

## Proposed changes

- ADD: profession requirements to advanced search

## Screenshots

image
image

PR #7664, opened 2026-06-28
[feature] Show profession requirements in Journey

### LINKED ISSUE #7580: [Feature] Show required profession in Advanced Search
### Description

Like we have Required Level and Required Class, maybe we ought to have Required Profession and only show that line if the quest has a profession requirement

Image

BreakBB:
## Issue references

Fixes #7580
Closes #7584


Modules/Journey/QuestDetailsFrame.lua
```diff
@@ -350,6 +352,52 @@ function QuestDetailsFrame:Draw(container, quest)
...
+    -- Required Profession
+    local requiredSkill = QuestieDB.QueryQuestSingle(quest.Id, "requiredSkill")
+    local requiredSpecialization = QuestieDB.QueryQuestSingle(quest.Id, "requiredSpecialization")
+    local requiredRanks = QuestieDB.QueryQuestSingle(quest.Id, "requiredRanks")
+    local professionParts = {}
+
+    if requiredSkill and requiredSkill[1] and QuestieProfessions:GetProfessionName(requiredSkill[1]) then
+        local specializationName = requiredSpecialization and QuestieProfessions:GetSpecializationName(requiredSpecialization)
+        local entry = specializationName and l10n(specializationName) or l10n(QuestieProfessions:GetProfessionName(requiredSkill[1]))
+        if requiredSkill[2] and requiredSkill[2] > 1 then
```
Cabro:
```suggestion
        if requiredSkill[2] and requiredSkill[2] >= 1 then
```

It would be great to also see when you need skill 1, or it feels very weird otherwise (at least personally)

PR #7668, opened 2026-06-29
Comms and Visibility Proposal

Logonz:
## Summary
Previously, party objective visibility was implemented by filtering `QuestieComms.remoteQuestLogs` through tracker state. That made the comms quest log stop meaning “this is the remote player’s current quest log/progress” and start meaning “this is the subset of quests currently shown/tracked.” This PR restores the ownership boundary: `QuestieComms` owns factual quest-log/progress state, while the new modern comms path advertises prefix support and sends separate per-quest visibility intent.

## Before
```text
local tracker / hidden state
  -> filters QuestieComms quest-log packets
  -> remoteQuestLogs loses quests the player still has
  -> receivers cannot tell "quest absent" from "quest hidden/untracked"
```

That caused protocol ambiguity:

```text
remoteQuestLogs[questId][playerName] missing
  could mean:
    - player does not have the quest
    - player has it but untracked it
    - player hid it
    - state is stale after a tracking option change
```

## After
```text
QuestieComms
  -> syncs absolute quest-log/progress state

CommsHello / QuestieH1
  -> advertises which comm prefixes this client listens to

CommsVisibility / QuestieV1
  -> syncs per-quest show/hide intent separately

QuestiePartyObjectives + tooltips
  -> read quest progress from remoteQuestLogs
  -> read display permission from remoteQuestVisibility
```

Visibility payload is intentionally simple:

```lua
{
    [questId] = true,  -- show this quest's party objectives
    [questId] = false, -- suppress this quest's party objectives
}
```

Missing visibility data defaults to `true` for old clients / unknown peers.

## Example
Bob has quest `1234`, but Bob untracks it.

Before this proposal:

```text
Bob untracks quest 1234
  -> QuestieComms stops sending quest 1234
  -> Alice's remoteQuestLogs no longer has Bob/1234
  -> Alice cannot know whether Bob abandoned the quest or merely hid it
```

After this proposal:

```text
Bob still sends quest 1234 progress through QuestieComms
Bob also sends QuestieV1 visibility:
  [1234] = false

Alice stores:
  remoteQuestLogs[1234]["Bob"] = objective progress
  remoteQuestVisibility["Bob"][1234] = false

Party objective rendering:
  progress exists, but visibility says do not draw it
```

## Responsibility / ownership
`remoteQuestLogs` is the wrong owner for tracker/visibility state because it represents quest ownership and objective progress. Removing a quest from that table should mean the remote player no longer has or shares that quest-log entry, not that their UI currently hides it.

The new ownership split is:

```text
QuestieComms
  owns remote quest presence/progress

CommsVisibility
  owns remote display intent

QuestiePartyObjectives / tooltips
  combine both when deciding what to show
```

## Protocol shape
This starts the modern comms family without replacing the legacy quest-log protocol yet.

```text
QuestieH1 = hello / prefix-state advertisement
QuestieV1 = full quest visibility snapshot
```

`QuestieH1` is intentionally dumb:

```lua
{
    QuestieH1 = true,
    QuestieV1 = true,
    questie = true,
    Questie = true,
    REPUTABLE = true,
}
```

Semantics:

```text
true  = this client is listening on this prefix
false = this client knows the prefix contract but is not listening/supporting it
nil   = unknown / old client
```

The hello payload does not define what a prefix means. Local code does. Unknown remote prefixes are ignored and never dynamically registered.

## Serialization
Modern prefixes use:

```text
Lua table
  -> C_EncodingUtil.SerializeCBOR
  -> C_EncodingUtil.CompressString(... Deflate ...)
  -> LibDeflate:EncodeForWoWAddonChannel
```

LibDeflate is used only for addon-channel-safe encoding, not compression.

We chose full visibility snapshots instead of incremental updates because a max quest log is small. Measured 25 high-ID quest visibility entries as a map at roughly ~94 bytes on the wire, comfortably below a single addon-message payload.

## Caveats / deferred work
- This does not implement full typed-prefix replacement for quest-log update/full-log/remove packets yet.
- Visibility snapshots are currently group-sent, not peer-selected through hello negotiation.
- No live in-game `QuestieV1` smoke test was run.
- WHISPER/group gating still depends on WoW APIs accepting the sender name shape AceComm provides.

## Validation
- `git diff --check`: passed
- `luacheck -q -- Database Localization Modules Public Questie.lua`: `0 warnings / 0 errors`
- Targeted Busted bundle: `73 successes`
- Full Busted suite: `busted -p ".test.lua" .` — `1126 successes`


docs/addon-channel-serialization-report.md
```diff
@@ -0,0 +1,405 @@
+# Addon-channel serialization and SAY/YELL transport report
+
+## Scope
+
+This report summarizes live Questie communication experiments run against a running Classic client. The goal was to compare QuestieComms packet shapes, Questie's current serializer, Blizzard's built-in `C_EncodingUtil` CBOR/compression APIs, and Classic-only `SAY`/`YELL` addon-channel byte behavior.
```
BreakBB, line 5: I don't think we should consider SAY/YELL at all at this point, since it is a delicate topic with big impact.

docs/addon-channel-serialization-report.md
```diff
@@ -0,0 +1,405 @@
...
+CPU caveat from the same test context:
+
+| Operation | Approx. microseconds per operation |
+| --- | ---: |
+| Native Deflate compress | 4.3 |
+| Native Deflate decompress | 0.9 |
+| LibDeflate dictionary compress | 200 |
+| LibDeflate dictionary decompress | 34 |
+
+Dictionary compression improves size, but it is much slower and adds protocol/dictionary-version coupling. Reusing and hardening the V2 packet shape is a cleaner first step than inventing a V1 dictionary layer.
```
BreakBB, line 170: No clue if the data is reliable or AI slop, but this is a HUGE difference in performance, while the actual size difference is not.

Also I am confused that AceSerializer doesn't seem to have been part of the comparison, while it is the defacto standard from what I know. And we already use it.

docs/questie-comms-design.md
```diff
@@ -0,0 +1,166 @@
...
+
+That table means: the remote player has this quest, and this is the objective progress Questie knows for that player.
+
+Absence from `remoteQuestLogs` must not be overloaded to mean "hidden", "untracked", or "not shown". A missing entry should mean the quest is actually absent, removed, unknown, or not yet synchronized. UI/display preferences are a different kind of state.
+
+Future hidden/tracked/shown behavior should live in a separate store and protocol. For example, a later module may keep remote visibility state keyed by player and quest, but that state must not create, remove, or corrupt `remoteQuestLogs` entries by itself.
+
+## Terms and semantics
+
+`QuestieH1` uses a boolean prefix-state model:
```
BreakBB, line 71: Do we really want to tri-state-boolean, instead of something readable?

docs/questie-comms-design.md
```diff
@@ -0,0 +1,166 @@
...
+## `QuestieH1` hello design
+
+`QuestieH1` is intentionally dumb. Its decoded payload is only a boolean map of Questie addon-message prefixes:
+
+```lua
+{
+    QuestieH1 = true,
+    questie = true,
+    Questie = true,
+    REPUTABLE = true,
```
BreakBB: We don't need REPUTABLE at all as it is dead and will most likely cause more harm than good. We should actually remove that part of the code right away.

Modules/Network/CommsEncoding.lua
```diff
@@ -0,0 +1,288 @@
+--[[
+CommsEncoding owns Questie's modern typed-prefix wire encoding:
+
+    Lua payload table
+        -> C_EncodingUtil.SerializeCBOR(payload)
+        -> C_EncodingUtil.CompressString(cbor, Enum.CompressionMethod.Deflate)
+        -> addon-channel-safe byte escaping
+
+The addon-channel-safe byte codec below is an altered, trimmed subset of LibDeflate
```
BreakBB: What exactly is the benefit of stripping some code out of Lib and make it our own, to maintain code? I personally feel like we should simply re-use existing external tools and not built a new `QuestieSerializer` mess.

So either full import of LibDeflate or `AceSerializer` as used in the `Comms` module or something similar.

Also why is `C_EncodingUtil` not enough?

Modules/Network/CommsHello.lua
```diff
@@ -0,0 +1,337 @@
...
+The meaning of those prefixes lives in local code, not in the hello payload.
+
+Known prefixes default to false, meaning this client knows the prefix contract but is not
+currently listening/parsing it. The modules that actually register and parse a prefix mark
+that prefix true after registering their own receiver, so disabled or sunset handlers naturally
+advertise false without needing to keep this file in sync with deleted parser code.
+]]
+---@alias QuestieCommsPrefixState table<string, boolean>
+
+---@class CommsHello : QuestieModule
```
BreakBB: I think we can find a better name than `CommsHello`. Personally hearing the name, I have no clue what it does. Sounds either like a Hello World example or some simple "I'll greet players" module.

Modules/Network/CommsPrefixRegistry.lua
```diff
@@ -0,0 +1,337 @@
...
+end
+
+---Schedules a jittered hello so group roster events do not make every client speak at once.
+---@param _reason string? Debug-only call-site label reserved for future logging.
+---@return nil
+function CommsHello:ScheduleHello(_reason)
+    -- Send with timer debounce.
+    _CancelHelloTimer()
+
+    helloTimer = C_Timer.NewTimer(math.random() * 2, function()
```
BreakBB, line 140: Is a max offset for messages between 0-2 seconds enough?

Modules/Tracker/QuestieTracker.test.lua
```diff
@@ -14,24 +14,111 @@ describe("QuestieTracker", function()
...
+        QuestiePlayer = QuestieLoader:ImportModule("QuestiePlayer")
+        QuestiePlayer.currentQuestlog = {}
+        QuestieDB = QuestieLoader:ImportModule("QuestieDB")
+        QuestieDB.GetQuest = spy.new(function() return {zoneOrSort = 12} end)
 
         dofile("Modules/Tracker/QuestieTracker.lua")
         QuestieTracker = QuestieLoader:ImportModule("QuestieTracker")
     end)
 
+    local function setFunctionUpvalue(fn, upvalueName, value)
```
BreakBB: Looks like the code is not testable, so we are hacking values into the QuestieTracker? 🫠 I think we should simply call `QuestieTracker.Initialize()` in the `before_each`.

Modules/Network/CommsVisibility.lua
```diff
@@ -0,0 +1,291 @@
...
+        end
+    end
+
+    return snapshot
+end
+
+-------------------------
+-- Initialization.
+-------------------------
+---@return boolean
```
BreakBB, line 81: This is not needed. The tests already check for spy calls.

BreakBB (commented):
Overall I am not against the format of these new modules. I have mixed feelings towards CommsHello though and I do think we want to reuse any existing serialization solution instead of building our own with a mix of things. It's not like we are sending huge data packets left and right anyways.

I did left some rough feedback, but didn't look into everything in detail. I especially skipped most tests as this review was meant as a first "I understand the structure you suggest".

docs/addon-channel-serialization-report.md
Logonz: This document is a full information document, there are some characters that are now allowed in SAY and YELL so it is just information - not that we should use it.

docs/addon-channel-serialization-report.md
Logonz: We don't use AceSerializer except in the "newly" introduced Comms.lua - the other comms use QuestieSerializer i think... So it was just me not knowing it had been introduced and started to be used when i did the comparison
LibDeflate dictionary mode is pure lua, not something we will want to use. LibDeflate Deflate is faster but not dictionary so this again is just data basically.

docs/questie-comms-design.md
Logonz: The simple reason is that it is very cheap to serialize and compress and it can then be used raw in checks and so on - the data is connected to the logic. Though i do concede that it can be seen as a bit complicated though the usage is relatively simple.
We could just have it be a list and the "tri-state" is generated from it instead but all suggestions here are welcome.

docs/questie-comms-design.md
Logonz: I agree, i just wanted to include it to be thorough

Modules/Network/CommsEncoding.lua
Logonz: The reason i stripped it out is because the file is almost 3500 lines of code but we can just add it back as a library (Though it should be said the library code has not changed in 6 years).

I do like using the LibDeflate variant though because we can convert any byte-range to a addon-safe message so we can do things like compression, which is not something we can do with AceSerializer.

C_EncodingUtil unforunately creates non-addon-message-safe data so it has to be converted to a safe format.

Modules/Network/CommsHello.lua
Logonz: I have no problem renaming it, come up with ideas
CommsPrefix maybe?

Modules/Network/CommsPrefixRegistry.lua
Logonz: Maybe doing 1 + (math.random() * 2) for a min 1 second might be better.

Modules/Network/CommsVisibility.lua
Logonz: The code might have moved, what do you mean?

BreakBB: That function returns a boolean which is only needed for tests. I do think that is bad design, especially since that boolean does not add anything to the rest of the test assertions already in place.

Modules/Network/CommsPrefixRegistry.lua
BreakBB: Well this was just some random question that came into my mind. Maybe as something we should keep an eye out for when testing things.

Modules/Network/CommsHello.lua
BreakBB: Maybe something like `CommsPrefixRegistry`?

Modules/Network/CommsEncoding.lua
BreakBB: > I do like using the LibDeflate variant though because we can convert any byte-range to a addon-safe message so we can do things like compression, which is not something we can do with AceSerializer.

That is something I didn't know (or understand from the code/comments). I am fine to not **fully** import LibDeflate either, it might be worth though to move it to Libs, so we still think of it like "this is third party, we don't touch this, it just works" 🤷🏻 I am honest that is something that would help me personally.

PR #7689, opened 2026-07-11
[fix] Auto abandon breadcrumb followup quest on initial login

anon1231823:
Auto-abandoning breadcrumb followup quests was not working when logging in for the first time and playing on an established character. You had to manually abandon before auto-abandon work would.


anon1231823:
Any issues?

Issue #7382, opened 2026-01-28
Issues with timed quests.


Cabro:
### Description

This is for quests without objectives:

1 - when timer ends, you are stuck with "1 second" in tracker
2 - no "failed" added to quest state in tracker

<img width="623" height="281" alt="Image" src="https://github.com/user-attachments/assets/be5e3842-8e41-4b4e-80e1-63d9679ae1db" />

<img width="691" height="302" alt="Image" src="https://github.com/user-attachments/assets/756b6ce7-32cf-4eb8-9a64-99357096131f" />

3 - you are being forced the blizzard text (that's not even the completion text) after quest turns complete, even if you have the "show blizzard completion text" unchecked

4 - after checking off and back on the "hide blizzard completion text" the failed quest shows as "complete"

<img width="1937" height="724" alt="Image" src="https://github.com/user-attachments/assets/99a7db3e-4c83-4b46-9b65-90a6a6cb2f71" />

5 - it's not putting the quest tooltip back on NPC (it used to do that previously) 

<img width="692" height="920" alt="Image" src="https://github.com/user-attachments/assets/dc172ed7-6ca2-4cfb-9b8a-01205144c0b7" />

6 - blizzard timer shows even when you dont want it to

https://github.com/user-attachments/assets/3b64ae5c-2327-43b0-896a-fe97a2b22c9c

### Questie version

11.18

### Game flavor

TBC

Cabro:
7 - blizzard timer shows up for a split second after you get a tracker update due to some quest objective progress - FIXED

BreakBB:
1. Not a problem anymore
2. Fixed
3. Fixed by #7698
4. Fixed by #7702
5. Fixed
6. & 7. Fixed by https://github.com/Questie/Questie/commit/884df338aad16d070f423929ed8b1536c6ef5210

PR #7698, opened 2026-07-17
[fix] Fix complete timed quests always showing Blizzard completion text

BreakBB:
Related to #7382

### DIFF
--- a/Modules/Tracker/QuestieTracker.lua
+++ b/Modules/Tracker/QuestieTracker.lua
@@ -749,7 +749,7 @@ function QuestieTracker:Update()
                     end
 
                     -- Set minimizable quest flag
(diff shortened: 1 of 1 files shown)


Cabro:
So it's semi-fixed. It stops showing the Blizzard completion text, but it's not showing what it should when you have the `Auto Minimize Completed Quests` checkbox enabled. It should not show the `Quest Complete!` on objective line, it should instead add ` (Complete)` to the quest title.
image

If possible, it would be great to still show the timer as it is for these cases, as there are many timed quests without objective that would be collapsed and thus hiding the timer when the option for `Auto Minimize Completed Quests` is enabled.

BreakBB:
Changes included in #7704

PR #7718, opened 2026-07-24
Init improvements

### LINKED ISSUE #7717: Questie error on initialization "script ran too long"
### Description

When the add on is initialized, I have an error : Error during initialization! AddOns/Questie/Modules/Quest/QuestFinisher.lua:164:script ran too long"

This bug is happening since the update of the game the 22-07-2026.
Issue #7711 that was supposed to make the bug away actually made it worse.

Before this update, when the error occurred, questie would not display anything but I could reload the interface it eventually worked and displayed no error.

Now that issue #7711 is shipped and that I have upgraded to version 11.33.2 it is actually worse because : 
1. I now always have the error, no matter how much reload I do
2. When I have the error, the quests I currently have are displayed on the map, but they are NOT actualized when I progess on quests. The only way to upgrade my quest progess on the map is to do a /reload. So everytime I finish a quest I must /reload to see the markers go away on the map.

UPDATE : I reopened the game later and the bug seems to be gone for now on. I don't know why, I didn't updated any addons nor did I rebooted my PC. I will update the issue again if the bug comes back.

### Questie version

v11.33.2

### Game flavor

Era/HC

BreakBB:
## Issue references

Fixes #7717

## Proposed changes

-
-

### DIFF
--- a/Database/compiler.lua
+++ b/Database/compiler.lua
@@ -913,31 +913,17 @@ function QuestieDBCompiler:CompileTableCoroutine(tbl, types, order, lookup, data
     local count = 0
     local indexLookup = {};
 
-    local max_id = 0
     local yieldCount = 0
     for id in pairs(tbl) do
-        if id > max_id then
-            max_id = id
-        end
+        count = count + 1
+        indexLookup[count] = id
         yieldCount = yieldCount + 1
         if yieldCount >= TICKS_PER_YIELD * 100 then
             yieldCount = 0
             coYield()
         end
     end
-    yieldCount = 0
-    -- iterate table tbl in numerical order to get ids in order to indexLoopup list. iterating over pairs(tbl) gives ids in non determined order
-    for id=0,max_id do
-        if tbl[id] then
-            count = count + 1
-            indexLookup[count] = id
-        end
-        yieldCount = yieldCount + 1
-        if yieldCount >= TICKS_PER_YIELD * 100 then
-            yieldCount = 0
-            coYield()
-        end
-    end
+    table.sort(indexLookup)
     count = count + 1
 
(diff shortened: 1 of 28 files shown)


Database/compiler.lua
```diff
@@ -1095,6 +1048,10 @@ function QuestieDBCompiler:Compile()
     QuestieDBCompiler:CompileItems()
     print("\124cFFAAEEFF"..l10n("Questie DB update complete!"))
 
+    if Questie.db.profile.debugEnabled then
+        print("\124cFFAAEEFF" .. "Compiling took " .. (GetTime() - QuestieDBCompiler.startTime) .. " seconds")
```
Logonz: Change this to use `GetTimePreciseSec` here and for `QuestieDBCompiler.startTime` instead - much more accurate.

Modules/Quest/QuestieQuest.lua
```diff
@@ -724,7 +784,9 @@ function QuestieQuest:UpdateQuest(questId)
                             "[QuestieQuest:UpdateQuest] All Quest Objective(s) are Complete! Manually setting quest to Complete!")
 
                         -- Only remove the map icons, but keep the tooltips
-                        QuestieMap:UnloadQuestFrames(questId)
+                        ThreadLib.ThreadInstant(function()
```
Logonz: You should probably use the same pattern as before to not create a race-condition

```
            ThreadLib.ThreadCallbackInstant(function()
                QuestieMap:UnloadQuestFrames(questId)
            end, function()
                QuestFinisher.AddFinisher(quest)
            end)
```

However i do not like this at all as we set states before the unload and add happens as well

PR #7728, opened 2026-08-01
Enable AutoCompleteFrame for Titan Reforged Classic

kuletco:
On Titan Reforged Classic, a quest after 《昔日的传奇》 named 《另辟蹊径》 cannot popup the AutoCompleteFrame, so need disable this addon.

I enable this feature for Titan as Mop or Cata.

## Screenshots

AutoCompleteFrame

### DIFF
--- a/Modules/Tracker/QuestieTracker.lua
+++ b/Modules/Tracker/QuestieTracker.lua
@@ -104,7 +104,7 @@ function QuestieTracker.Initialize()
     trackerHeaderFrame = TrackerHeaderFrame.Initialize(trackerBaseFrame, QuestieTracker.Update)
     trackerQuestFrame = TrackerQuestFrame.Initialize(trackerBaseFrame, trackerHeaderFrame)
 
-    if Expansions.Current >= Expansions.Cata then
+    if Expansions.Current >= Expansions.Cata or Questie.IsTitanReforged then
         AutoCompleteFrame.Initialize(trackerBaseFrame)
     end
 
--- a/Questie-WOTLKC.toc
+++ b/Questie-WOTLKC.toc
@@ -1,4 +1,4 @@
-## Interface: 38000, 38001
+## Interface: 38000, 38002
 ## Title: Questie|cFF00FF00 v11.33.3|r
 ## Author: Aero/Logon/Muehe/TheCrux(BreakBB)/Drejjmit/Dyaxler/Cheeq/TechnoHunter/Everyone else
 ## Notes: A standalone Classic QuestHelper
@@ -248,6 +248,7 @@ Modules\Journey\QuestieSearch.lua
 Modules\Journey\QuestieSearchResults.lua
 
 # Tracker
+Modules\Tracker\AutoCompleteFrame.lua
 Modules\Tracker\QuestieTracker.lua
 Modules\Tracker\Sorter\Sorter.lua
 Modules\Tracker\Sorter\byComplete.lua


Modules/Tracker/QuestieTracker.lua
```diff
@@ -104,7 +104,7 @@ function QuestieTracker.Initialize()
     trackerHeaderFrame = TrackerHeaderFrame.Initialize(trackerBaseFrame, QuestieTracker.Update)
     trackerQuestFrame = TrackerQuestFrame.Initialize(trackerBaseFrame, trackerHeaderFrame)
 
-    if Expansions.Current >= Expansions.Cata then
+    if Expansions.Current >= Expansions.Cata or Expansions.isTitan then
```
BreakBB: You can use `Questie.IsTitanReforged` instead.

Modules/Expansions.lua
```diff
@@ -17,4 +17,5 @@ Expansions.Era = expansionOrderLookup[WOW_PROJECT_CLASSIC or 2]
 Expansions.Tbc = expansionOrderLookup[WOW_PROJECT_BURNING_CRUSADE_CLASSIC or 5]
 Expansions.Wotlk = expansionOrderLookup[WOW_PROJECT_WRATH_CLASSIC or 11]
 Expansions.Cata = expansionOrderLookup[WOW_PROJECT_CATACLYSM_CLASSIC or 14]
-Expansions.MoP = expansionOrderLookup[WOW_PROJECT_MISTS_CLASSIC or 19]
\ No newline at end of file
+Expansions.MoP = expansionOrderLookup[WOW_PROJECT_MISTS_CLASSIC or 19]
+Expansions.isTitan = select(4, GetBuildInfo()) > 38000
```
BreakBB: This is not needed if using `Questie.IsTitanReforged`

Questie-WOTLKC.toc
```diff
@@ -1,4 +1,4 @@
-## Interface: 38000, 38001
+## Interface: 38000, 38002
```
BreakBB, line 1: Oh, we didn't notice there was yet another patch for titan reforged. I'll make this version bump in a separate commit as more than this is required.

BreakBB (changes requested):

Modules/Tracker/QuestieTracker.lua
Cabro:
```suggestion
    if Expansions.Current >= Expansions.Cata or Questie.IsTitanReforged then
```

Modules/Expansions.lua
```diff
@@ -17,4 +17,5 @@ Expansions.Era = expansionOrderLookup[WOW_PROJECT_CLASSIC or 2]
 Expansions.Tbc = expansionOrderLookup[WOW_PROJECT_BURNING_CRUSADE_CLASSIC or 5]
 Expansions.Wotlk = expansionOrderLookup[WOW_PROJECT_WRATH_CLASSIC or 11]
 Expansions.Cata = expansionOrderLookup[WOW_PROJECT_CATACLYSM_CLASSIC or 14]
-Expansions.MoP = expansionOrderLookup[WOW_PROJECT_MISTS_CLASSIC or 19]
\ No newline at end of file
+Expansions.MoP = expansionOrderLookup[WOW_PROJECT_MISTS_CLASSIC or 19]
```
Cabro: This is not needed

Modules/Tracker/QuestieTracker.lua
kuletco: Thank you, I need to clean up the code.

PR #7729, opened 2026-08-01
Enable AutoCompleteFrame for Titan Reforged Classic

kuletco:
On Titan Reforged Classic, a quest after 《昔日的传奇》 named 《另辟蹊径》 cannot popup the AutoCompleteFrame, so need disable this addon.

I enable this feature for Titan as Mop or Cata.

## Screenshots

AutoCompleteFrame


BreakBB (approved):
Thanks a lot @kuletco ! ❤️ We'll make sure to ship this fix really soon in a new release.

PR #7778, opened 2026-08-18
[fix] Hide vendors when not active

anon1231823:
Winter Veil / DMF vendors should not show when their respective events are not active.

### DIFF
--- a/Database/Corrections/Holidays/QuestieEvent.lua
+++ b/Database/Corrections/Holidays/QuestieEvent.lua
@@ -254,8 +254,10 @@ function QuestieEvent:Load()
     SetCVar("calendarShowDarkmoon", shouldShowDmfEvents and "1" or "0")
 
     -- TODO: Also handle WotLK which has a different starting schedule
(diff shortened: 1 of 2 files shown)


Database/Corrections/Holidays/QuestieEvent.lua
```diff
@@ -254,11 +254,30 @@ function QuestieEvent:Load()
     SetCVar("calendarShowDarkmoon", shouldShowDmfEvents and "1" or "0")
 
     -- TODO: Also handle WotLK which has a different starting schedule
-    if (Questie.IsClassic and (((not Questie.IsAnniversaryEra) and (not Questie.IsAnniversaryHardcore)) or (ContentPhases.activePhases.Anniversary >= 3)))
-            or Questie.IsTBC then
+    local dmfAvailable = (Questie.IsClassic and (((not Questie.IsAnniversaryEra) and (not Questie.IsAnniversaryHardcore)) or (ContentPhases.activePhases.Anniversary >= 3)))
+            or Questie.IsTBC
```
Cabro:
```suggestion
            or Questie.IsTBC or Questie.IsWotlk
```

This variant of DMF is also available in Wotlk

PR #7782, opened 2026-08-19
[Feature] Quest link on mouseover

anon1231823:
Tested with Chattynator and default blizzard chat

https://github.com/user-attachments/assets/e5613d27-e5a1-43b8-911f-de5a1a4c857a

https://github.com/user-attachments/assets/e2aba68d-0bbd-4096-9b74-b57b96e5f70c

### DIFF
--- a/Modules/QuestLinks/Link.lua
+++ b/Modules/QuestLinks/Link.lua
@@ -21,10 +21,14 @@ local ZoneDB = QuestieLoader:ImportModule("ZoneDB")
 local QuestieReputation = QuestieLoader:ImportModule("QuestieReputation")
 
 QuestieLink.lastItemRefTooltip = ""
+-- Points to whichever tooltip frame is currently being written to (ItemRefTooltip for clicks, GameTooltip for hovers)
+---@type Frame
+QuestieLink._activeTooltip = nil
 
 -- Forward declaration
 local _AddQuestTitle, _AddQuestStatus, _AddQuestDescription, _AddQuestRequirements, _AddDungeonInfo, _GetQuestStarter, _GetQuestFinisher, _AddPlayerQuestProgress
 local _AddTooltipLine, _AddColoredTooltipLine, _GetObjectiveText
+local _GetQuestStarterName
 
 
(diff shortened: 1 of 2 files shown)


Modules/QuestLinks/Link.lua
```diff
@@ -21,6 +21,9 @@ local ZoneDB = QuestieLoader:ImportModule("ZoneDB")
 local QuestieReputation = QuestieLoader:ImportModule("QuestieReputation")
 
 QuestieLink.lastItemRefTooltip = ""
+-- Points to whichever tooltip frame is currently being written to (ItemRefTooltip for clicks, GameTooltip for hovers)
+---@type Frame
+QuestieLink._activeTooltip = nil
```
Logonz: This is scary, having a global mutable variable that required to be set and cleared in correct sequences.
Old code like this is why Questie is so unstable 😆 

I would say if you want to have 2 different tooltips you have to make it functional, pass the tooltip around in the functions (Requires a bigger change to basically all functions)

Modules/QuestLinks/Link.lua
```diff
@@ -423,6 +428,111 @@ _AddPlayerQuestProgress = function(quest, starterName, starterZoneName, finisher
...
+-- Chattynator hook: Chattynator replaces default chat frames entirely.
+-- Its handler frame is ChattynatorHyperlinkHandler (a ScrollingMessageFrame
+-- with hyperlinksEnabled="true" that receives propagated events from its chat frames).
+local function HookChattynator()
+    local chattynator = _G.Chattynator
+    local handler = chattynator and chattynator.API and chattynator.API.GetHyperlinkHandler()
+    if not handler then return false end
+
+    local existingOnEnter = handler:GetScript("OnHyperlinkEnter")
+    handler:SetScript("OnHyperlinkEnter", function(_, hyperlink)
```
Logonz: Here you overwrite blizzards OnHyperlinkEnter, this can introduce problems if other addons do it or expect to be called and isn't. Did you test if a regular HookScript worked here?

PR #7784, opened 2026-08-19
Isolate Titan Reforged corrections in their own compiled database

Logonz:
## Issue references

N/A

### Summary
Previously, Titan Reforged-only records were compiled into the shared WotLK database because Titan and ordinary WotLK used the same correction and cache path. Questie then relied on the general quest blacklist to hide those records outside Titan. This change gives Titan Reforged dedicated correction ownership and a separate compiled cache, matching the existing SoD model and preparing the Titan dataset for the upcoming Static DB.

### Before

```text
shared WotLK correction tables
  -> compile ordinary WotLK and Titan-only records together
  -> persist both in the standard WotLK bins and pointer maps
  -> hide Titan-only quests outside Titan through the general blacklist
```

Static Titan records and per-session overrides were also split across the same WotLK correction modules, making their lifecycle and ownership unclear.

### After

```text
ordinary WotLK
  -> shared WotLK corrections
  -> standard compiled database

Titan Reforged
  -> shared WotLK corrections + dedicated Titan static corrections
  -> dedicated Titan compiled database
  -> Titan runtime overrides applied per session
```

The general blacklist now handles quest availability rather than compensating for records compiled into the wrong database.

### Example

Quest `93950` previously received a pointer in the ordinary WotLK compiled database and was then hidden through `not Questie.IsTitanReforged`. It is now absent from the ordinary WotLK pointer map and compiled only into the Titan Reforged database, where it belongs.

### Responsibility / ownership

- Shared WotLK correction modules own records valid for the WotLK database.
- Dedicated Titan quest, NPC, item, object, and quest-tag modules own Titan-only static records.
- `QuestieDBCache` selects the standard, SoD, or Titan compiled namespace consistently for compilation, loading, reset, and corruption recovery.
- Titan runtime overrides remain in `MinimalInit()` because they modify inherited WotLK records per session.
- Titan availability blacklists remain Titan-gated; the blacklist no longer acts as database isolation.

This mirrors the SoD boundary and gives the Static DB work a clear Titan-specific dataset and cache target.

### Validation

- `busted -p ".test.lua" .`: 1436 successes
- WotLK database validator: passed
- Titan-mode WotLK validator: passed
- `luacheck -q -- Database Localization Modules Public Questie.lua`: passed
- `python3 build.py --all`: passed
- Lua syntax and `git diff --check`: passed
- Pointer isolation check: all 48 Titan quests absent from ordinary WotLK and present on Titan

## Screenshots

N/A — database ownership and cache refactor with no UI changes.


anon1231823:
Should SoD corrections also be it's own entity

BreakBB:
I don't like how this PR is some sort of full new feature for the database instead of just re-doing the SoD way for TitanReforged as a first step...

Logonz:
> I don't like how this PR is some sort of full new feature for the database instead of just re-doing the SoD way for TitanReforged as a first step...

I do not really think it is a new feature?
I break out all TitanReforge corrections into individual files
Do you mean the `QuestieDBCache.GetActiveStorage()` parts?
I do think those are valuable to add because we basically move out the logic of the different versions out of all the files and instead centralizes it.

BreakBB (approved):

PR #7791, opened 2026-08-21
Fix for Cluttered Quest Tooltips

### LINKED ISSUE #7289: Dungeon tooltips are too messy because they show every mob that can drop quest-starting items
### Description

Image

We only need to see the quest 1 time... not once per every mob that can drop it when I mouse over a Dungeon. (=

### Questie version

TBC

### Game flavor

TBC

Gogo1951:
## Issue references

Fixes https://github.com/Questie/Questie/issues/7289

## Proposed changes

- Shows "Multiple Sources" for quest tooltips.

### Before

image

### After

image

### DIFF
--- a/Localization/Translations/TooltipsUI.lua
+++ b/Localization/Translations/TooltipsUI.lua
@@ -145,6 +145,18 @@ local tooltipsUILocales = {
         ["ruRU"] = "Разработчик Questie",
         ["zhCN"] = "Questie 开发者",
         ["zhTW"] = "Questie 開發者",
+    },
+    ["Multiple Sources"] = {
+        ["enUS"] = true,
+        ["deDE"] = "Mehrere Quellen",
+        ["esES"] = "Múltiples fuentes",
+        ["esMX"] = "Múltiples fuentes",
+        ["frFR"] = "Sources multiples",
+        ["koKR"] = "여러 출처",
+        ["ptBR"] = "Múltiplas fontes",
+        ["ruRU"] = "Несколько источников",
+        ["zhCN"] = "多个来源",
+        ["zhTW"] = "多個來源",
     }
 }
 
--- a/Modules/FramePool/QuestieFrame.lua
+++ b/Modules/FramePool/QuestieFrame.lua
@@ -25,6 +25,7 @@ local HBDPins = LibStub("HereBeDragonsQuestie-Pins-2.0")
 ---@field Name string
 ---@field IsObjectiveNote boolean
 ---@field StarterType string|nil
+---@field DungeonEntrance boolean|nil @True when the icon is drawn on a dungeon entrance instead of the real spawn
 
 ---@class IconTexture : Texture
 ---@field r number
--- a/Modules/Quest/AvailableQuests/AvailableQuests.lua
+++ b/Modules/Quest/AvailableQuests/AvailableQuests.lua
@@ -801,6 +801,7 @@ _AddStarter = function(starter, quest, tooltipKey, limit)
                     if (coords[1] == -1 or coords[2] == -1) then
                         local dungeonLocation = ZoneDB:GetDungeonLocation(zone)
                         if dungeonLocation then
+                            data.DungeonEntrance = true
                             for _, value in ipairs(dungeonLocation) do
                                 QuestieMap:DrawWorldIcon(data, value[1], value[2], value[3])
                                 added = added + 1
--- a/Modules/Quest/QuestFinisher.lua
+++ b/Modules/Quest/QuestFinisher.lua
@@ -106,6 +106,7 @@ _AddFinisherToMap = function(finisher, quest, key, playerZone)
                 if (coords[1] == -1 or coords[2] == -1) then
                     local dungeonLocation = ZoneDB:GetDungeonLocation(finisherZone)
                     if dungeonLocation then
+                        data.DungeonEntrance = true
                         for _, value in ipairs(dungeonLocation) do
                             local zone = value[1];
                             local x = value[2];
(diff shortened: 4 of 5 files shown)


anon1231823:
I have a different idea. What about separating with a comma for each mob? Instead of Listing it out it could be like

mob A, mob B, mob C, ..., mob n
Quest name

instead of 
mob A
quest name
mob B
quest name
...
mob n
quest name

Gogo1951:
I mocked this up with real data to see how it'd actually look, and I don't think it holds up:

image

It ends up being a wall of text that isn't really telling the player anything actionable. They already know they need to go into the dungeon, and once they're inside, the unit tooltips show which mobs drop what they're after.

"Multiple Sources" seems like an easy UX win.

Gogo1951:
And just keep in mind we still have other quests to display, as well as quest progress to display.

image

What's been submitted is about as clean as it gets.

image

Gogo1951:
Here's something I do think is an improvement. 

image

What changed. The tooltip is now two columns with a header: a Questie // Available Quests title with the (Hold Shift) hint on the right. Below, each quest giver gets its own green line with its quests underneath — title left, info right — and blank lines between entries. The old layout ran source and quest on separate lines with no separation, so Blackrock Depths became an unbroken 27-line wall. The Multiple Sources collapse is unchanged.

Why it's better. The right column now carries the quest tag — (Dungeon), (Repeatable) — stacked so you can scan an entrance and see at a glance which quests repeat. Shift swaps it for XP and actual objectives with quantities: what to collect and how much. That replaces the old shift view, which repeated an identical Instance: Blackrock Depths line under all eighteen quests plus full flavor prose, quadrupling the height for almost nothing actionable. Quests with nothing to collect show the turn-in instead, in dimmer gray. Shift is now about twice the default height, not four times.

Gogo1951:
BRD was a ludicrous edge case. Here's a more realistic example of the UX changes I'm proposing.

Default shows quests, Shift adds objectives, Alt adds chain and objectives. And unlike today, the tooltip tells you what each modifier does instead of leaving you to guess.

image

Hey @BreakBB What do you think?

anon1231823:
Yeah, I guess my comma separated list is a fail.

anon1231823:
> Here's something I do think is an improvement.
> image
> 
> What changed. The tooltip is now two columns with a header: a Questie // Available Quests title with the (Hold Shift) hint on the right. Below, each quest giver gets its own green line with its quests underneath — title left, info right — and blank lines between entries. The old layout ran source and quest on separate lines with no separation, so Blackrock Depths became an unbroken 27-line wall. The Multiple Sources collapse is unchanged.
> 
> Why it's better. The right column now carries the quest tag — (Dungeon), (Repeatable) — stacked so you can scan an entrance and see at a glance which quests repeat. Shift swaps it for XP and actual objectives with quantities: what to collect and how much. That replaces the old shift view, which repeated an identical Instance: Blackrock Depths line under all eighteen quests plus full flavor prose, quadrupling the height for almost nothing actionable. Quests with nothing to collect show the turn-in instead, in dimmer gray. Shift is now about twice the default height, not four times.

I think the objetive in the tooltip is kind of cool. But what happened to the quest objective when you hold shift?

This is what we have now
image

Gogo1951:
> Yeah, I guess my comma separated list is a fail.

Everything eventually breaks at scale. (=

Gogo1951:
> I think the objetive in the tooltip is kind of cool. But what happened to the quest objective when you hold shift?

image

Logonz:
I definitely do not like showing the objectives, that is basically removing one clutter to add another clutter.
I don't know if I like the combination, but if I had to choose I only think quest givers that have one quest should overlap.
So if you look at Shadowforge Citizen it provides a crumpled up note + other quests, in your variat it is inside "Multiple sources" which is not true at all. The other quests are not available from multiple sources.

I think like a cutoff value say 3 npcs in the tooltip provide the same quest and only that quest, those would be combined, but if they provide any other unique quest they would not be combined at all.
Because we don't want to lose information just reduce clutter.

The main reason I don't like it is if shift uncollapses everything you will get even worse jump in the tooltip than today, you will basically have to scan the entire tooltip again with your eyes to now find what you where looking for reducing the ease of finding information quickly for a little bit smaller tooltip in capital cities. Don't know if the trade is worth it.

Yttrium-tYcLief:
There's plenty more to discuss and figure out here but I'm hung up on one specific point. @Gogo1951 In your mockups, you're proposing showing raw quest objective data ("Dark Iron Geologist x8, Silk Cloth x15"). Where are you getting that data from?

anon1231823:
> There's plenty more to discuss and figure out here but I'm hung up on one specific point. @Gogo1951 In your mockups, you're proposing showing raw quest objective data ("Dark Iron Geologist x8, Silk Cloth x15"). Where are you getting that data from?

I was wondering the same for the objectives like "Turn in to..."

Gogo1951:
Hey @Yttrium-tYcLief

I can see that Questie doesn't store the quantity data, which I do think is a missing piece. It would be better if Questie had this info, for sure.

But often it says in the helper text and you can just parse it in, "10 fragments" or something like this. It's not perfect.

Locations are easier, we know the zones the things drop in. So we can show those (if something drops in a zone other than the quest giver), or "Auction House" if it's a commodity.

Just been playing around, it's working more or less. 

image

Two goals:

1. Make it easier to see what quests take me out of a zone that I'm currently in. (Those I can do later... or never. Ha!)

2. Remind myself if I need to pick up precursors before I go to a zone.

Ultimately, accomplishing these goals helps players quest faster and happier.

Gogo1951:
Playing around some more, I do really like seeing where Objectives are. 

image

I gotta figure out a few loose ends, but this is a real improvement for quest planning. And it just helps memory, "Oh right it's that quest..."

Gogo1951:
And I really haven't found any quests where we can't extract the quantity.

So far it's working pretty great. Needs a lot of UX work though.

image

Gogo1951:
image

It's just so helpful to see what items can be bought. Playing around with UX but man, this has saved me so many trips already as I'm working through these zones.

PR #7794, opened 2026-08-22
[Feature] Auto reject quests from non-friends

anon1231823:
The option to reject quests from non-friends has been grayed out in the auto tab for years and non-functional, now it is.

Quests that we have labeled to not auto-accept, we will also not auto-reject them. This prevents for example auto-rejecting an escort quest.


Modules/Auto/AutoQuesting.lua
```diff
@@ -9,7 +9,111 @@ local _StartStoppedTalkingTimer, _AllQuestWindowsClosed, _IsAllowedNPC, _IsQuest
 
 local shouldRunAuto = true
 
+--- Checks if the given player is on the friends list. Returns nil when no player name is given.
+---@param playerName string|nil
+---@return boolean|nil
+local function _IsFriend(playerName)
+    if not playerName then
+        return nil
+    end
```
BreakBB: We don't need a `nil` check here as you already did that outside of the function. Then the types are clean as well.

```suggestion
--- Checks if the given player is on the friends list.
---@param playerName string
---@return boolean
local function _IsFriend(playerName)
```

Modules/Auto/AutoQuesting.lua
```diff
@@ -9,7 +9,111 @@ local _StartStoppedTalkingTimer, _AllQuestWindowsClosed, _IsAllowedNPC, _IsQuest
...
+    if (not AutoQuesting.IsModifierHeld()) and unitType == "Player" then
+        if Questie.db.profile.autoAccept.rejectSharedInBattleground and UnitInBattleground("player") then
+            DeclineQuest()
+            Questie:Print(l10n("Automatically rejected quest shared by player."))
+            return
+        end
+
+        if Questie.db.profile.autoreject_nonfriend then
+            local playerName = UnitName("questnpc")
+            if playerName and _IsFriend(playerName) == false then
```
BreakBB, line 102: In this case we can use `C_FriendList.IsFriend` because we have the GUID of the player sharing the quest. Would make things easier as we don't have to iterate all possible friend lists. I know that API is available in Anniversary TBC and MoP, not sure about Era.

I am also not sure if `UnitGUID("questnpc")` also works for escort quests 🤔 But if, we could use the same for the other `_IsFriend` call.

anon1231823: > In this case we can use `C_FriendList.IsFriend` because we have the GUID of the player sharing the quest. Would make things easier as we don't have to iterate all possible friend lists. I know that API is available in Anniversary TBC and MoP, not sure about Era.
> 
> I am also not sure if `UnitGUID("questnpc")` also works for escort quests 🤔 But if, we could use the same for the other `_IsFriend` call.

I don't think `C_FriendList.IsFriend` will also account for Battle.Net friends

PR #7801, opened 2026-08-25
[fix] Correct Darkmoon Faire scheduling and locations

Logonz:
## Issue references

Related to #7778

### Summary
Darkmoon Faire scheduling and location handling were split across expansion-specific code. This could leave quest markers hidden or NPCs mapped to the wrong location. Calendar queries could also miss an active Faire when another month was selected, and startup could cache NPC locations before holiday overrides were applied.

This PR centralizes DMF resolution in a dedicated module with explicit expansion/season rules for timing, location, and phase availability.

- Keep calculated monthly schedules for Era/TBC and use exact fourteen-day civil-date cycles for SoD, avoiding recurrence drift and local timezone/DST dependence.
- Use native calendar timestamps for WotLK/Titan, Cata, and MoP instead of inferring their schedules. WotLK/Titan locations come from calendar artwork; Cata/MoP use Darkmoon Island.
- Select the queried calendar month explicitly and restore the player's month/filter afterward. Retry incomplete data with a bounded startup wait, then apply corrections before DB initialization.
- Centralize mainland NPC location overrides, preserve normal spawns and both faction announcement quests on the island, and correct the zone IDs of six island NPCs.
- Register SoD deck quests for event-specific turn-in handling while preserving exclusions for the replaced original decks.
- Preserve the actual in-game hour/minute in the legacy calendar compatibility fallback.

### Findings and validation
- Live MoP data confirms two September 2026 Faires, not a universal first-Sunday monthly schedule. It also reproduced missing event records when querying September with October selected. `SetMonth(0)` refreshes the selected month; it does not select today's month.
- MoP and Titan tester captures confirm native timestamps and matching day-event/holiday texture values. Titan's captured August Elwynn and September Mulgore records match the DBC mappings. Terokkar is supported by DBC evidence, not a separate live capture.
- CI unit tests, Luacheck, and database/loader/localization validation checks pass.

Native API captures do not establish normal startup, cold-login readiness, or actual opening/closing transitions for the corrected addon. Those remain validation limits, not additional implemented features. DMF state remains a one-shot startup snapshot; this PR does not add live transition refresh.

The [DMF module guide](https://github.com/Questie/Questie/blob/248464b02ae282564cdafb67f94f89a8ac908137/Database/Corrections/Holidays/DMF.md) contains the configuration contract, runtime flow, evidence, and reproduction details.

## Screenshots

No new UI elements. Native calendar observations are recorded in the module guide.


Database/Corrections/Holidays/QuestieEvent.test.lua
```diff
@@ -243,16 +269,62 @@ describe("QuestieEvent", function()
...
+
+            Questie.IsClassic = true
+            Questie.IsSoD = true
+
+            QuestieEvent:Load()
+
+            assert.spy(DarkmoonFaireFixes.GetNpcFixes).was.called_with("MULGORE")
+        end)
+
+        it("should apply NPC corrections without visible Darkmoon Faire quests", function()
```
BreakBB, line 303: I don't really understand this test case. Without DMF we don't want corrections to be applied, do we?

Maybe the name could be better?

Database/Corrections/Holidays/darkmoonFaireFixes.lua
```diff
@@ -0,0 +1,101 @@
...
+}
+
+---Returns NPC spawn corrections for a detected location, or nil when the location has no correction data.
+---@param eventLocation DMFLocation
+---@return table<NpcId, table>?
+function DarkmoonFaireFixes.GetNpcFixes(eventLocation)
+    return npcFixesByLocation[eventLocation]
+end
+
+return DarkmoonFaireFixes
```
BreakBB, line 101: This is no longer needed, is it?

Database/Corrections/Holidays/QuestieEvent.test.lua
Logonz: Hmm yeah you are correct.
Taking it one step further I wonder if we should clear the coordinates for the DMF mobs unless DMF is active so basically always apply a correction but a clearing one as well.

Database/Corrections/Holidays/darkmoonFaireFixes.lua
Logonz: You are correct

Database/Corrections/Holidays/QuestieEvent.test.lua
BreakBB: That is something @Cabro should chime in to. I think we are hiding the quests outside of DMF and therefore the NPC locations don't really matter.

Cabro: It would be great if we can do a clearing spawns correction, in my eyes.

Database/Corrections/Holidays/darkmoonFaireFixes.test.lua
BreakBB, line 1: We shouldn't start testing DB corrections like this.

Database/Corrections/Holidays/DMF.md
BreakBB, line 1: Not needed anymore is it? Looks like full scale context bloat.

Database/Corrections/Holidays/QuestieEvent.lua
```diff
@@ -68,7 +68,8 @@ QuestieEvent.eventQuests = {}
 
 ---@type table<QuestId, boolean>
 QuestieEvent.activeQuests = {}
-QuestieEvent.calendarDataCached = false
+---@type boolean @True once the startup snapshot has been applied.
+QuestieEvent.initialized = false
```
BreakBB, line 72: Why do we need this? Why would `QuestieEvent.Initialize` be called more than once on init?

Database/Corrections/Holidays/QuestieEvent.lua
```diff
@@ -84,66 +85,130 @@ local alwaysTurnInAbleQuests = {
...
-local DMF_START_DAY_BY_FIRST_WEEKDAY = {
-    [1] = 9,  -- 1st is a Sunday -> Monday is 9th
-    [2] = 8,  -- 1st is a Monday -> Monday is 8th
-    [3] = 7,  -- 1st is a Tuesday -> Monday is 7th
-    [4] = 6,  -- 1st is a Wednesday -> Monday is 6th
-    [5] = 5,  -- 1st is a Thursday -> Monday is 5th
-    [6] = 4,  -- 1st is a Friday -> Monday is 4th
-    [7] = 10, -- 1st is a Saturday -> Monday is 10th
-}
+local initializationStarted = false
```
BreakBB, line 114: Same as above

Database/Corrections/Holidays/QuestieEvent.test.lua
```diff
@@ -1,857 +1,442 @@
...
-            assert.is_nil(QuestieEvent.eventQuests)
-            assert.is_equal(0, #QuestieEvent.activeQuests)
-            assert.spy(getNumDayEventsMock).was.called_with(0, 23)
+    describe("resolver integration", function()
+        it("uses Era civil dates without depending on a calendar API", function()
+            Expansions.Current = Expansions.Era
+            QuestieCompat.GetCurrentCalendarTime = function()
+                return {year = 2026, month = 8, monthDay = 10, hour = 3, minute = 0}
+            end
+            C_Calendar = nil
```
BreakBB, line 294: Era also has the `C_Calendar.OpenCalendar` APIs available but as no-op, so I am not sure this is a correct test case or a false positive.

Database/Corrections/Holidays/QuestieEvent.test.lua
```diff
@@ -1,857 +1,442 @@
...
+                    startDate = case[7], endDate = case[8], startHour = 10, startMinute = 0, endHour = 10, endMinute = 0,
+                }}
+                QuestieEvent.eventQuests = {{"Other", 123}}
+                QuestieEvent:Load()
+                assert.equals(case[9], QuestieEvent.activeQuests[123] == true)
+                if case[9] then
+                    assert.spy(print).was.called(1)
+                else
+                    assert.spy(print).was.not_called()
+                end
```
BreakBB, line 407: why not split this up in two parameterized tests to remove this bad practice?

Database/Corrections/cataNPCFixes.test.lua
BreakBB, line 1: Not a useful test and should be removed.

BreakBB (changes requested):

PR #7802, opened 2026-08-25
Send available instead unavailable daily quests

BreakBB:
This needs testing and a rebase.


Modules/Network/DailyQuestComms/DailyQuestComms.lua
```diff
@@ -8,31 +8,27 @@ local CommsEncoding = QuestieLoader:ImportModule("CommsEncoding")
...
----@class RequestUnavailableDailyQuestsEvent
----@field eventName "RequestUnavailableDailyQuests"
+---@class RequestAvailableDailyQuestsEvent
+---@field eventName "RequestAvailableDailyQuests"
 ---@field data table<NpcId, QuestId[]>
 
----@alias CommEvent HideDailyQuestsEvent|RequestUnavailableDailyQuestsEvent
+---@alias CommEvent AvailableDailyQuestsEvent|RequestAvailableDailyQuestsEvent
 
 local COMM_PREFIX = "QuestieDailiesV2"
```
Logonz: You should probably bump this because you don't want to send available to an unavailable consumer

BreakBB: Absolutely, yes! Just pushed this yesterday to put it out of "local scope" and get a first rabbit review 😬

PR #7805, opened 2026-08-27
[fix] Autoabandon breadcrumb

anon1231823:
Autoabandon breadcrumbs and show all breadcrumbs in tooltip
image

https://github.com/user-attachments/assets/12d583b9-ddfb-4bd6-991e-e1a0de5a3058

### DIFF
--- a/Modules/EventHandler/QuestEventHandler.lua
+++ b/Modules/EventHandler/QuestEventHandler.lua
@@ -193,17 +193,59 @@ function QuestEventHandler.InitQuestLogStates(changes)
 end
 
 local _AbandonQuest = function(questId, breadcrumbQuestId)
-    if not QuestiePlayer.currentQuestlog[questId] then
-        return
-    end
     local questLogIndex = GetQuestLogIndexByID(questId)
-    if questLogIndex then
+    if questLogIndex and questLogIndex > 0 then
         SelectQuestLogEntry(questLogIndex)
         SetAbandonQuest()
         AbandonQuest()
         local questLink = QuestieLink:GetQuestHyperLink(questId)
         local breadcrumbLink = QuestieLink:GetQuestHyperLink(breadcrumbQuestId)
         Questie:Print(l10n("Automatically abandoned quest %s because breadcrumb quest %s is not completed.", questLink, breadcrumbLink))
+        return true
+    end
+    return false
+end
+
+--- Checks if a quest has incomplete breadcrumbs and announces/abandons accordingly.
+---@param questId number
+local _CheckBreadcrumbs = function(questId)
+    if not (Questie.db.profile.questAnnounceIncompleteBreadcrumb or Questie.db.profile.autoAccept.abandonBreadcrumbFollowup) then
+        return
+    end
+
+    local breadcrumbs = QuestieDB.QueryQuestSingle(questId, "breadcrumbs")
+    if breadcrumbs then
+        local lastIncompleteBreadcrumbId = nil
+        for _, breadcrumbQuestId in ipairs(breadcrumbs) do
+            -- We want to let users know when they picked up a quest without finishing its breadcrumb
+            if (not Questie.db.char.complete[breadcrumbQuestId]) and (not QuestiePlayer.currentQuestlog[breadcrumbQuestId]) then
+                local requiredRaces = QuestieDB.QueryQuestSingle(breadcrumbQuestId, "requiredRaces")
+                local requiredClasses = QuestieDB.QueryQuestSingle(breadcrumbQuestId, "requiredClasses")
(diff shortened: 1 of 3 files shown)


Modules/EventHandler/QuestEventHandler.lua
```diff
@@ -235,43 +277,26 @@ function QuestEventHandler.QuestAccepted(questLogIndex, questId)
...
-                        if Questie.db.profile.autoAccept.abandonBreadcrumbFollowup then
-                            _AbandonQuest(questId, breadcrumbQuestId)
-                        end
-                    end
-                end
-            end
-        end
+--- Checks all existing quests in the quest log for incomplete breadcrumbs.
+--- Called during login initialization since QUEST_ACCEPTED does not fire for quests already in the log.
+function QuestEventHandler.CheckExistingQuestBreadcrumbs()
```
Logonz: This function is never used?

Modules/EventHandler/QuestEventHandler.lua
```diff
@@ -193,17 +193,59 @@ function QuestEventHandler.InitQuestLogStates(changes)
...
         local breadcrumbLink = QuestieLink:GetQuestHyperLink(breadcrumbQuestId)
         Questie:Print(l10n("Automatically abandoned quest %s because breadcrumb quest %s is not completed.", questLink, breadcrumbLink))
+        return true
+    end
+    return false
+end
+
+--- Checks if a quest has incomplete breadcrumbs and announces/abandons accordingly.
+---@param questId number
+local _CheckBreadcrumbs = function(questId)
```
Logonz: I know this function / code has been moved from within this file, but this function should not be here.

QuestEventHandler is the "junction box" - this function is specific to the announce, auto module and maybe others.

(ps. same goes for _AbandonQuest , don't we have any AbandonQuest modules? I would say that it should be a library function somewhere)

My suggestion is to move these things out of QuestEventHandler and call them instead to keep it clean.

Modules/EventHandler/QuestEventHandler.lua
```diff
@@ -235,43 +277,26 @@ function QuestEventHandler.QuestAccepted(questLogIndex, questId)
...
+function QuestEventHandler.CheckExistingQuestBreadcrumbs()
+    for questId, _ in pairs(QuestiePlayer.currentQuestlog) do
+        _CheckBreadcrumbs(questId)
     end
 end
 
 ---@param questId number
 function _QuestEventHandler:HandleQuestAccepted(questId, isRetry)
+    -- The quest may have been abandoned (e.g. auto-abandon for incomplete breadcrumb) while waiting for the cache
+    local questLogIndex = GetQuestLogIndexByID(questId)
```
Logonz, line 229: Good change! Much more stable.

Modules/Tooltips/MapIconTooltip.lua
```diff
@@ -322,12 +322,12 @@ function MapIconTooltip:Show()
                                 QuestiePlayer.HasRequiredRace(requiredRaces) and QuestiePlayer.HasRequiredClass(requiredClasses) and
                                 (not exclusiveQuestCompleted) and (not Questie.db.char.complete[availableUntilCompleted]) then
                                 if firstBreadcrumb then
-                                    local breadcrumbLevel, _ = QuestieLib.GetEffectiveQuestLevel(breadcrumbId)
-                                    local questTitle, rewardString = _MapIconTooltip.GetNextQuestInChainLines(breadcrumbId, breadcrumbLevel, breadcrumbTitleIndent)
                                     tooltipRows:AddLine(breadcrumbLabelPrefix .. l10n("Breadcrumb Quests") .. l10n(": "), 0.86, 0.86, 0.86)
-                                    tooltipRows:AddDoubleLine(questTitle, rewardString, 1, 1, 1)
                                     firstBreadcrumb = false
                                 end
+                                local breadcrumbLevel, _ = QuestieLib.GetEffectiveQuestLevel(breadcrumbId)
```
Logonz, line 328: Just for my own knowledge why was this moved out of the if-case? What was the behavior before, what is the behavior after?

anon1231823: Moved it out so that every eligible breadcrumb quest renders its own title/reward line, not just the first one. Previously the label was only added around the first breadcrumb's detail, so additional eligible breadcrumbs produced no entries.

Modules/EventHandler/QuestEventHandler.lua
```diff
@@ -501,6 +461,7 @@ function _QuestEventHandler:UpdateAllQuests(doRetryWithoutChanges)
                 end
             end
 
+            --Questie.Debug(Questie.DEBUG_INFO, "Quest:", questId, "objectives:", table.concat(objIds, ","), "will be updated")
```
Logonz: Remove this comment or enable the print

PR #7820, opened 2026-09-01
[fix] Prune departed party members from tooltips

### LINKED ISSUE #7780: Party icons tooltip remembered party members after they left
### Description

Only me (Cheeqbehu) + the druid (Ironpaws) were in group when i took the quest. the white names on there are party members that are not in group anymore. they left BEFORE i accepted the quest.

Image

there was no mouseover tooltip on the NPC with the other 2 party members when i talked.

### Questie version

11.36

### Game flavor

TBC

aleksandr-kotlyar:
## Issue references

Fixes #7780

## Proposed changes

- Remove departed members' group quest snapshots and tooltip registrations on roster updates, including same-size replacements and Name-Realm players.
- Track group and Nearby/YELL objectives independently per player and quest, keeping the group snapshot as the effective projection while grouped and restoring the exact Nearby snapshot when group data is removed.
- Remove only group-specific snapshots on `GROUP_LEFT`, while preserving existing Nearby/YELL completion and lazy four-minute pruning behavior.
- Reject late `PARTY`, `RAID`, `INSTANCE_CHAT`, and `WHISPER` packets from departed senders before deserialization; `YELL` and other non-group transports remain unchanged.
- Keep full quest-list blocks additive and make no comms protocol changes.

## Validation

- Lua 5.1 syntax check: passed for all four changed files
- Luacheck: 0 warnings / 0 errors in 365 files
- Targeted Busted: 22 successes / 0 failures
- Full Busted: 1799 successes / 0 failures (`TZ=Europe/Berlin`)
- `git diff --check`: passed

## Screenshots

Existing reproduction is in #7780. Live two-client verification remains pending.

### DIFF
--- a/Modules/Network/QuestieComms.lua
+++ b/Modules/Network/QuestieComms.lua
@@ -37,6 +37,11 @@ QuestieComms.remotePlayerClasses = {}
 QuestieComms.remotePlayerEnabled = {}
 QuestieComms.remotePlayerTimes = {}
 
+-- Players whose quest data was received while they were in the group. Keep this separate from
+-- remotePlayerTimes because that table tracks nearby/YELL players, whose data must not be pruned
+-- just because they are not in the current party or raid.
+local remoteGroupPlayers = {}
+
 -- The idea here is that all messages with the same "base number" are compatible
 -- New message versions increase the number by 0.1, and if the message becomes "incompatible" you increase with 1
 -- Say if the message is 1.5 it is valid as long as it is < 2. If it is 2.5 it is not compatible for example.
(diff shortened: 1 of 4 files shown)


BreakBB:
Hey @aleksandr-kotlyar thanks for tackling this.

Just so you know, even though there are remaining code parts `YELL` is fully unused in Questie.

Also there is a somewhat conflicting PR https://github.com/Questie/Questie/pull/7648 with which we _started_ working on some improvements towards the "party comms".

This PR will need a proper review, which I won't have time for over the next days. I'll put it high on my list.

@Cabro feel free to test this already.

Modules/Network/QuestieComms.lua
```diff
@@ -37,6 +37,24 @@ QuestieComms.remotePlayerClasses = {}
 QuestieComms.remotePlayerEnabled = {}
 QuestieComms.remotePlayerTimes = {}
 
+local REMOTE_PLAYER_TIMEOUT = 60 * 4
+local GROUP_SOURCE = "group"
+local NEARBY_SOURCE = "nearby"
```
BreakBB: There are no "nearby" tooltip sources in Questie anymore. We fully disabled YELL years ago and won't re-enable it. So feel free to remove any code you added to handle this irrelevant case.

There is still a lot of code about it around though, which we simply didn't care to remove yet (shame on us >_>). I also won't remove it now, so this PR doesn't get in conflict with these changes.

PR #7824, opened 2026-09-01
[db] Integrate Questie with QuestieTDB

Logonz:
## Problem

The dependency-clean baseline removes Questie's runtime entity database, but retained consumers still need to read entity data and publish Questie-owned runtime policy.

## Solution

- Bind `QuestieDB` reads, key enums, ID maps, and Objective Order data to `LibQuestieDB` Contract Version 1.
- Replace compiler-driven login initialization with provider locale, Correction, cache, and consumer initialization.
- Publish Questie-owned policy through write-through Dynamic Correction slots.
- Adapt Townsfolk, Available Quests, tooltips, Darkmoon state, and runtime Item repair to the composed provider view.
- Add a focused QuestieTDB Contract fake and tests for provider integration, correction replacement and withdrawal, initialization order, and retained consumers.
- Preserve the current external entity locale compatibility path and document the proposed migration to third-party-owned Dynamic Corrections.

## Validation

- `busted -p ".test.lua" .` — 1,533 successes
- `luacheck -q -- Database Localization Modules Public Questie.lua` — 0 warnings and 0 errors across 329 files
- `lua cli/validate-loader-usage.lua` — no new runtime loader calls
- `git diff --check` — clean


docs/adr/0002-questie-policy-corrections.md
```diff
@@ -0,0 +1,26 @@
+# 2. Questie publishes Policy Corrections; the provider owns entity facts
```
BreakBB, line 1: Does "Policy Corrections" mean: "Corrections based on the rules Questie wants to follow?" - The term just doesn't click for me.

Modules/Libs/QuestieLib.lua
```diff
@@ -446,6 +448,68 @@ function QuestieLib.ContinueOnQuestObjectivesLoad(questId, onSuccess, onFailure,
...
+        QuestieCorrections.SetCorrection("Item", "RuntimeItemRepair", repairedItemNames)
+    end)
+end
+
+---Asks the client for the names of objective Items the composed database lacks and repairs each one
+---through the name-only RuntimeItemRepair Policy Correction slot when the asynchronous load completes.
+---Runs on quest accept and for every quest already in the log at login.
+---@param questId QuestId
+---@return nil
+function QuestieLib.RepairMissingItemNames(questId)
```
BreakBB, line 480: Why do we need this at all?

test/QuestieDBMock.lua
BreakBB, line 1: I have no idea what to think of this. It basically means we are testing against a full fleshed close of the new QuestieDB, which is more than just a simple "mock" and always needs to be in sync with the actual QuestieDB. Imo this gives an incorrect feeling of safety.

I'd prefer to move this mock to the QuestieDB repository and grab it from there as actual dependency just like it is done in non WoW-addon projects. That way QuestieDB is in charge of keeping production code and mock in sync.

QUESTIE-OBJECT-NAME-INDEX.md
BreakBB, line 1: Is this still needed?

TDB-FINDINGS.md
BreakBB, line 1: Is this still needed?

TDB-PLAN.md
BreakBB, line 1: Is this still needed?

TDB-STATUS.md
BreakBB, line 1: Is this still needed?

CI-PROVIDER-COMPATIBILITY.md
BreakBB, line 1: Is this still needed?

Localization/l10n.lua
```diff
@@ -130,3 +145,107 @@ end
...
+    -- Quests: `[questId] = {name, {objective, ...}}`. Generators built before Questie dropped the
+    -- description element emit `{name, {description, ...}, {objective, ...}}`, so a third element
+    -- is the objectives when present.
+    local questKeys, questExists = Meta.QuestMeta.questKeys, LibQuestieDB.Quest.Exists
+    local quests = {}
+    for questId, data in pairs(_LookupRows(override.questLookup)) do
+        if type(data) == "table" and questExists(questId) then
+            local row = {
+                [questKeys.name] = _Text(data[1]),
+                [questKeys.objectivesText] = _Lines(data[3]) or _Lines(data[2]),
```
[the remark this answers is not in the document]
Logonz: Addressed in the current branch: Localization/l10n.lua selects the third slot when it is present rather than falling back based on its contents. Localization/l10n.test.lua covers empty and malformed legacy objectives. Resolving this thread.

TDB-FINDINGS.md
```diff
@@ -0,0 +1,141 @@
...
+Proposed action: accumulate names and publish once per frame (`C_Timer.After(0, ...)` guarded by a
+pending flag). Idempotency and nil-name handling stay as they are.
+
+### F3. Provider: `requiredRaces` inference — withdrawn on Era, open on SoD
+
+**Owner:** QuestieTDB, issue #13 for SoD. **Severity:** none on Era; SoD unmeasured.
+
+First reading (wrong): composed `requiredRaces` equalled raw for every quest, and applying
+upstream's starter-faction rule to composed data flagged 16 Era quests (35 plus 5 on SoD) as
+"should have been inferred". Concluded the inference was inactive.
```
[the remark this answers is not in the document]
Logonz: The current TDB-FINDINGS.md already distinguishes the 16 Era probe artifacts from the SoD count of 40, explicitly describing 40 as an upper bound requiring remeasurement. Resolving this specific finding. Historical labeling and stale follow-up status remain separate documentation work.

Logonz:
Disposition of two nitpicks in the [first CodeRabbit review summary](https://github.com/Questie/Questie/pull/7824#pullrequestreview-5082927277):

- **Import LibQuestieDB through QuestieLoader:** No change planned. LibQuestieDB is an external provider, not a registered QuestieLoader module. Importing it through that loader is not an available replacement for the provider global. A local alias is optional, but does not justify adding an adapter.
- **Copy correction rows in the fake:** No change planned on the proposed rationale. The inspected QuestieDB provider implementation also retains the supplied rows table in its correction registry (`entry.data = rows`). Copying only in the fake would introduce a difference rather than establish parity. Any change to mutation semantics should first be agreed and tested against the real provider.

These are summary-embedded nitpicks, not individually resolvable GitHub threads. The other outstanding findings remain open.

Modules/Tooltips/TooltipHandler.lua
```diff
@@ -103,47 +103,55 @@ function _QuestieTooltips:AddItemDataToTooltip()
...
         if tooltipData then
-            for _, line in pairs (tooltipData) do
-                if (not alreadyAddedObjectiveLines[line]) then
+            for _, line in pairs(tooltipData) do
+                if not alreadyAddedObjectiveLines[line] then
                     alreadyAddedObjectiveLines[line] = true
                     GameTooltip:AddLine(line)
                 end
             end
             addedObjects = addedObjects + 1
```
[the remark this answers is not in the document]
Logonz: Addressed in ba71c590d. Empty tables previously passed the truthiness check and consumed the ten-object cap without adding any lines. The handler now checks `tooltipData and next(tooltipData)`, so empty results do not prevent later objects from contributing quest lines. Added a regression test with ten empty results followed by a visible eleventh result; the existing ten-nonempty-results limit remains covered. `GetTooltip`'s return contract is unchanged.

Database/DropTables/dropDB.lua
```diff
@@ -34,26 +24,35 @@ DropDB.correctionKeys = { -- all keys must be negative or they'll be parsed as r
     PSERVER = -2,
 }
 
+---@return boolean valid
+---@return string? report
 function DropDB:Initialize()
+    local QuestieItemDropCorrections = LibQuestieDB.Support.Get("QuestieItemDropCorrections")
     if Questie.IsClassic then
+        local QuestieClassicItemDrops = LibQuestieDB.Support.Get("QuestieClassicItemDrops")
```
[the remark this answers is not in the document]
Logonz: Addressed in 8fb009e2b. DropDB validates the selected support containers before reading their fields, and ZoneDB guards its load-time bindings and reports malformed containers during initialization. Both use the same protected decoder for missing/non-string payloads, compilation errors, execution errors, and non-table results. Failures return the existing validation report instead of throwing at those access/decode sites. Tests cover these boundaries and preserve the valid-data behavior, including MoP's use of Cata private-server drops.

Modules/Libs/QuestieLib.lua
Logonz: This preserves the client-data fallback for objective items missing from the database. QuestieDB's read API is synchronous, but retrieving an uncached item name from the WoW client can require an asynchronous callback.

The flow is: Questie encounters a missing objective item → requests it through `Item:ContinueOnItemLoad` → receives the name when available → publishes a name-only `RuntimeItemRepair` correction. Subsequent database reads can then find it.

Questie owns the quest-loading flow and knows which missing items it needs, so it also owns the request and callback. Keeping this here avoids making a database read start client requests or change its return contract.

We're also keeping this database refactor narrow: preserve the existing fallback while changing where entity data comes from, rather than redesign asynchronous item loading at the same time. QuestieDB could expose a separate asynchronous loading API, but moving this behavior would be a future improvement with its own review and tests, not part of this refactor. Does that explain why we're retaining it here?

test/QuestieDBMock.lua
Logonz: I agree that QuestieDB should own this test support. Today, provider behavior changes in QuestieDB while its imitation is maintained in Questie, so our consumer tests can stay green against an outdated implementation.

The intended flow is: QuestieDB changes its implementation → updates and checks its own test double → Questie consumes that tested revision through a small setup adapter. Questie's behavior tests stay here; provider semantics and their conformance checks move to QuestieDB.

I've opened https://github.com/Questie/QuestieDB/issues/22 with the ownership boundary, implementation approach, and acceptance criteria. It includes removing QuestieLoader coupling from the shared fixture and having provider CI verify it against the real implementation, rather than just moving the files.

In the meantime, 8ee3a077e adds a CI job against real QuestieDB master and makes a missing configured checkout fail instead of skipping. That catches drift in the existing Source-mode conformance cases, but does not replace the ownership move or prove full fidelity. Does handling that move in the linked follow-up work for you?

CI-PROVIDER-COMPATIBILITY.md
Logonz: The separate root-level document is gone in fdebdfb3e. Its useful requirements now live under "Planned release compatibility checks" in `docs/questiedb-integration.md`, beside the actual testing setup. The guide distinguishes the implemented master conformance job from the still-unimplemented supported-release/Baked coverage and required-release/advisory-master split, so those requirements are preserved without a second status document.

QUESTIE-OBJECT-NAME-INDEX.md
Logonz: Not as a separate refactor document. fdebdfb3e removes it and moves the maintained behavior into the Object-hover section of `docs/questiedb-integration.md`. That keeps the provider/tooltip ownership boundary, uniqueness and zone filtering, party registrations, and cache lifecycle together with the rest of the integration.

TDB-FINDINGS.md
Logonz: fdebdfb3e removes the root-level findings file. The useful measurements and failure explanations are retained in `docs/tdb-history.md`, explicitly marked as historical Contract Version 1 evidence. Current Contract Version 2 validation requirements live in `docs/questiedb-integration.md`, so old smoke results no longer share a status document with current acceptance criteria.

TDB-PLAN.md
Logonz: Removed in fdebdfb3e. Its remaining flavor/locale/calendar checks are now in the integration guide's live-validation matrix. The temporary agent instructions, workstation paths, and old checkout-specific tasks were dropped rather than carried into the maintained documentation.

TDB-STATUS.md
Logonz: fdebdfb3e replaces it with `docs/questiedb-integration.md`, which documents the current ownership, initialization, corrections, testing, and unfinished integration checks. Historical branch hashes and validation results moved into `docs/tdb-history.md`. There is now one maintained integration guide instead of overlapping refactor status and plan files.

TDB-STATUS.md
```diff
@@ -0,0 +1,74 @@
...
+`Database/Corrections/QuestieCorrections.lua`, Stage 1 of `Modules/QuestieInit.lua`, and the header
+of `Localization/l10n.lua`.
+
+## Branches and merge order
+
+- `QuestieTDB-remove-baseline`: deletes the compiler, raw data, generated localization, static
+  corrections, and validators from `master`. Intentionally non-functional. Never merges alone.
+- `QuestieTDB-implementation`: cut from the baseline, lands the provider binding and Policy
+  Corrections. Green: full Busted, production luacheck, loader-usage validation.
+- Merge `implementation` into `baseline`, then the combined branch into `master`. Immediately before
```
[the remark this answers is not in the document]
Logonz: The old merge instruction is removed in fdebdfb3e along with `TDB-STATUS.md`. `docs/tdb-history.md` retains the full branch names as historical provenance, explicitly not a current merge recipe. The maintained integration guide no longer presents shorthand branch names as instructions.

Database/QuestXP/QuestieXP.lua
```diff
@@ -2,11 +2,15 @@
...
+---@type SupportValidation
+local SupportValidation = QuestieLoader:ImportModule("SupportValidation")
+
 ---@type Expansions
 local Expansions = QuestieLoader:ImportModule("Expansions")
 
+-- Shared, read-only base XP; level and buff adjustments stay in this wrapper.
 ---@type table<QuestId,table<Level,XP>> -- { questId={level, xp}, ..... }
-QuestXP.db = {}
+QuestXP.db = LibQuestieDB.Support.Get("QuestXP").db
```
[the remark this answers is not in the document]
Logonz: Addressed in 63d4f3645. QuestXP now checks that the support result is a table before reading `.db`, so missing or malformed support reaches the existing initialization validation instead of failing during file loading. Added tests for missing and non-table support; existing coverage still verifies that valid XP data retains the provider's shared table.

docs/adr/0002-questie-policy-corrections.md
Logonz: Yes, roughly: these are Questie-owned overrides of provider data, selected using Questie's settings or runtime state. Examples are the active Darkmoon location and gathering-node display settings. Rules that only decide what Questie shows stay local; we use a Correction when the value must change in the provider's composed entity rows. I've made that distinction explicit in the ADR in 63d4f3645, without renaming the APIs.

docs/adr/0002-questie-policy-corrections.md
```diff
@@ -0,0 +1,26 @@
...
+
+Some entity values depend on state only Questie has: which Darkmoon Faire location the calendar
+selected, the active TBC Content Phase, the display policy that hides gathering nodes, and Item
+names the client loads asynchronously. QuestieTDB must not learn Questie's calendar, settings, or
+lifecycle, so Questie writes those values itself as Corrections under owner `"Questie"`.
+
+Ownership follows the information needed to choose the value. Anything QuestieTDB can decide from
+facts it owns (class, race, faction, expansion, season, SoD, Titan Reforged) is provider data, and
+Questie publishes no copies. Blacklists, hidden quests, event visibility, and availability checks are
+Questie display policy, not Corrections; they decide what Questie shows, not what an entity is.
```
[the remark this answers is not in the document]
Logonz: Addressed in 63d4f3645. The ADR now distinguishes Questie-only visibility rules, which stay local, from rules that change composed provider entity rows, which are published as Corrections. Gathering-node display policy is explicitly included in the latter category.

docs/tdb-history.md
```diff
@@ -0,0 +1,129 @@
...
+  the provider as a blank name and a cleared field. Now skipped in `l10n`.
+- **Empty-to-empty Darkmoon withdrawal reapplied.** Withdrawing a never-published slot is now a no-op.
+- **Stage 1 contract gate ran after locale work.** Moved before any provider call.
+- **Mock ID maps kept identity across applies**, which made the pointer-rebind tests vacuous. The
+  mock now swaps map identity like the provider; the tests fail when the rebinds are removed.
+- **`LibQuestieDB` leaked between test files.** Isolated in the Contract fake.
+
+Deferred by the reviewers and still open in TDB-STATUS: per-Item repair coalescing, Townsfolk's
+per-login SavedVariables writes, the dead `objectCache`, the Isle of Quel'Danas profile-versus-global
+split, and the `Expansions.Current` and `C_Calendar` leaks in the Event and QuestieLib suites.
```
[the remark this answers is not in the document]
Logonz: Addressed in 63d4f3645 and retained through fdebdfb3e. `docs/tdb-history.md` records item-repair coalescing and Townsfolk persistence removal as completed, not open follow-ups. The current unfinished work is now in `docs/questiedb-integration.md` rather than the removed cutover-status file.

TDB-FINDINGS.md
```diff
@@ -0,0 +1,185 @@
+# QuestieDB live smoke findings
+
+Everything found while running the `QuestieTDB-implementation` branch in a real client through
+the Lua bridge. One entry per finding: what was observed, the evidence, who owns the fix, and the
+proposed action. Resolved entries stay here with their resolution so the same probe is not rerun.
+Current state and merge gates: `TDB-STATUS.md`.
```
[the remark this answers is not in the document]
Logonz: Addressed in 63d4f3645 and preserved in fdebdfb3e. The Era/SoD runs now live in `docs/tdb-history.md`, explicitly labeled historical Contract Version 1 evidence. They do not establish current Contract Version 2 compatibility; the current matrix, including fresh Era/SoD runs, is in `docs/questiedb-integration.md`.

Logonz:
Follow-up on the remaining summary-only CodeRabbit findings:

- **Restore `Questie.IsSoD` in Townsfolk tests** ([review](https://github.com/Questie/Questie/pull/7824#pullrequestreview-5082927277)): addressed in 63d4f3645. The suite now captures and restores the flag alongside its other saved state.
- **Use a plain function for `LoadContentPhaseFixes`** ([review](https://github.com/Questie/Questie/pull/7824#pullrequestreview-5088585350)): deferred as style-only cleanup. Keeping the declaration and callers unchanged avoids unrelated churn in the database refactor.
- **Additional LuaLS annotations for migration callbacks and the mock-loader variables**, including `cli/integrationTests/6734.test.lua` ([review](https://github.com/Questie/Questie/pull/7824#pullrequestreview-5245717288)): also deferred as separate annotation cleanup. These were not implemented by the review-fix commits.

These findings are embedded in review summaries and have no individual threads to resolve. The two previously rejected provider-global/table-copying suggestions remain covered by https://github.com/Questie/Questie/pull/7824#issuecomment-5729002254.

PR #7845, opened 2026-09-17
[locale] Add itIT support to Questie translations

anon1231823:
Translation files: real Italian (informal 'tu' register) in every Localization/Translations entry except ExtraObjectives and lookups, following the pattern of frFR/esES/ptBR locales. Infinitive verb forms for labels and menu commands, imperatives for user instructions.

Plumbing: itIT in l10n supportedLocals, language dropdown, FormatDate date formatting, plus tests.

Pending confirmation that Forever will support italian, although it appears to be the case
image


BreakBB:
1. We want to mark translations generated by an AI and **not verified** from a human profound enough in that language
2. We are close to replace our entire database by [QuestieDB](https://github.com/Questie/QuestieDB) which will also cause the lookup data to be moved. We need to keep this in mind before starting to add lookup data.

anon1231823:
> 1. We want to mark translations generated by an AI and **not verified** from a human profound enough in that language
> 
> 2. We are close to replace our entire database by [QuestieDB](https://github.com/Questie/QuestieDB) which will also cause the lookup data to be moved. We need to keep this in mind before starting to add lookup data.

1. It can be assumed it's all AI translated. However, I did review to ensure blizzard terms are being used in places such as raid and dungeon. So it is already marked. "itIT" is the marking. Hopefully @[Alessandro-Barbieri](https://github.com/Alessandro-Barbieri) can take a look at this and give feedback.

2. No lookup data is translated and intentionally omitted

anon1231823:
I can open this up later if needed, but it looks like you can't actually change language to italian. I don't think Forever is going to support it. Wago also doesn't have any of the zone names translated.

PR #7848, opened 2026-09-18
Feature/forever

Logonz:
## Issue references

Fixes #

## Proposed changes

-
-


Modules/QuestieCompat.lua
```diff
@@ -1,16 +1,19 @@
...
 
 local tinsert = table.insert
 
--- QuestieDB loads later in the TOC, but ImportModule hands back the table CreateModule will populate, so this
--- reference is the real module by the time any function below runs. This file is only able to import at all
--- because QuestieLoader now loads ahead of it; it used to run first, when the global did not yet exist.
+-- Compatibility loads before VersionCheck and embedded libraries, so Questie.IsForever is not available yet.
+-- Forever shares Retail's project ID; use the same interface range as VersionCheck instead.
+local _, _, _, interfaceVersion = GetBuildInfo()
+local isForever = interfaceVersion >= 16000 and interfaceVersion < 17000
```
Logonz, line 22: We should think about if we should move VersionCheck earlier, it is becoming more and more important.

Modules/Tooltips/Tooltip.lua
```diff
@@ -511,29 +593,32 @@ function QuestieTooltips:Initialize()
...
+                local sName, spell = self:GetSpell()
+                if (uName == nil and unit == nil and iName == nil and link == nil and sName == nil and spell == nil) and (
+                        QuestieTooltips.lastGametooltip ~= GameTooltipTextLeft1:GetText() or
+                        (not QuestieTooltips.lastGametooltipCount) or
+                        _QuestieTooltips:CountTooltip() < QuestieTooltips.lastGametooltipCount
+                        or QuestieTooltips.lastGametooltipType ~= "object"
+                    ) and (not self.ShownAsMapIcon) then -- We are hovering over a Questie map icon which adds its own tooltip
+                    local playerZone = QuestiePlayer:GetCurrentZoneId()
+                    _QuestieTooltips.AddObjectDataToTooltip(GameTooltipTextLeft1:GetText(), playerZone)
+                    GameTooltip:Show() -- Classic must resize after appending lines outside the native render pass.
```
Logonz, line 621: Please validate this moved `Show()` call before merging. We are keeping the new ownership model: Classic shows after object augmentation; Forever lets Blizzard's processor show after its post-calls.

This is not strictly a relocation with identical behavior: the Classic caller now reaches `Show()` even when the renderer returns early (tooltips disabled or missing name), and `lastGametooltipType` is now assigned before `Show()`. Those differences need explicit checking.

**Test recipe, on both Classic and Forever:**

1. In Questie options, enable **General → Enable World Tooltips** and **Advanced → Show Object IDs**. Find a collection object for an active quest that already has Questie objective lines. A Wanted poster for an available quest can also cover initial rendering. Note the client/build, object name, and quest ID.
2. Hover it, move onto empty ground until the tooltip disappears, and repeat five times. Alternate between that object, a nearby NPC, and a bag item. Expect one Object ID row, one copy of each Questie line, correct tooltip height with no clipped text, and no lines carried over from the previous target. Empty-ground/fade transitions must not revive an empty or stale tooltip.
3. Turn **Enable World Tooltips** off and hover the same object again. Expect only native content, normal sizing/fading, and no forced reappearance. Re-enable it and confirm Questie's additions return. This specifically exercises the renderer's early return while this caller still invokes `Show()`.
4. On Forever, also test a stationary hover while ordinary quest progress changes, when a suitable shared objective/party setup allows it. Expect updated counts without a second title/objective block. Then leave and re-hover the same object. Record this case as untested if no actual refresh occurred; re-hovering alone does not prove rebuild handling.
5. During ordinary combat, hover the same object, an NPC, and a bag item, then repeat after leaving combat. Expect no Lua/protected-action errors, flicker, or stuck tooltip. If data becomes inaccessible, native content should remain intact. Record the exact error/stack and whether it occurs while entering combat, hovering, or leaving combat.

Restore the two tooltip settings afterward. Please report Classic and Forever results separately, especially sizing, disabled-tooltip behavior, and stationary refreshes.

Already checked: 1,958 automated tests, clean lint, and out-of-combat Classic rendering on 2.5.6 build 69795. These do **not** establish live Forever refresh/combat safety or separate Era coverage. Unit/Item handlers still have their existing text/count logic, so errors there should be distinguished from this Object path.

